PR #1, opened 2024-01-29
Chore/audit exclude

jkoenig134:


.ci/runChecks.sh
```diff
@@ -5,4 +5,4 @@ npm run build --workspaces
 npm run lint:eslint
 npm run lint:prettier
 npx -ws license-check
-npx better-npm-audit audit
+npx better-npm-audit audit --exclude 1096302
```
slavistan, line 8: Just a question I've had for a while: Why are we doing the npm audits in the first place if we're exluding any vulnerabilities anyway?

jkoenig134: we exclude when we don't have the option to fix it. We check it to even HAVE the option to fix it :)

stnmtz: In a sense, we need to check for Security Issues regularly (and not only while pushing). On the other hand, Github is doing that for us e.g. via CodeQL. However, we need to start somewhere and we're doing it within our pipelines.

Actively excluding security issues is a way not to block our development by external security issues (as the pipeline would fail), while keeping track of which issues are still open.

stnmtz (approved):

.ci/runChecks.sh
slavistan: >Actively excluding security issues is a way not to block our development by external security issue

Same thing as running only the command to fix what's possible to fix and ignore everything afterwards. If we did things this way, there would be no need to manually `--exclude` anything, which is a slight nuisance. What do you think about that?

jkoenig134: We don't have that command. There's no easy way to find out what is fixable and what not (because we can sometimes go to other libs first and fix things there and then go back and fix it here).

stnmtz: An automated call to npm audit fix might fix the security issue, if it has been fixed already. However, it might be better to update the references instead. An automated npm-check-updates might break other things. Worse, a security vulnerability might be inside the code without us knowing, because nobody looked at the warnings within the pipeline logs.
Thus, we decided to go the manual route, blocking the build and forcing the devs to think about the mitigation to this security issue, to mitigate it manually OR to accept it manually.

PR #16, opened 2024-02-09
Feature/query messages with attachments

jkoenig134:
## Summary by CodeRabbit

- **New Features**
	- Enhanced message querying to handle specific input values related to attachments.
- **Tests**
	- Updated test utilities to support message exchange with attachments.
	- Added new tests for querying messages with attachments.

### DIFF
--- a/packages/runtime/test/lib/testUtils.ts
+++ b/packages/runtime/test/lib/testUtils.ts
@@ -221,9 +221,9 @@ export async function sendMessage(transportServices: TransportServices, recipien
     return response.value;
 }
 
-export async function exchangeMessage(transportServicesCreator: TransportServices, transportServicesRecipient: TransportServices): Promise<MessageDTO> {
-    const recipientAddress = (await getRelationship(transportServicesCreator)).peer;
-    const messageId = (await sendMessage(transportServicesCreator, recipientAddress)).id;
+export async function exchangeMessage(transportServicesCreator: TransportServices, transportServicesRecipient: TransportServices, attachments?: string[]): Promise<MessageDTO> {
+    const recipientAddress = (await transportServicesRecipient.account.getIdentityInfo()).value.address;
+    const messageId = (await sendMessage(transportServicesCreator, recipientAddress, undefined, attachments)).id;
     const messages = await syncUntilHasMessages(transportServicesRecipient);
     expect(messages).toHaveLength(1);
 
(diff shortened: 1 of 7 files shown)


packages/runtime/test/lib/testUtils.ts
```diff
@@ -235,15 +235,8 @@ export async function exchangeMessage(transportServicesCreator: TransportService
...
-    const recipientAddress = (await getRelationship(transportServicesCreator)).peer;
-    const messageId = (await sendMessage(transportServicesCreator, recipientAddress, undefined, [file.id])).id;
-    const messages = await syncUntilHasMessages(transportServicesRecipient);
-    expect(messages).toHaveLength(1);
-
-    const message = messages[0];
-    expect(message.id).toStrictEqual(messageId);
 
-    return message;
+    return await exchangeMessage(transportServicesCreator, transportServicesRecipient, [file.id]);
```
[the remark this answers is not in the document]
jkoenig134: Our lint rules and code standards require this await

packages/runtime/test/transport/messages.test.ts
```diff
@@ -189,4 +190,13 @@ describe("Message query", () => {
...
     });
+
+    test("query Messages withAttachments", async () => {
+        await exchangeMessageWithAttachment(transportServices1, transportServices2);
+        await exchangeMessage(transportServices1, transportServices2);
+
+        const messages = await transportServices2.messages.getMessages({ query: { attachments: "+" } });
+
+        expect(messages.value.every((m) => m.attachments.length > 0)).toBe(true);
+    });
```
tnotheis, line 201: That test is weird. :) I would save the two messages and in the assertion, I would do the following assertions:

```c#
expect(messages).toHaveLength(1);
expect(messages[0].id).toBe(messageWithAttachments.Id)
```

tnotheis (changes requested):

packages/runtime/test/transport/messages.test.ts
jkoenig134: actually I can't rely on having an empty list of messages when there were other tests before.

jkoenig134: But I can add the second check, that the first message is contained in the array.

tnotheis: Ahh, I expected the transportServices to be reinitialized before each test.

jkoenig134: Is it better now?

sebbi08 (approved):

tnotheis (approved):

PR #23, opened 2024-02-14
Refactor/runtime tests with new test utilities

Magnus-Kuhn:
no work on the tests attribute.test.ts and attribute-e2e.test.ts because Milena is doing them

(Presumably) a race condition for the syncUntil-function causes tests to occasionally fail.

### DIFF
--- a/packages/runtime/test/consumption/iqlQuery.test.ts
+++ b/packages/runtime/test/consumption/iqlQuery.test.ts
@@ -4,7 +4,7 @@ import { IQLQueryJSON, ReadAttributeRequestItemJSON } from "@nmshd/content";
 import { DateTime } from "luxon";
 import { ConsumptionServices, LocalAttributeDTO, LocalRequestDTO, MessageDTO, OutgoingRequestCreatedEvent, OutgoingRequestStatusChangedEvent, TransportServices } from "../../src";
 import { IncomingRequestReceivedEvent, IncomingRequestStatusChangedEvent } from "../../src/events";
-import { establishRelationship, RuntimeServiceProvider, syncUntilHasMessages } from "../lib";
+import { establishRelationship, RuntimeServiceProvider, syncUntilHasMessageWithRequest, syncUntilHasMessageWithResponse } from "../lib";
 
 /* Disable timeout errors if we're debugging */
 if (process.env.NODE_OPTIONS !== undefined && process.env.NODE_OPTIONS.search("inspect") !== -1) {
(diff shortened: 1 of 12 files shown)


jkoenig134 (changes requested):
Only one small comment. Rest LGTM.

packages/runtime/test/lib/testUtils.ts
```diff
@@ -87,7 +87,7 @@ export async function syncUntilHasMessageWithRequest(transportServices: Transpor
 
 export async function syncUntilHasMessageWithResponse(transportServices: TransportServices, requestId: string | CoreId): Promise<MessageDTO> {
     const filterResponseMessagesByRequestId = (syncResult: SyncEverythingResponse) => {
-        return syncResult.messages.filter((m) => m.content["@type"] === "ResponseWrapper" && m.content.requestId === requestId.toString());
+        return syncResult.messages.filter((m) => (m.content["@type"] === "Response" || m.content["@type"] === "ResponseWrapper") && m.content.requestId === requestId.toString());
```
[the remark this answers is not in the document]
Magnus-Kuhn: the runtime test "sender: sync Message with Response" in requests.test.ts sends a Response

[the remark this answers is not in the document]
jkoenig134: you should switch to ResponseWrapper there.

packages/runtime/test/consumption/iqlQuery.test.ts
```diff
@@ -320,20 +313,13 @@ describe("IQL Query", () => {
...
-    });
-
-    test("sender: complete the outgoing Request with Response from Message", async () => {
+    test("sender: sync Message with Response and complete the outgoing Request with Response from Message", async () => {
         let triggeredEvent: OutgoingRequestStatusChangedEvent | undefined;
         sEventBus.subscribeOnce(OutgoingRequestStatusChangedEvent, (event) => {
             triggeredEvent = event;
         });
 
+        sResponseMessage = await syncUntilHasMessageWithResponse(sTransportServices, sLocalRequest.id);
```
slavistan, line 322: This `sReponseMessage` should be a local variable (as well as a lot of others which are defined in the header of the "IQL Query" test suite), which is part of a bigger problem: the tests share state, which is not something we want, as that makes it impossible to run and debug a single test in isolation.

Should be rectified either in this PR or a separate one.

slavistan: I'm still uncertain about if the tests run in parallel or not. If some of these tests are in fact run in parallel sharing state between them would explain any occasional test failures.

jkoenig134: @slavistan tests does never run in parallel, that's only the case for test suites.

slavistan: I created a separate backlog item for these coupled tests.

jkoenig134 (approved):

PR #24, opened 2024-02-15
Feature/add MessageWasReadAtChanged event

jkoenig134:
Needed in the app to react on these changes.

### DIFF
--- a/packages/runtime/src/events/EventProxy.ts
+++ b/packages/runtime/src/events/EventProxy.ts
@@ -16,7 +16,14 @@ import {
     PeerSharedAttributeSucceededEvent,
     RepositoryAttributeSucceededEvent
 } from "./consumption";
-import { MessageDeliveredEvent, MessageReceivedEvent, MessageSentEvent, PeerRelationshipTemplateLoadedEvent, RelationshipChangedEvent } from "./transport";
+import {
+    MessageDeliveredEvent,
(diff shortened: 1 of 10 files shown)


packages/runtime/src/events/EventProxy.ts
```diff
@@ -16,7 +16,14 @@ import {
     PeerSharedAttributeSucceededEvent,
     RepositoryAttributeSucceededEvent
 } from "./consumption";
-import { MessageDeliveredEvent, MessageReceivedEvent, MessageSentEvent, PeerRelationshipTemplateLoadedEvent, RelationshipChangedEvent } from "./transport";
+import {
+    MessageDeliveredEvent,
+    MessageReceivedEvent,
+    MessageSentEvent,
+    MessageWasReadAtChangedEvent,
```
sebbi08, line 23: Shouldn't this be called `MessageWasRead**And**ChangedEvent` ?

jkoenig134: the property of the `Message` is called `wasReadAt`. This event notifies for when this property was changed.

sebbi08: ah, ok still a little bit confusing but also have no better idea

sebbi08 (approved):

PR #26, opened 2024-02-16
init;wip

RuthDiG:


tnotheis (changes requested):
I just stumbled over this PR, and we really shouldn't merge this. While in general it is a good idea to avoid such strings, in this specific case it's a huge mistake. 
This is because error codes are part of our public API. This means that changing an error code is always a breaking change, because users of the runtime and the connector may check for these error codes. 

This PR makes accidental changes to error codes way too likely, because the developer may not notice that by changing the method name he also changes the error code.

RuthDiG:
> I just stumbled over this PR, and we really shouldn't merge this. While in general it is a good idea to avoid such strings, in this specific case it's a huge mistake. This is because error codes are part of our public API. This means that changing an error code is always a breaking change, because users of the runtime and the connector may check for these error codes.
> 
> This PR makes accidental changes to error codes way too likely, because the developer may not notice that by changing the method name he also changes the error code.

I understand so we should reject this issue? https://js-soft.atlassian.net/browse/NMSHDR-15

tnotheis:
> I understand so we should reject this issue? https://js-soft.atlassian.net/browse/NMSHDR-15

IMO yes. @slavistan Did you think of what I described above when you created the task?

slavistan:
I disagree with your points\* but I don't have a strong enough opinion on this topic to warrant any discussion. The idea here was more to provide an instructive task for somebody to get their hands dirty with our codebase and the typescript fundamentals while at the same time doing something arguably useful.

 \* If you're interested in the details: Apropos
- breaking changes: We're introducing a lot of breaking changes at the moment across a lot of our libraries, so now is the time.
- changes to a public api: We're the only ones using the consumption library. If anything, the runtime lib api may be considered public in a sense, but that wouldn't necessarily be affected by the changes here, as the runtime exports its own errors.
- mistakes: Mistakes are possible everywhere, this isn't specific in any way to the changes suggested here. Plus, changes introduced to the error codes would be caught by our tests.

But again, no strong opinion. Yeet it, if you don't like it, all good 👍

tnotheis:
> We're introducing a lot of breaking changes at the moment across a lot of our libraries, so now is the time

I'm not saying it's a breaking change right now. Instead, the changes increase the likelihood that breaking changes will be introduced inadvertently in the future.

> We're the only ones using the consumption library.

First of all: the runtime is an open source project. So in theory, there can be others using it as well. But more important: the Connector passes the error code it receives from the transport or consumption library on to the caller. This means that everybody integrating the Connector into their environment is affected by changes to the error codes.

> changes introduced to the error codes would be caught by our tests.

If you can prove that all error codes are covered by tests, I'm fine with the changes ;) But the truth is that we're far from 100% test coverage - at least back then when I was developing for the runtime.

jkoenig134:
I second @tnotheis opinion and therefore close this.

PR #27, opened 2024-02-16
Fix/validation for birthdate

britsta:
When a BirthDate is created, validation is currently only performed on the individual BirthDay, BirthMonth and BirthYear components. This means, for example, that it is checked that the selected BirthMonth is an integer between 1 and 12.

There is currently no cross-component validation. This means that it is possible, for example, to create 31 February 2000 as the BirthDate. Cross-component validation for BirthDate should now be added to prevent this.


packages/content/src/attributes/types/birth/BirthDate.ts
```diff
@@ -37,6 +39,33 @@ export class BirthDate extends AbstractComplexValue implements IBirthDate {
...
+                "The BirthDate is not a valid date. The chosen day does not exist in the chosen month."
+            );
+        }
+        if (value.month.value === 2 && !value.isInLeapYear() && value.day.value === 29) {
+            throw new ValidationError(
+                BirthDate.name,
+                nameof<BirthDate>((x) => x.year),
+                "The BirthDate is not a valid date. The 29 February only exists in leap years."
+            );
+        }
```
jkoenig134, line 60: we have a great library (luxon) that handles everything date related for us. This library is also able to create dates from day month and year and validate it afterwards. Call me if you need guidance on how to use it.

jkoenig134 (changes requested):
only one small thing, rest LGTM.

packages/content/src/attributes/types/birth/BirthDate.ts
britsta: I now use the isInLeapYear function from Luxon instead of implementing my own. Was that your main concern or should I ideally express the IF queries with Luxon as well?

jkoenig134: Actually you can just use 

```ts
const isValid = DateTime.fromObject({ day: value.day.value, month: value.month.value, year: value.year.value }).isValid;
```

to check if the date is valid, I don't see a reason to give the user this fine-granular feedback about why this date isn't correct.

packages/content/src/attributes/types/birth/BirthDate.ts
```diff
@@ -37,6 +39,36 @@ export class BirthDate extends AbstractComplexValue implements IBirthDate {
...
+            year: value.year.value
+        }).isInLeapYear;
+
+        if (value.month.value === 2 && !valueIsInLeapYear && value.day.value === 29) {
+            throw new ValidationError(
+                BirthDate.name,
+                nameof<BirthDate>((x) => x.year),
+                "The BirthDate is not a valid date. The 29 February only exists in leap years."
+            );
+        }
```
slavistan, line 67: You could just construct a luxon `DateTime` object and check if its valid.

```ts
import { DateTime } from "luxon";

const invalid = DateTime.fromObject({ year: 2023, month: 2, day: 29 });
console.log(invalid.isValid); // false

const valid = DateTime.fromObject({ year: 2024, month: 2, day: 29 });
console.log(valid.isValid); // true
```

This `.isValid` will tell you clearly whether a day is valid or not and luxon will catch any other errors unrelated to leap years, too. I understand you want to manually check for different cases to return nice errors, but I strongly suspect this to be overengineering. If anybody enters an invalid date, a simple message that the date is, in fact, invalid will help the user enough to rectify the errors.

The main use case for errors will be due to people mixing up the year, day and month fields when constructing the error. People messing up leap year dates is rather a fringe case, so no use putting too much effort into it. although I don't have a problem with you doing it anyway. Just make sure you catch _all_ errors before taking care of individual edge cases.

jkoenig134: yeah - check my comment. That's exactly what I was saying. I'd definitely vote for checking the Date at the end using luxon and make sure you didn't missed sth.

jkoenig134: now the question is only 

a) *only* validation by luxon
b) do the things we know ourselves

I'd vote with a.

slavistan: I'd second that. Just check `.isValid` and return a generic error message such as "the date you entered is invalid".

packages/content/src/attributes/types/birth/BirthDate.ts
```diff
@@ -41,28 +40,21 @@ export class BirthDate extends AbstractComplexValue implements IBirthDate {
...
-
-        if ((value.month.value === 2 && value.day.value === 30) || ([2, 4, 6, 9, 11].includes(value.month.value) && value.day.value === 31)) {
             throw new ValidationError(
                 BirthDate.name,
-                nameof<BirthDate>((x) => x.day),
-                "The BirthDate is not a valid date. The chosen day does not exist in the chosen month."
+                `${nameof<BirthDate>((x) => x.day)}, ${BirthDate.name}.${nameof<BirthDate>((x) => x.month)} or ${BirthDate.name}.${nameof<BirthDate>((x) => x.year)}`,
+                "An unexpected error has occured with a BirthDate."
             );
         }
```
jkoenig134, line 48: `throw new Error("this should never happen");'

jkoenig134: also, please make this inline.

packages/content/src/attributes/types/birth/BirthDate.ts
```diff
@@ -37,6 +38,29 @@ export class BirthDate extends AbstractComplexValue implements IBirthDate {
...
+            );
+        }
+
+        const dateTime = DateTime.fromObject({ day: value.day.value, month: value.month.value, year: value.year.value });
+        const isValid = dateTime.isValid;
+
+        if (!isValid) {
+            throw new ValidationError(
+                BirthDate.name,
+                `${nameof<BirthDate>((x) => x.day)}, ${BirthDate.name}.${nameof<BirthDate>((x) => x.month)} or ${BirthDate.name}.${nameof<BirthDate>((x) => x.year)}`,
```
jkoenig134, line 56: I think we should only use `BirthDate.name` here as we cannot specify which caused the error

britsta: The error message then looks like this: "BirthDate. :: ..."

I don't like it when there is a dot after the BirthDate without a property following it. Is there a way to remove the dot?

jkoenig134: argh, that's ugly..

jkoenig134: wtb this:

```
export class ValidationErrorWithoutProperty extends ValidationError {
    public constructor(type: string, reason: string, cause?: Error) {
        super(type, "", reason, cause);

        this.message = `${type} :: ${reason}`;
    }
}
```

if you use this you can yeet the property completely

britsta: Sounds good, in which file/which location would you implement that class?

jkoenig134: `src/ValidationErrorWithoutProperty.ts`

jkoenig134 (approved):

packages/runtime/test/consumption/attributes/attributes-e2e.test.ts
jkoenig134, line 1: can you move this file back?

packages/runtime/test/consumption/attributes/attributes.test.ts
jkoenig134, line 1: see above

jkoenig134 (approved):

PR #28, opened 2024-02-16
fix/extend ValueHints for all AbstractAddresses

jkoenig134:
## Summary by CodeRabbit

- **Refactor**
	- Improved the method of copying value hints in address-related features for enhanced performance and reliability.

### DIFF
--- a/packages/content/src/attributes/types/address/DeliveryBoxAddress.ts
+++ b/packages/content/src/attributes/types/address/DeliveryBoxAddress.ts
@@ -64,7 +64,7 @@ export class DeliveryBoxAddress extends AbstractAddress implements IDeliveryBoxA
     public state?: State;
 
     public static override get valueHints(): ValueHints {
-        return ValueHints.from({
+        return super.valueHints.copyWith({
             propertyHints: {
                 [this.propertyNames.userId.$path]: ValueHints.from({}),
                 [this.propertyNames.deliveryBoxId.$path]: ValueHints.from({}),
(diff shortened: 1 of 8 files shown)


jkoenig134:
@coderabbitai review

packages/content/src/attributes/types/address/DeliveryBoxAddress.ts
```diff
@@ -64,7 +64,7 @@ export class DeliveryBoxAddress extends AbstractAddress implements IDeliveryBoxA
     public state?: State;
 
     public static override get valueHints(): ValueHints {
-        return ValueHints.from({
+        return super.valueHints.copyWith({
```
slavistan, line 67: May I kindly ask you to briefly explain what the purpose of these changes are?

jkoenig134: Sure.

Here we are "extending" [the valueHints from AbstractAddress](https://github.com/nmshd/runtime/blob/main/packages/content/src/attributes/types/address/AbstractAddress.ts#L22) like we do with the render hints underneath.

This call now adds propertyHints for recipient to the valueHints.

This is just a bugfix and was missed before. With StreetAddress for example [we also do this properly](https://github.com/nmshd/runtime/blob/main/packages/content/src/attributes/types/address/StreetAddress.ts#L66).

jkoenig134: I you need a deeper explanation I'd suggest to not let this block the PR and we do it afterwards.

slavistan (approved):

PR #29, opened 2024-02-19
refactor/decouple runtime tests

Magnus-Kuhn:
create helper methods in the testUtils


slavistan:
In `attribute.test.ts`:

1. L67: Why is the event bus of service 3 not reset, too?
2. L72: In `cleanupAttributes()` use one nested for-loop over all clients and their respective attributes. Also, I find `cleanup` unnecessarily vague. We're deleting attributes, simple as that. How about `deleteAttributes`?

packages/runtime/test/consumption/iqlQuery.test.ts
```diff
@@ -109,29 +110,21 @@ describe("IQL Query", () => {
...
-        const result = await sTransportServices.messages.sendMessage({
-            content: sLocalRequest.content,
-            recipients: [(await rTransportServices.account.getIdentityInfo()).value.address]
-        });
-
-        expect(result).toBeSuccessful();
-
-        sRequestMessage = result.value;
+        await sendMessageWithRequest(sRuntimeServices, rRuntimeServices, requestContent);
     });
```
slavistan, line 116: Just test that the function doesn't throw and remove the eslint exception.

```ts
await expect(() => sendMessageWithRequest(sRuntimeServices, rRuntimeServices, requestContent)).not.toThrow();
```

packages/runtime/test/lib/testUtils.ts
```diff
@@ -234,6 +255,11 @@ export async function exchangeMessage(transportServicesCreator: TransportService
     return message;
 }
 
+export async function exchangeMessageWithRequest(sender: TestRuntimeServices, recipient: TestRuntimeServices, request: CreateOutgoingRequestRequest): Promise<MessageDTO> {
```
slavistan, line 258: Poor naming. We're not exchanging any messages here. Rather we send a message and wait for its reception.

How about `sendMessageWithRequestAndWaitForReception` or even better, reuse `sendMessageWithRequest` and add a parater `waitForReception`, which will execute the final `syncUntil...`.

packages/runtime/test/lib/testUtilsWithInactiveModules.ts
```diff
@@ -0,0 +1,136 @@
...
+    rResponseMessage: MessageDTO;
+    sResponseMessage: MessageDTO;
+}
+
+export interface RelationshipRequestWithRelationshipIfAccepted {
+    request: LocalRequestDTO;
+    relationship: RelationshipDTO | undefined;
+}
+
+export async function exchangeMessageAndReceiverRequiresManualDecision(
```
slavistan, line 21: Confusing name :( My brain doesn't know how to parse it. Can you come up with something else?

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -487,3 +510,876 @@ describe("get repository, own shared and peer shared attributes", () => {
         });
     });
 });
+
+describe(CreateRepositoryAttributeUseCase.name, () => {
```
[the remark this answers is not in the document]
Magnus-Kuhn: Agreed.

Magnus-Kuhn:
> 2. L72: In `cleanupAttributes()` use one nested for-loop over all clients and their respective attributes. Also, I find `cleanup` unnecessarily vague. We're deleting attributes, simple as that. How about `deleteAttributes`?

I like the previous name hinting at the usage - and as someone who has struggled with excess attributes, I understand cleanup to mean delete

packages/runtime/test/lib/testUtilsWithInactiveModules.ts
Magnus-Kuhn: I'm now at "exchangeMessageWithRequestAndRequireManualDecision" - however this is only used in the more technical no-module case. That outsiders don't understand it seems more of a feature than a bug

packages/runtime/src/extensibility/facades/consumption/IncomingRequestsFacade.ts
```diff
@@ -23,6 +23,7 @@ import {
 } from "../../../useCases";
 
 export class IncomingRequestsFacade {
+    [key: string]: any;
```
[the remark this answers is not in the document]
Magnus-Kuhn: this was because I originally typed the method names as string in dual function accept/request tests, I changed it to a more specific type to remove this

packages/runtime/test/lib/testUtilsWithInactiveModules.ts
slavistan: Sounds good.

packages/runtime/test/modules/RequestModule.test.ts
```diff
@@ -133,6 +142,9 @@ describe("RequestModule", () => {
         });
```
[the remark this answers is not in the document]
Magnus-Kuhn: Fixed it and some more problematic tests - however it's not possible to run any arbitrary combination of tests - e.g. the last "new relationship" test fails for some partially completed previous relationship creation flows

slavistan (approved):
Not bumping the version since we only changed tests and no implementation code.

PR #30, opened 2024-02-19
Refactor/attributes-e2e test

Milena-Czierlinski:
Use the new test utilities to clean up the attributes e2e test. Also some general refactoring according to the clean code book recommended by Timo.


packages/runtime/src/useCases/consumption/attributes/AttributeSuccessionResponse.ts
```diff
@@ -1,6 +1,13 @@
...
 
 export interface AttributeSuccessionResponse {
     predecessor: LocalAttributeDTO;
     successor: LocalAttributeDTO;
 }
+
+export interface AttributeSuccessionWithNotificationResponse {
+    predecessor: LocalAttributeDTO;
+    successor: LocalAttributeDTO;
+    notificationId: CoreId;
```
jkoenig134, line 12: we never return CoreIds from DTOs or responses. In this case we need a string

packages/runtime/src/useCases/consumption/attributes/AttributeSuccessionResponse.ts
```diff
@@ -1,6 +1,13 @@
+import { CoreId } from "@nmshd/transport";
 import { LocalAttributeDTO } from "../../../types";
 
 export interface AttributeSuccessionResponse {
     predecessor: LocalAttributeDTO;
     successor: LocalAttributeDTO;
 }
+
+export interface AttributeSuccessionWithNotificationResponse {
```
jkoenig134, line 9: you export this type to use it only in the tests, why?

jkoenig134: I think you can just use `NotifyPeerAboutRepositoryAttributeSuccessionResponse` everywhere where you used `AttributeSuccessionWithNotificationResponse` and `AttributeSuccessionResponse` and then remove this entire file :)

Milena-Czierlinski: Right, initially (in cns-runtime) I used AttributeSuccessionWithNotificationResponse as Reponse type for the use cases NotifyPeerAboutRepositoryAttributeSuccession and SucceedRelationshipAttributeAndNotifyPeer.  I realized that I needed it for a helper, which handles RepositoryAttributes, as well as RelationshipAttributes. Hence, it didn't seem right to me to use one of both reponses, but instead implement a new one and use it in the use cases as well. Would it be agreeable to keep AttributeSuccessionWithNotificationReponse and use it in the two use cases?

AttributeSuccessionReponse was already there. It implements the same than SucceedRepositoryAttributeReponse. I guess this can be deleted and SucceedRepositoryAttributeReponse can be used instead.

packages/runtime/src/useCases/consumption/attributes/AttributeSuccessionResponse.ts
Milena-Czierlinski: Only we currently do in NotifyPeerAboutRepositoryAttributeSuccessionResponse and SucceedRelationshipAttributeAndNotifyPeerResponse. Independently of the result of the discussion below, should I change the type of notificationId?

jkoenig134: yes, please!

PR #33, opened 2024-02-22
Refactor/disable timeout for all runtime tests while debugging

Magnus-Kuhn:


packages/runtime/test/customMatchers.ts
```diff
@@ -1,6 +1,11 @@
 import { ApplicationError, EventConstructor, Result } from "@js-soft/ts-utils";
 import { MockEventBus } from "./lib";
 
+/* Disable timeout errors if we're debugging */
+if (process.env.NODE_OPTIONS !== undefined && process.env.NODE_OPTIONS.search("inspect") !== -1) {
+    jest.setTimeout(1e9);
+}
```
jkoenig134, line 7: instead of abusing the customMatchers file, please create a new one and reference it in the package.json

slavistan: A little more detail: You can configure files to be loaded by jest before any tests run in the `package.json`'s _jest.setupFilesAfterEnv_ array.

packages/runtime/test/jestSettings.ts
jkoenig134, line 1: sorry, I wasn't specific in my comment before.

please use `disableTimeoutforDebugging.ts`

jkoenig134 (approved):

PR #34, opened 2024-02-22
feature/add query to filter attributes to GetRepositoryAttributesUseCase

sebbi08:
This pull request adds the GetRepositoryAttributesRequestQuery interface to the GetRepositoryAttributesUseCase. This allows for more specific querying of attributes when retrieving repository attributes.


jkoenig134:
two things:

1. remove these `// default: true` comments or make these jsdocs, so that they are exposed to the runtime user
2. runtime version bump (minor)

packages/runtime/src/useCases/consumption/attributes/GetRepositoryAttributes.ts
```diff
@@ -1,12 +1,25 @@
...
+    query?: GetRepositoryAttributesRequestQuery;
+}
+
+export interface GetRepositoryAttributesRequestQuery {
+    createdAt?: string;
+    "content.@type"?: string | string[];
+    "content.tags"?: string | string[];
+    "content.validFrom"?: string | string[];
+    "content.validTo"?: string | string[];
+    "content.key"?: string | string[];
```
Milena-Czierlinski, line 19: Afaik only RelationshipAttributes have a key, so this can be omitted here. Same for isTechnical and confidentiality.

sebbi08:
> If we add a query to GetRepositoryAttributes, should we also do it for GetOwnSharedAttributes and GetPeerSharedAttributes?

The both already have it do you mean `GetVersionsOfAttribute` and `GetSharedVersionsOfRepositoryAttribute`

Milena-Czierlinski:
Indeed, they do. My bad. Regarding `GetVersionsOfAttribute` and `GetSharedVersionsOfRepositoryAttribute`, I don't think it is necessary to add a query there, since mainly only the value changes.

packages/runtime/src/useCases/consumption/attributes/GetRepositoryAttributes.ts
```diff
@@ -26,15 +38,16 @@ export class GetRepositoryAttributesUseCase extends UseCase<GetRepositoryAttribu
     }
 
     protected async executeInternal(request: GetRepositoryAttributesRequest): Promise<Result<GetRepositoryAttributesResponse>> {
-        const query: any = {
-            shareInfo: { $exists: false }
-        };
+        const query: GetAttributesRequestQuery = request.query ?? {};
+        const flattenedQuery = flattenObject(query);
+        const dbQuery = GetAttributesUseCase.queryTranslator.parse(flattenedQuery);
```
jkoenig134, line 43: OH, I just stumbled upon this, please copy the query translator. We shouldn't use query translators from other UseCases. Is this the case for others??

sebbi08: Yea, it is used in 4 use cases

image

slavistan: I have something in the back of my head about that we shouldn't use this `flattenedQuery` method, that it's more a relict of past and that we're supposed to use query objects instead of the flat query strings. @jkoenig134 could you comment on that? This may just be an issue of copy-pasting stale code.

jkoenig134: I think it's old, but still safe to use. I'd say we can use it like this for now.

jkoenig134:
@sebbi08 runtime bump and then ready to go!

slavistan (approved):

jkoenig134 (changes requested):
I block this until it get it's version bumps.

jkoenig134 (approved):

jkoenig134:
merging this b/c @sebbi08 is unavailable.

jkoenig134 (approved):

PR #45, opened 2024-02-29
Chore/Adjust to deletionInfo

Milena-Czierlinski:


packages/runtime/src/useCases/consumption/attributes/GetAttributes.ts
```diff
@@ -35,6 +35,9 @@ export interface GetAttributesRequestQuery {
     "shareInfo.notificationReference"?: string | string[];
     "shareInfo.peer"?: string | string[];
     "shareInfo.sourceAttribute"?: string | string[];
+    deletionInfo?: string | string[];
+    "deletionInfo.deletionStatus"?: string | string[];
+    "deletionInfo.deletionDate"?: string | string[];
```
jkoenig134, line 40: i think there are more Get...Attributes UseCases that need this.

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -131,6 +131,9 @@ export class LocalAttribute extends CoreSynchronizable implements ILocalAttribut
         if (typeof peerAddress !== "undefined") {
             isOwnSharedAttribute &&= this.shareInfo!.peer.equals(peerAddress);
         }
+
+        isOwnSharedAttribute &&= !this.hasPeerSharedAttributeDeletionInfo();
```
jkoenig134, line 135: why does the existence of a deletion info changes the type of an attribute?

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -115,7 +115,7 @@ export class LocalAttribute extends CoreSynchronizable implements ILocalAttribut
     }
 
     public isRepositoryAttribute(ownAddress: CoreAddress): this is RepositoryAttribute {
-        return this.isIdentityAttribute() && !this.isShared() && this.isOwnedBy(ownAddress);
+        return this.isIdentityAttribute() && !this.isShared() && this.isOwnedBy(ownAddress) && !this.hasDeletionInfo();
```
jkoenig134, line 118: like below - I think the deletion info doesn't change the type.

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -147,6 +150,9 @@ export class LocalAttribute extends CoreSynchronizable implements ILocalAttribut
         if (typeof peerAddress !== "undefined") {
             isPeerSharedAttribute &&= this.isOwnedBy(peerAddress);
         }
+
+        isPeerSharedAttribute &&= !this.hasOwnSharedAttributeDeletionInfo();
```
jkoenig134, line 154: see above :)

packages/consumption/src/consumption/CoreErrors.ts
```diff
@@ -150,6 +150,10 @@ class Attributes {
         );
     }
 
+    public cannotSucceedAttributesWithDeletionInfo() {
+        return new CoreError("error.consumption.attributes.cannotSucceedAttributesWithDeletionInfo", "You cannot succeed attributes with a deletionInfo.");
+    }
```
jkoenig134, line 155: I think the error and description doesn't actually show the user what is going wrong (the user doesn't necessarily know about a deletionInfo, just about the deletion itself), so we should just inform the user about the deletion itself but not about the deletionInfo

jkoenig134 (approved):

jkoenig134 (changes requested):
accidentally approved

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
Milena-Czierlinski: RepositoryAttributes are not allowed to have a deletionInfo, since it doesn't make sense. Own shared Attributes may only have the deletionStatus DeletedByPeer or ToBeDeletedByPeer, since DeletedByOwner and ToBeDeleted doesn't make sense and isn't implemented. Vice versa for peer shared Attributes.

packages/consumption/src/consumption/CoreErrors.ts
Milena-Czierlinski: The deletionInfo states that the Attribute may have been deleted by the peer of the Relationship or it may be marked for deletion by yourself or by the owner. So an Attribute may have a deletionInfo, even though no deletion took place, yet. Therefore, I think it is most accurate to use the term deletionInfo. But, I can extent the error message.

slavistan: I'd prefer the technical error message since we're in the consumption library, i.e. _we_ are the user.

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
slavistan: >since it doesn't make sense

By this you mean that we agreed to delete them immediately without any prior process to go through that would warrant a `deletionInfo` to keep track of any state, right?

Just making sure we're on the same boat.

packages/consumption/src/consumption/CoreErrors.ts
jkoenig134: we push this to the runtime and from the runtime to the connector so NO, we aren't :)

jkoenig134: > But, I can extent the error message.
yep, please make sure what the deletion info is about and why you can't do this

slavistan: > we push this to the runtime and from the runtime to the connector so NO, we aren't :)

Well, _if_ we do this, fair enough.

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
Magnus-Kuhn: here also: I think we didn't want to wipe the shareInfo of a shared attribute after deletion, so the isShared check covers this already

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -172,6 +175,35 @@ export class LocalAttribute extends CoreSynchronizable implements ILocalAttribut
...
+
+    public hasPeerSharedAttributeDeletionInfo(): boolean {
+        return this.deletionInfo?.deletionStatus === DeletionStatus.DeletedByOwner || this.deletionInfo?.deletionStatus === DeletionStatus.ToBeDeleted;
+    }
+
+    public hasOwnSharedAttributeDeletionInfo(): boolean {
+        return this.deletionInfo?.deletionStatus === DeletionStatus.DeletedByPeer || this.deletionInfo?.deletionStatus === DeletionStatus.ToBeDeletedByPeer;
+    }
+
+    public hasDeletionInfo(): boolean {
```
jkoenig134, line 203: just a small reminder - no direct CTA: this could be a typeguard, which could help you later on (or maybe also now?) when accessing the deletion info.

jkoenig134 (approved):

PR #47, opened 2024-03-01
Feature/Deletion of own shared Attributes

Milena-Czierlinski:


packages/consumption/src/modules/notifications/itemProcessors/attributeDeleted/OwnSharedAttributeDeletedByOwnerNotificationItemProcessor.ts
```diff
@@ -0,0 +1,82 @@
...
+            updatedAttribute = await this.consumptionController.attributes.updateAttributeUnsafe(attribute);
+        }
+
+        const predecessors = await this.consumptionController.attributes.getPredecessorsOfAttribute(attribute.id);
+        for (const predecessor of predecessors) {
+            if (typeof predecessor.deletionInfo === "undefined") {
+                predecessor.setDeletionInfo(deletionInfo, this.accountController.identity.address);
+                await this.consumptionController.attributes.updateAttributeUnsafe(predecessor);
+            }
+        }
```
slavistan, line 64: You could use a single loop over `[attribute, ...predecessors]`. This would get rid of repetition and would communicate more clearly that you're doing the same thing to the attribute and all of its predecessors.

Milena-Czierlinski: I thought about doing so, but I need `updatedAttribute` for the event that is returned and I thought this way it is cleaner than overwriting the variable all the time or filling an array, if I only need one entry, anyway.

packages/consumption/src/modules/notifications/itemProcessors/attributeDeleted/OwnSharedAttributeDeletedByOwnerNotificationItemProcessor.ts
```diff
@@ -0,0 +1,82 @@
...
+        const attribute = await this.consumptionController.attributes.getLocalAttribute(notificationItem.attributeId);
+        if (typeof attribute === "undefined") return;
+
+        attribute.deletionInfo = undefined;
+        await this.consumptionController.attributes.updateAttributeUnsafe(attribute);
+
+        const predecessors = await this.consumptionController.attributes.getPredecessorsOfAttribute(attribute.id);
+        for (const predecessor of predecessors) {
+            predecessor.deletionInfo = undefined;
+            await this.consumptionController.attributes.updateAttributeUnsafe(predecessor);
```
slavistan, line 79: Use a single loop over `[attribute, ...predecessors]` :-)

packages/consumption/src/modules/notifications/itemProcessors/attributeDeleted/OwnSharedAttributeDeletedByOwnerNotificationItemProcessor.ts
```diff
@@ -0,0 +1,82 @@
...
+    public override async rollback(notificationItem: OwnSharedAttributeDeletedByOwnerNotificationItem, _notification: LocalNotification): Promise<void> {
+        const attribute = await this.consumptionController.attributes.getLocalAttribute(notificationItem.attributeId);
+        if (typeof attribute === "undefined") return;
+
+        attribute.deletionInfo = undefined;
+        await this.consumptionController.attributes.updateAttributeUnsafe(attribute);
+
+        const predecessors = await this.consumptionController.attributes.getPredecessorsOfAttribute(attribute.id);
+        for (const predecessor of predecessors) {
+            predecessor.deletionInfo = undefined;
```
slavistan, line 78: Is `DeletedByOwner` -> `undefined` the only legal state transition?

Milena-Czierlinski: Yes, peer shared Attributes may only have the DeletionStatus `ToBeDeleted` (if an according Request was accepted) or `DeletedByOwner`. `DeletedByPeer` and `ToBeDeletedByPeer` are only allowed for own shared Attributes. If the Attribute had the DeletionStatus `ToBeDeleted`, processing this Notification won't change its status (line 53 and 60). Hence, the peer shared Attribute will only be set to the DeletionStatus `DeletedByOwner`, if the DeletionStatus was undefined before. Now, calling the rollback function, we know that we only need to reset it to undefined.

packages/consumption/src/modules/notifications/itemProcessors/attributeDeleted/OwnSharedAttributeDeletedByOwnerNotificationItemProcessor.ts
Milena-Czierlinski: You were right, I can just use `attribute` further on and don't need `updatedAttribute`.

slavistan (approved):

PR #53, opened 2024-03-05
fix/escape regExp correctly

Siolto:
The new RegExp will match following cases:

- www.google.com
- http://www.google.com
- mailto:somebody@google.com
- somebody@google.com
- www.url-with-querystring.com/?url=has-querystring

We need to specially escape the characters for the validation in the app


jkoenig134:
Version bump is missing

jkoenig134 (approved):

PR #54, opened 2024-03-06
fix/correct phoneNumber regExp

Siolto:


packages/content/src/attributes/types/strings/AbstractPhoneNumber.ts
```diff
@@ -4,14 +4,16 @@ import { AbstractString } from "../AbstractString";
...
+    @validate({ min: 3, max: 100, regExp: new RegExp(/^[\d+\-() ]{3,100}$/) })
     public override value: string;
 
     public static override get valueHints(): ValueHints {
         return super.valueHints.copyWith({
             min: 3,
             max: 100,
-            pattern: "/^[d+-x#*()/[] ]{3,100}$/"
+            // prettier-ignore
+            // eslint-disable-next-line no-useless-escape
```
jkoenig134, line 15: it should not be necessary to have these two ignores.

Siolto: prettier is erasing my `\` and eslint complains about useless-escapes. Not sure how I could do it differently

jkoenig134: wtb double backslash?

Siolto: would break the regex in dart... I could do triple backslash 😄  but that would cause the same problem

jkoenig134 (approved):

PR #56, opened 2024-03-06
test/fix race condition, decouple remaining runtime attribute tests

Magnus-Kuhn:


Magnus-Kuhn:
I decided against refactoring the getAttributes / query tests because the intention behind creating more attributes than strictly required might have been to check for false positives.

packages/runtime/test/dataViews/requestItems/ShareAttributeRequestItemDVO.test.ts
```diff
@@ -223,8 +222,8 @@ describe("ShareAttributeRequestItemDVO", () => {
         const baselineNumberOfAttributes = await syncAndGetBaselineNumberOfAttributes(sRuntimeServices, {
             query: { "content.value.@type": "DisplayName", "shareInfo.peer": rAddress }
         });
+        sEventBus.reset();
```
jkoenig134, line 225: the eventbus should be resetted after every test, why is this required?

Magnus-Kuhn: The syncs also process the accept response of the previous test - and the test that should wait for the processing of the current attribute, would instead run after processing of the old attribute causing an error in the status check

jkoenig134: you can also set filters when waiting for events

packages/runtime/test/dataViews/requestItems/ShareAttributeRequestItemDVO.test.ts
```diff
@@ -224,7 +223,6 @@ describe("ShareAttributeRequestItemDVO", () => {
             query: { "content.value.@type": "DisplayName", "shareInfo.peer": rAddress }
         });
         const senderMessage = await exchangeAndAcceptRequestByMessage(sRuntimeServices, rRuntimeServices, requestContent, responseItems);
-        await sEventBus.waitForEvent(OutgoingRequestStatusChangedEvent, (e) => e.data.newStatus === LocalRequestStatus.Completed);
         const dto = senderMessage;
```
jkoenig134, line 226: shouldn't this be re-added? same for other files where events were removed

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -1110,140 +1110,150 @@ describe("Get (shared) versions of attribute", () => {
     let sOSRAVersion0: LocalAttributeDTO;
     let sOSRAVersion1: LocalAttributeDTO;
     let sOSRAVersion2: LocalAttributeDTO;
```
jkoenig134, line 1112: I don't know when this was introduced, but WHAT IS THIS?! shouldn't those local variables be removed during decoupling the events?

cc @slavistan

packages/runtime/test/dataViews/requestItems/ShareAttributeRequestItemDVO.test.ts
Magnus-Kuhn: this is included in the method `exchangeAndAcceptRequestByMessage` above, I was wrong to put this line in in the first place

jkoenig134: IC

packages/runtime/test/consumption/attributes.test.ts
jkoenig134: IC

Magnus-Kuhn: the helper methods also utilize this setup, are you sure you want this changed? (it would make the helper signatures larger)

packages/runtime/test/lib/testUtils.ts
```diff
@@ -508,12 +508,14 @@ export async function waitForRecipientToReceiveNotification(
 }
 
 /**
- * finds how many attributes will be there in addition to the desired ones
- * Assumes that
+ * finds how many attributes will be there from previous tests in addition to the desired ones
  */
 export async function syncAndGetBaselineNumberOfAttributes(sender: TestRuntimeServices, filter: GetAttributesRequest): Promise<number> {
     await sleep(1000);
```
jkoenig134, line 514: I didn't reviewed the last tests PR, can we get rid of the sleep in any way?

sleeps in tests are bad b/c
1. they bloat the test time (could be that they wait to long)
2. the don't wait long enough and everything explodes

Magnus-Kuhn: I think it's possible but querying the outcoming resp. incoming requests to decide which ones are in the status "recipient accepted but sender has not synced" also feels awkward

Magnus-Kuhn: I have replaced it, but the success of this check run required re-running the mongodb test

jkoenig134: means it becomes flaky b/c you removed the sleep?

Magnus-Kuhn: Tests have always been flaky - the test run that ended up "approving" the merge into main didn't require re-running, then later without any changes it has failed again. Also I re-ran Test #356 from before the change and it also succeeded.

jkoenig134: I mean - sleeps can never be the cure for flakyness. If it's still flaky you should go over the tests again.

Magnus-Kuhn: those specific tests ("check the messageDVO for the sender after acceptance") in the (Complex)ReadAttributeDVO tests look fine to me - I'll ask someone to have a look at it

jkoenig134 (approved):

PR #57, opened 2024-03-08
Feature/Deletion of RepositoryAttributes

Milena-Czierlinski:
## Summary by CodeRabbit

- **New Features**
	- Introduced functionality for deleting repository attributes, including comprehensive validation and permission checks.
	- Enhanced attribute succession logic, ensuring accurate predecessor and successor relationships and content validation.
- **Bug Fixes**
	- Refined error handling for attribute comparisons and specific cases related to successor attributes.
- **Refactor**
	- Updated validation and error handling methods for improved clarity and specificity.
	- Removed outdated logic affecting attribute control flow.
- **Tests**
	- Expanded test coverage to include new scenarios for attribute deletion and succession.
- **Documentation**
	- Updated schema definitions and added detailed descriptions for new and existing properties.


Milena-Czierlinski:
@coderabbitai review

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -599,31 +599,41 @@ export class AttributesController extends ConsumptionBaseController {
...
+            }
+
+            const successorSourceVersionIds = (await this.getVersionsOfAttribute(successorSource.id)).map((x) => x.id.toString());
+            if (typeof predecessorSource.succeededBy === "undefined" || !successorSourceVersionIds.some((id) => id === predecessorSource.succeededBy?.toString())) {
+                return ValidationResult.error(CoreErrors.attributes.successorSourceDoesNotSucceedPredecessorSource());
+            }
+
+            if (!_.isEqual(predecessorSource.content, predecessor.content)) {
+                return ValidationResult.error(CoreErrors.attributes.predecessorSourceContentIsNotEqualToCopyContent());
+            }
```
slavistan, line 636: This error would only occur, if we ourselves had errors in the implementation of a succession or a sharing of attributes. This should be an assertion (for devs trying to catch bugs) than runtime validation of for this particular method. We're no using assertions though, so I'd say yeet it. What do you think?

jkoenig134: But it's also to intercept malicius requests/responses.. IMO it doesn't hurt to validate this, there can't be too much validation..

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -125,10 +125,6 @@ export class LocalAttribute extends CoreSynchronizable implements ILocalAttribut
             return isOwnSharedAttribute;
         }
 
-        if (this.isIdentityAttribute()) {
```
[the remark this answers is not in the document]
Milena-Czierlinski: I adjusted OwnSharedIdentityAttribute and also OwnSharedRelationshipAttribute, since in the case of own shared third party RelationshipAttributes the `shareInfo.sourceAttribute` property may be defined.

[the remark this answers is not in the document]
slavistan: Gotcha. I also noticed that the type `RepositoryAttribute` is incorrectly defined:

```ts
export type RepositoryAttribute = IdentityAttribute & { shareInfo: undefined };
```

This should rather be

```ts
export type RepositoryAttribute = LocalAttribute & { content: IdentityAttribute } & { shareInfo: undefined };
```

Right?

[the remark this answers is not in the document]
jkoenig134: It is! But this can also be done "inline".

```

export type RepositoryAttribute = LocalAttribute & { content: IdentityAttribute, shareInfo: undefined };
```

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -964,9 +975,32 @@ export class AttributesController extends ConsumptionBaseController {
...
+        return ownSharedIdentityAttributePredecessors;
+    }
+
+    public async getSharedSuccessorsOfRepositoryAttribute(repositoryAttribute: LocalAttribute, query?: any): Promise<LocalAttribute[]> {
+        if (typeof query === "undefined") {
+            query = {};
+        }
+
+        const ownSharedIdentityAttributeSuccessors: LocalAttribute[] = [];
+        while (repositoryAttribute.succeededBy) {
```
slavistan, line 990: Use the `typeof foo !== "undefined"` syntax. Using a different construct here suggests that you're doing something else than just checking for `undefined`. Same for the implementation for `getSharedPredecessorsOfRepositoryAttribute`.

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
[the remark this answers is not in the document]
jkoenig134: I see the other types can also be simplified, all those `& {...}` are kind of hard to process.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -964,9 +975,32 @@ export class AttributesController extends ConsumptionBaseController {
...
         }
 
-        return ownSharedIdentityAttributeVersions;
+        return ownSharedIdentityAttributePredecessors;
+    }
+
+    public async getSharedSuccessorsOfRepositoryAttribute(repositoryAttribute: LocalAttribute, query?: any): Promise<LocalAttribute[]> {
+        if (typeof query === "undefined") {
+            query = {};
+        }
```
slavistan, line 987: Create a deep copy of query if it's defined or else the function call will mutate the query object passed as an argument to the function. That's highly unidiomatic :-)

```ts
const query = { "content.@type": "GivenName" }
const successors = getSharedSuccessorsOfRepositoryAttribute(repoAttrId, query)

// query now has an extra 'shareInfo.sourceAttribute' key.
```

Same for the `getSharedPredecessorsOfRepositoryAttribute` method.

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -2067,19 +2193,39 @@ describe("AttributesController", function () {
                 await consumptionController.attributes.succeedOwnSharedIdentityAttribute(ownSharedIdentityAttributeV1PeerB.id, ownSharedIdentityAttributeParamsV2PeerB));
         });
 
+        test("should return all shared predecessors for all peers", async function () {
```
slavistan, line 2196: Check the order of returned predecessors / successors, if that's a behavior you want users of your function to rely on.

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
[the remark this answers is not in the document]
jkoenig134: @slavistan & @Milena-Czierlinski as this could lead to errors I have done this in another branch to main: https://github.com/nmshd/runtime/pull/67

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -933,6 +934,27 @@ describe(NotifyPeerAboutRepositoryAttributeSuccessionUseCase.name, () => {
...
 
+    test("should throw if the predecessor repository attribute was deleted", async () => {
+        const sRAVersion0 = (await services1.consumption.attributes.getAttribute({ id: sOSIAVersion0.shareInfo!.sourceAttribute! })).value;
+        await services1.consumption.attributes.deleteRepositoryAttribute({ attributeId: sRAVersion0.id });
+
+        const notificationResult = await services1.consumption.attributes.notifyPeerAboutRepositoryAttributeSuccession({
+            attributeId: sRAVersion1.id,
+            peer: services2.address
+        });
+        expect(notificationResult).toBeAnError(/.*/, "error.runtime.attributes.noOtherVersionOfRepositoryAttributeHasBeenSharedWithPeerBefore");
```
slavistan, line 945: Tiny improvement: "no _previous_ version ... has been shared"

packages/consumption/test/modules/attributes/AttributesController.test.ts
Milena-Czierlinski: I check for the order per peer, but for the same version shared with multiple peers the order of returns varies.

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -35,11 +35,13 @@ export interface ILocalAttribute extends ICoreSynchronizable {
...
 }
 
-export type OwnSharedIdentityAttribute = LocalAttribute & { content: IdentityAttribute } & {
+export type OwnSharedIdentityAttribute = LocalAttribute & {
+    content: IdentityAttribute;
     shareInfo: LocalAttributeShareInfo;
 };
 
-export type OwnSharedRelationshipAttribute = LocalAttribute & { content: RelationshipAttribute } & {
+export type OwnSharedRelationshipAttribute = LocalAttribute & {
```
slavistan, line 43: Own shared relationship attributes never have the `shareInfo.sourceAttribute` field set. So its

```ts
export type OwnSharedRelationshipAttribute = LocalAttribute & {
    content: RelationshipAttribute;
    shareInfo: LocalAttributeShareInfo & { sourceAttribute: undefined };
};
```

slavistan: image

Milena-Czierlinski:
I mistakenly added the removal of the 'succeeds' field of own shared IdentityAttribute successors, if a RepositoryAttribute is deleted. Since the own shared IdentityAttributes aren't automatically deleted in that process, the 'succeeds' field of an own shared IdentityAttribute successor will still link to an existing Attribute. Only if the shared copy is deleted, the shared successor needs to be adjusted. This is taken care of in #70 .

packages/runtime/src/useCases/consumption/attributes/DeleteRepositoryAttribute.ts
```diff
@@ -0,0 +1,68 @@
...
+
+        if (!repositoryAttribute.isRepositoryAttribute(this.accountController.identity.address)) {
+            return Result.fail(RuntimeErrors.attributes.isNotRepositoryAttribute(repositoryAttributeId));
+        }
+
+        const ownSharedIdentityAttributeCopies = await this.attributeController.getLocalAttributes({ "shareInfo.sourceAttribute": repositoryAttributeId.toString() });
+        const ownSharedIdentityAttributePredecessors = await this.attributeController.getSharedPredecessorsOfRepositoryAttribute(repositoryAttribute);
+        for (const ownSharedAttribute of [...ownSharedIdentityAttributeCopies, ...ownSharedIdentityAttributePredecessors]) {
+            if (!ownSharedAttribute.isOwnSharedAttribute(this.accountController.identity.address)) {
+                return Result.fail(RuntimeErrors.attributes.isNotOwnSharedAttribute(ownSharedAttribute.id));
```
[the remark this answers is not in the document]
Milena-Czierlinski: If everything goes right, this error should never be called. We need it, however, so that the compiler knows in the following line that `ownSharedAttribute.shareInfo` is defined and we can set its `sourceAttribute` property.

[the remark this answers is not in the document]
jkoenig134: I don't think we should return a result in this case and just throw an error. It's absolutely not expected that this happens (when nobody messes up the data from outside).

packages/runtime/src/useCases/consumption/attributes/DeleteRepositoryAttribute.ts
```diff
@@ -39,7 +39,7 @@ export class DeleteRepositoryAttributeUseCase extends UseCase<DeleteRepositoryAt
         const ownSharedIdentityAttributePredecessors = await this.attributeController.getSharedPredecessorsOfRepositoryAttribute(repositoryAttribute);
         for (const ownSharedAttribute of [...ownSharedIdentityAttributes, ...ownSharedIdentityAttributePredecessors]) {
             if (!ownSharedAttribute.isOwnSharedAttribute(this.accountController.identity.address)) {
-                return Result.fail(RuntimeErrors.attributes.isNotOwnSharedAttribute(ownSharedAttribute.id));
+                throw RuntimeErrors.attributes.isNotOwnSharedAttribute(ownSharedAttribute.id);
```
jkoenig134, line 42: RuntimeErrors are for **outward facing** communication, so if we want the user to handle it. Please don't use this error, as the user then would think it's his fault.

jkoenig134: I think this comment got lost (is marked as outdated), but not resolved. Same opinion for the same issue.

Another question is: what do we do with the other attributes that were skipped in the process?

sebbi08: Another problem could be that this might leave the attributes in an inconsisten state some with some without source attribute. Why dont we move all the Validation stuff in the validator ? so that we know we can proceed without error

Milena-Czierlinski: @jkoenig134 I adjusted the error. If you are not content with my changes, please be more specific about how to change it.

jkoenig134: My Idea was not to throw a runtime error (with ErrorCode) and instead just an Error

Milena-Czierlinski: I went with throwing an error without code now. Putting it in the validator would duplicate much logic, which doesn't seem clean to me. Also, if we care about this here now, I guess there are a lot of places existing code needs to be checked for throwing Errors (without code) vs returning failing results.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -1338,12 +1360,12 @@ describe("Get (shared) versions of attribute", () => {
                 const result1 = await services1.consumption.attributes.getSharedVersionsOfRepositoryAttribute({ attributeId: version.id });
                 expect(result1.isSuccess).toBe(true);
                 const returnedVersions1 = result1.value;
-                expect(returnedVersions1).toStrictEqual([sOSIAVersion2, sOSIAVersion2FurtherPeer]);
+                expect(new Set(returnedVersions1)).toStrictEqual(new Set([sOSIAVersion2, sOSIAVersion2FurtherPeer]));
```
jkoenig134, line 1363: out of curiosity - why do we need that?

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -889,15 +890,12 @@ describe("AttributesController", function () {
...
-                    await consumptionController.attributes.deleteAttributeUnsafe(attribute.id);
-                }
-            });
-
-            test("should catch if the source attributes do not succeed one another", async function () {
-                const predecessorRepo = await consumptionController.attributes.createLocalAttribute({
+            let predecessorRA: LocalAttribute;
+            let successorRA: LocalAttribute;
+            let predecessorOSIA: LocalAttribute;
+            let successorOSIAParams: IAttributeSuccessorParams;
```
jkoenig134, line 896: please do not use abbreviations (same for the other test files)

packages/runtime/test/consumption/attributes.test.ts
Milena-Czierlinski: The order in which the (databank) query returns copies of the same version shared with different peers varies. Using Set, the order of the compared arrays is neglected.

packages/runtime/src/useCases/consumption/attributes/DeleteRepositoryAttribute.ts
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        const ownSharedIdentityAttributes = await this.attributeController.getLocalAttributes({ "shareInfo.sourceAttribute": repositoryAttributeId.toString() });
+        const ownSharedIdentityAttributePredecessors = await this.attributeController.getSharedPredecessorsOfRepositoryAttribute(repositoryAttribute);
+        for (const ownSharedAttribute of [...ownSharedIdentityAttributes, ...ownSharedIdentityAttributePredecessors]) {
+            if (!ownSharedAttribute.isOwnSharedAttribute(this.accountController.identity.address)) {
+                throw new Error(
+                    `The Attribute ${ownSharedAttribute.id} does not fulfill the requirements of an own shared Attribute, so it cannot be adjusted adequately, deleting its source Attribute.`
+                );
+            }
```
jkoenig134, line 45: you should check this before iterating the attributes to make sure to change all or nothing

packages/runtime/src/useCases/consumption/attributes/DeleteRepositoryAttribute.ts
```diff
@@ -0,0 +1,70 @@
...
+    attributeId: AttributeIdString;
+}
+
+class Validator extends SchemaValidator<DeleteRepositoryAttributeRequest> {
+    public constructor(@Inject schemaRepository: SchemaRepository) {
+        super(schemaRepository.getSchema("DeleteRepositoryAttributeRequest"));
+    }
+}
+
+export class DeleteRepositoryAttributeUseCase extends UseCase<DeleteRepositoryAttributeRequest, void> {
```
jkoenig134, line 17: this usecase has a massive amount of logic, I think most could and should be extracted to the AttributesController

packages/runtime/src/useCases/consumption/attributes/DeleteRepositoryAttribute.ts
```diff
@@ -0,0 +1,70 @@
...
+    public constructor(
+        @Inject private readonly attributeController: AttributesController,
```
jkoenig134, line 19:
```suggestion
        @Inject private readonly attributesController: AttributesController,
```

packages/consumption/test/modules/attributes/AttributesController.test.ts
Milena-Czierlinski: In the context of Attribute successions, we use these abbreviations in many places. It is much less cluttered than writing e.g. successorOwnSharedIdentityAttributeParams.  I have a strong opinion against removing them in the test files.

packages/runtime/src/useCases/consumption/attributes/DeleteRepositoryAttribute.ts
```diff
@@ -0,0 +1,70 @@
...
+
+            ownSharedAttribute.shareInfo.sourceAttribute = undefined;
+            await this.attributeController.updateAttributeUnsafe(ownSharedAttribute);
+        }
+
+        if (typeof repositoryAttribute.succeededBy !== "undefined") {
+            const successor = await this.attributeController.getLocalAttribute(repositoryAttribute.succeededBy);
+            if (typeof successor === "undefined") {
+                throw new Error(`The Attribute ${repositoryAttribute.succeededBy} was not found, even though it is specified as successor of Attribute ${repositoryAttribute.id}.`);
+            }
```
jkoenig134, line 55: this should also be checked BEFORE changing any attributes to make sure to not corrupt the data

packages/consumption/test/modules/attributes/AttributesController.test.ts
jkoenig134: Yeah, I missed that. Has to be changed everywhere b/c this is VERY hard to process when you are not into the topic. Let's start by changing this in new code!

jkoenig134 (changes requested):

packages/runtime/test/consumption/attributes.test.ts
jkoenig134: IMO these results should be sorted so this should not be nessecary

packages/consumption/test/modules/attributes/AttributesController.test.ts
Milena-Czierlinski: I understand that we should try to use descriptive variable names. Nevertheless, we should aim at finding an appropriate balance between being expressive enough and keeping the code readable. So, I agree with you not to use abbreviations in the source code that can only be understood by someone who is deep into the topic. However, in tests where the context is known and where there are many repetitions, in my opinion it is legitimate to use abbreviations.
 
Also, I'd appreciate it not to use the imperative form in an ongoing discussion, in order to not create the impression that further arguments are not listened to.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -951,6 +1029,12 @@ export class AttributesController extends ConsumptionBaseController {
         }
 
         const ownSharedIdentityAttributes = await this.getLocalAttributes(query);
+        try {
+            ownSharedIdentityAttributes.sort((a, b) => (a.shareInfo!.peer < b.shareInfo!.peer ? 1 : -1));
+        } catch (e: unknown) {
+            throw CoreErrors.attributes.cannotSortAttributesByPeers(e);
+        }
```
jkoenig134, line 1036: sort can throw an error?

jkoenig134: ah, make sure to use ? instead of !

Milena-Czierlinski: This gives an error, since `a.shareInfo.peer` might be undefined.

packages/consumption/test/modules/attributes/AttributesController.test.ts
jkoenig134: > Also, I'd appreciate it not to use the imperative form in an ongoing discussion, in order to not create the impression that further arguments are not listened to.

👍🏼 

> However, in tests where the context is known and where there are many repetitions

You can not hope or expect that the reader has read ALL the code and understands everything about it. I have to read this variable and understand what it's used for. I don't want to have a glossary to understand tests..

packages/consumption/src/modules/attributes/AttributesController.ts
jkoenig134: yeah, you have to ignore if one is undefined. `!` throws errors so you only may use it when you are **totally** sure that it will not be undefined (e.g. when you have checked it before but the compiler doesn't knows that). In this case you are not sure so it throws an error. This should not happen here tbh.

jkoenig134: Also I meant to share the attributes by creation date instead of sorting it by the peer. IMO it doesn't make much sense to do this.

Milena-Czierlinski: Ok, I thought sorting them by peer would give some kind of consistency throughout the versions. But I don't mind sorting them be their creation date.

jkoenig134: maybe one other thing: this is one of the best examples where we should **not** use error codes. This is nothing the user has caused, just some internal stuff that's going wrong.

jkoenig134: You can use `expect(foo).toStrictEqual(expect.arrayContaining([bar, baz]));` to ignore the order. Please try this before.

Milena-Czierlinski: This seems to be working. So, are you fine the returned result is not sorted?

jkoenig134: yeah, sure.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -932,30 +1020,33 @@ export class AttributesController extends ConsumptionBaseController {
...
+        const ownSharedIdentityAttributeVersions = [...ownSharedIdentityAttributeSuccessors.reverse(), ...ownSharedIdentityAttributes, ...ownSharedIdentityAttributePredecessors];
+        return ownSharedIdentityAttributeVersions;
+    }
+
+    public async getSharedPredecessorsOfRepositoryAttribute(repositoryAttribute: LocalAttribute, query?: any): Promise<LocalAttribute[]> {
+        if (typeof query !== "undefined") {
+            query = JSON.parse(JSON.stringify(query));
+        } else {
+            query = {};
+        }
```
jkoenig134, line 1044: this can be written as `query ?= {}` or you just add {} as a default in the method params

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -964,9 +1055,34 @@ export class AttributesController extends ConsumptionBaseController {
...
-        return ownSharedIdentityAttributeVersions;
+        return ownSharedIdentityAttributePredecessors;
+    }
+
+    public async getSharedSuccessorsOfRepositoryAttribute(repositoryAttribute: LocalAttribute, query?: any): Promise<LocalAttribute[]> {
+        if (typeof query !== "undefined") {
+            query = JSON.parse(JSON.stringify(query));
+        } else {
+            query = {};
+        }
```
jkoenig134, line 1069: see above

packages/runtime/src/useCases/consumption/attributes/DeleteRepositoryAttribute.ts
```diff
@@ -0,0 +1,49 @@
...
+        @Inject private readonly attributesController: AttributesController,
+        @Inject private readonly accountController: AccountController,
+        @Inject validator: Validator
+    ) {
+        super(validator);
+    }
+
+    protected async executeInternal(request: DeleteRepositoryAttributeRequest): Promise<Result<void>> {
+        const repositoryAttributeId = CoreId.from(request.attributeId);
+        const repositoryAttribute = await this.attributesController.getLocalAttribute(repositoryAttributeId);
```
jkoenig134, line 28: can be inline
```suggestion
        const repositoryAttribute = await this.attributesController.getLocalAttribute(CoreId.from(request.attributeId));
```

jkoenig134 (changes requested):
only small comments left

packages/runtime/src/useCases/consumption/attributes/DeleteRepositoryAttribute.ts
```diff
@@ -0,0 +1,48 @@
...
+        if (!repositoryAttribute.isRepositoryAttribute(this.accountController.identity.address)) {
+            return Result.fail(RuntimeErrors.attributes.isNotRepositoryAttribute(CoreId.from(request.attributeId)));
+        }
+
+        const validationResult = await this.attributesController.validateSourceAttributeDeletion(repositoryAttribute);
+        if (validationResult.isError()) {
+            return Result.fail(validationResult.error);
+        }
+
+        await this.attributesController.deleteRepositoryAttribute(repositoryAttribute);
```
jkoenig134, line 42: as the actual method is now called `deleteRepositoryAttribute` why do we need all of the validations before? Shouldn't they be triggered inside the method?

jkoenig134 (changes requested):

packages/runtime/src/useCases/consumption/attributes/DeleteRepositoryAttribute.ts
Milena-Czierlinski: Yes, the validations are performed inside `deleteRepositoryAttribute`. I thought it would be nice to wrap the consumptions errors, according to https://github.com/nmshd/runtime/blob/main/packages/runtime/src/useCases/consumption/attributes/SucceedRepositoryAttribute.ts#L55 .

jkoenig134: IC, better safe than sorry :)

jkoenig134 (approved):

PR #60, opened 2024-03-11
Chore/Add deletionInfo to DataViewExpander

Milena-Czierlinski:


jkoenig134 (approved):

packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -1059,7 +1063,9 @@ export class DataViewExpander {
                     notificationReference: localAttribute.shareInfo.notificationReference?.toString(),
                     sourceAttribute: localAttribute.shareInfo.sourceAttribute.toString(),
                     tags: identityAttribute.tags ? identityAttribute.tags : [],
-                    valueType
+                    valueType,
+                    deletionStatus: localAttribute.deletionInfo?.deletionStatus,
+                    deletionDate: localAttribute.deletionInfo?.deletionDate.toString()
                 };
```
britsta, line 1069: Not really related to this PR, but could it be that the nomenclature of these DVOs needs to be adapted? For example, "OwnSharedIdentityAttributeDVO" instead of  "SharedToPeerAttributeDVO"?

jkoenig134: if you decide to do so - please don't do this during the deletion implementation.

jkoenig134: this will blow up the scope of the deletion

Milena-Czierlinski: @jkoenig134 If I recall correctly, @slavistan had the same comment in the renaming PR (mayor bump) and you answered, we shouldn't rename this, because it would have further complications. Or am I mistaken?

jkoenig134: Yeah, I have something related in my mind 😄

PR #61, opened 2024-03-11
fix/flaky tests

Magnus-Kuhn:
extend synchronisation time


packages/runtime/test/lib/testUtils.ts
```diff
@@ -61,9 +62,9 @@ export async function syncUntil(transportServices: TransportServices, until: (sy
 
         iterationNumber++;
         criteriaMet = until(finalSyncResult);
-    } while (!criteriaMet && iterationNumber <= 10);
+    } while (!criteriaMet && iterationNumber <= timeoutMs / iterationSleepMs);
 
-    if (!criteriaMet) throw new Error("syncUntil: the criteria specified in syncUntil were not fulfilled");
+    if (!criteriaMet) throw new Error(`syncUntil: the criteria specified in syncUntil were not fulfilled, iterations run: ${iterationNumber}`);
```
slavistan, line 67: The iterations are quite meaningless. Use the timeout instead: `` `syncUntil timed out after ${timeoutMs} ms.` ``.

slavistan: Does this fix all flaky tests or are we "just" making progress? :-)

Magnus-Kuhn: this has run successfully twice in all variants on my computer - so I'd say it's done

packages/runtime/test/lib/testUtils.ts
jkoenig134, line 1: Wouldn't the same fix to increase the iterations to 20 or more? This is quite unreadable as you now compare iterationNumber to the timeout

jkoenig134: in this case iteration number would be **160**, quite high.

packages/runtime/test/lib/testUtils.ts
jkoenig134: > on my computer 

but the pipeline failed :/

jkoenig134 (approved):

PR #63, opened 2024-03-12
Fix/fix demo bootstrapping due to wrong working dir

slavistan:
Make `npm run demo` work again!


slavistan:
No change to the code itself, didn't bump the version.

packages/iql/package.json
```diff
@@ -45,7 +45,8 @@
     },
     "devDependencies": {
         "light-server": "^2.9.1",
-        "peggy": "^4.0.2"
+        "peggy": "^4.0.2",
+        "webpack": "^5.90.3"
```
jkoenig134, line 49: Should be removed. We have webpack as a dep in the root project.

packages/iql/package.json
```diff
@@ -21,7 +21,7 @@
         "build:node": "npm run build",
         "build:parser": "peggy -o ./src/iql.gen.js ./src/grammar.peggy",
         "build:watch:parser": "npx nodemon -w src/grammar.peggy --exec 'npm run build:parser'",
-        "demo": "npm run build && light-server -s ./demo -w 'src/grammar.peggy, demo/app.js, demo/attributes.js, demo/index.html # cd demo && npm run build && webpack --mode production'",
+        "demo": "npm run build && npx webpack --mode production --config ./demo/webpack.config.js && light-server -s ./demo -w 'src/grammar.peggy, demo/app.js, demo/attributes.js, demo/index.html # npx webpack --mode production --config ./demo/webpack.config.js'",
```
jkoenig134, line 24: why is there a comment in this script?

packages/iql/package.json
slavistan: It's still a dev dependency for the iql package.

packages/iql/package.json
slavistan: Nah, that's the command `light-server` will run when any of the watched files change. That command is separated by a hash inside the command string.

packages/iql/package.json
jkoenig134: it isnt as we track such "global stuff" in the root package json, remove it please.

packages/iql/package.json
jkoenig134: nvm, it's the execution script for the light-server watch, right?

jkoenig134 (changes requested):
wrong dependency

package-lock.json
jkoenig134, line 1: npm i :)

jkoenig134 (changes requested):

jkoenig134 (approved):

PR #67, opened 2024-03-18
Refactor/update attribute types

jkoenig134:


packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -31,27 +31,30 @@ export interface ILocalAttribute extends ICoreSynchronizable {
     parentId?: ICoreId;
 }
 
-export type OwnSharedIdentityAttribute = LocalAttribute & { content: IdentityAttribute } & {
+export type OwnSharedIdentityAttribute = LocalAttribute & {
+    content: IdentityAttribute;
     shareInfo: LocalAttributeShareInfo & { sourceAttribute: CoreId };
```
slavistan, line 36: `sourceAttribute` has to be optional now that we've allowed to have own shared identity attributes without a link to their repository attribute. So `shareInfo: LocalAttributeShareInfo` does the job 👍

jkoenig134: I only simplified this, I don't want to change sth on the logic :)

slavistan: Ah, gotcha.

slavistan (approved):

PR #69, opened 2024-03-19
Chore/remove unused dev dep to types glob

sebbi08:
The dependency creates a version problem with min match when combined with the connector


jkoenig134 (approved):

jkoenig134:
If there is a problem you will need a verison bump.

sebbi08:
> If there is a problem you will need a verison bump.

Only in dev time when both are combined, so next normal release should be fine

PR #71, opened 2024-03-21
fix/test race condition

Magnus-Kuhn:
Decided to then have two methods called `exchangeMessageWithRequest` (one each for RequestModule active resp. inactive) since the inactive case is not expected to be used in any new tests. (This is also consistent with the inactive case methods not being exported anywhere except their file.)


jkoenig134 (approved):

jkoenig134:
Waiting for the status `DecisionRequired` in a method called `exchangeMessageWithRequest` does introduce an unpredictable side-effect to that method. #73 solves this without introducing side-effects.

PR #72, opened 2024-03-22
fix/PeerRelationshipTemplateLoadedEvent is triggered for own templates

jkoenig134:


jkoenig134:
@Magnus-Kuhn FYI, I downgraded FerretDB in this branch, you can just wait for this to be merged instead of doing the same in #71

tnotheis (approved):

PR #73, opened 2024-03-22
fix/test race conditions

jkoenig134:


Magnus-Kuhn:
that's the same thing as https://github.com/nmshd/runtime/pull/71, @jkoenig134 and I discussed that we'd use the other PR

jkoenig134:
@Magnus-Kuhn for now only the ones we know. ferretdb bump is undone, we have to find the solution to the introduced errors anywhere else.

Magnus-Kuhn (approved):

PR #74, opened 2024-03-22
Feature/Add ThirdPartyOwnedRelationshipAttributeSuccession

Milena-Czierlinski:
This will be necessary, if we aim for a correct succession, sharing the successor of an already shared third party RelationshipAttribute as a Response for a ReadAttributeRequestItem or ProposeAttributeRequestItem.

### DIFF
--- a/packages/consumption/src/consumption/CoreErrors.ts
+++ b/packages/consumption/src/consumption/CoreErrors.ts
@@ -85,6 +85,10 @@ class Attributes {
         return new CoreError("error.consumption.attributes.predecessorIsNotPeerSharedRelationshipAttribute", "Predecessor is not a peer shared relationship attribute.");
     }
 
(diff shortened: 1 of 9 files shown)


slavistan:
Is this a feature we really want to keep? I remember Thomas mentioning, that this was something that sounded good on paper in the beginning but that's not really used anywhere. Just want to make sure we're on the same page here.

Milena-Czierlinski:
> Is this a feature we really want to keep? I remember Thomas mentioning, that this was something that sounded good on paper in the beginning but that's not really used anywhere. Just want to make sure we're on the same page here.

Are you talking about the succession of ThirdPartyRelationshipAttributes (which I think is necessary if we have ThirdPartyRelationshipAttributes) or ThirdPartyRelationshipAttributes in general?

jkoenig134 (approved):

PR #75, opened 2024-03-25
fix/update event namespaces for succession events

jkoenig134:


packages/runtime/package.json
Milena-Czierlinski, line 1: What was our policy on bumping the runtime if we bump an underlying lib/workspace?

jkoenig134: I don't get your question. I want this update to ship so I also updated the runtime.

Milena-Czierlinski: I had in mind, that we bump the runtime version itself, too, if an underlying lib is bumped. Or do we only do that in special cases?

jkoenig134: lol, I was TOTALLY sure I bumped the runtime :D that's where my confusion came from.

PR #76, opened 2024-03-25
Feature/Add AcceptResponseItems

Milena-Czierlinski:


Milena-Czierlinski:
This introduces basically two new features (that were already reviewed in separate PRs):
1. Answering a ReadAttributeRequestItem or ProposeAttributeRequestItem with an Attribute already shared in the respective Relationship doesn't create new shared copies, but returns the ID of the already shared Attribute.
2. Answering a ReadAttributeRequestItem or ProposeAttributeRequestItem with the successor of an Attribute already shared in the respective Relationship performs an Attribute succession of the shared Copies.

britsta:
Why was the base branch first changed from `main` to `release/v5` and then from `release/v5` to `main` again?

Milena-Czierlinski:
> Why was the base branch first changed from `main` to `release/v5` and then from `release/v5` to `main` again?

Initially, I renamed `getSharedVersionsOfRepositoryAttribute` to `getSharedVersionsOfAttribute` here (which is a breaking change). To be able to merge it to main now, the use case `getSharedVersionsOfAttribute` is already added, but we also keep the old one marked as deprecated. I will create a PR for v5 that deletes that use case there, as its functionality is provided and extended by the new use case.

jkoenig134 (approved):

jkoenig134 (changes requested):
version bumps missing

jkoenig134 (approved):
LGTM

PR #78, opened 2024-03-26
Feature/backbone specific addresses

jkoenig134:


jkoenig134:
created a draft to hold this branch up2date

cc @tnotheis

jkoenig134:
The current occuring error ist caused by the backbone having some identity-problems, thinking it is prod.enmeshed.eu :D @tnotheis ??? :D

.vscode/settings.json
```diff
@@ -49,10 +49,7 @@
     "xmlTools.splitAttributesOnFormat": true,
     "xmlTools.enforcePrettySelfClosingTagOnFormat": true,
     "editor.codeActionsOnSave": {
-        "source.organizeImports": "explicit"
+        "source.organizeImports": "always"
```
jkoenig134, line 52: @sebbi08 this fixes our imports problems :(

packages/runtime/src/useCases/common/validation/ValidatableStrings.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * @pattern id1[A-Za-z0-9]{32,33}
+ * @pattern did:e:[a-zA-Z\.]*:dids:[0-9a-f]{22}
```
slavistan, line 2: `did:e:[A-Za-z0-9.-]+:dids:[0-9a-f]{22}`

The domain part should allow for dashes and digits (e.g. _did:e:preprod-0.enmeshed.eu_), and shouldn't be optional (right?). Also, inside a character set `[...]` the period character is interpreted literally, so you don't need to escape it.

jkoenig134: Good catch!

tnotheis (approved):

packages/transport/src/modules/accounts/IdentityUtil.ts
```diff
@@ -1,56 +1,53 @@
-import { CoreBuffer, CryptoHash, CryptoHashAlgorithm, ICryptoSignaturePublicKey } from "@nmshd/crypto";
-import { CoreAddress, CoreErrors } from "../../core";
+import { CoreBuffer, CryptoHash, CryptoHashAlgorithm, Encoding, ICryptoSignaturePublicKey } from "@nmshd/crypto";
+import { CoreAddress } from "../../core";
 
-export class IdentityUtil {
-    public static async createAddress(publicKey: ICryptoSignaturePublicKey, realm: string): Promise<CoreAddress> {
-        if (realm.length !== 3) throw CoreErrors.general.realmLength();
+const enmeshedPrefix = "did:e:";
```
slavistan, line 4: Vague name. How about `enmeshedAddressPrefix`?

jkoenig134: yeah, i also thought about inlining it completely. As it's only used in this file I don't care about the name that much. will change.

tnotheis (approved):

packages/transport/src/modules/accounts/IdentityUtil.ts
jkoenig134: I already changed the variable name :D

PR #80, opened 2024-03-26
Feature/AttributeSuccessionAcceptResponseItem for ReadAttributeRequestItemProcessor

Milena-Czierlinski:
The new AttributeSuccessionAcceptResponseItem takes care of a correct shared Attribute succession, if a ReadAttributeRequestItem is answered with the successor of an already shared Attribute. The same implementation will follow for the ProposeAttributeRequestItemProcessor.

## Summary by CodeRabbit

- **New Features**
	- Enhanced attribute comparison and traversal with new methods for determining predecessors and successors.
	- Improved event handling in request processing to include `Event` as a possible return type, enhancing system responsiveness and interactivity.

- **Bug Fixes**
	- Adjusted attribute acceptance logic to correctly check for the presence of a new attribute.

- **Tests**
	- Expanded testing for attribute comparison and request processing, ensuring robustness and reliability.


Milena-Czierlinski:
@coderabbitai review

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -53,21 +55,100 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+            });
         }
 
-        return ReadAttributeAcceptResponseItem.from({
-            result: ResponseItemResult.Accepted,
-            attributeId: sharedLocalAttribute.id,
-            attribute: sharedLocalAttribute.content
-        });
+        // this should never be reached
+        throw new Error("The ReadAttributeRequestItem must be either answered with an existing or a new Attribute.");
```
jkoenig134, line 151: you can just remove the else block instead of throwing this error (serval makes sure that this is valid)

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -53,21 +55,100 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+
+            const predecessorOwnSharedAttribute = latestSharedVersion[0];
+            if (typeof predecessorOwnSharedAttribute.shareInfo?.sourceAttribute === "undefined") {
+                throw new Error(`The Attribute ${predecessorOwnSharedAttribute.id} does not fulfill the requirements of an own shared Attribute.`);
+            }
+
+            if (predecessorOwnSharedAttribute.shareInfo.sourceAttribute.toString() === existingSourceAttribute.id.toString()) {
+                // return new AttributeAlreadySharedResponseItem
+                throw new Error("NotImplementedError");
+            }
```
jkoenig134, line 91: you should not  throw an error here. either remove this block or noop here until this is implemented.

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -53,21 +55,100 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+                    return AttributeSuccessionAcceptResponseItem.from({
+                        result: ResponseItemResult.Accepted,
+                        successorId: successorOwnSharedAttribute.id,
+                        successorContent: successorOwnSharedAttribute.content,
+                        predecessorId: predecessorOwnSharedAttribute.id
+                    });
+                }
+            } else {
+                throw new Error("You cannot share the predecessor of an already shared Attribute version.");
+            }
```
jkoenig134, line 140: shouldn't this occur in the validation?

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -53,21 +55,100 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+                    result: ResponseItemResult.Accepted,
+                    attributeId: sharedLocalAttribute.id,
+                    attribute: sharedLocalAttribute.content
+                });
+            }
+
+            const predecessorOwnSharedAttribute = latestSharedVersion[0];
+            if (typeof predecessorOwnSharedAttribute.shareInfo?.sourceAttribute === "undefined") {
+                throw new Error(`The Attribute ${predecessorOwnSharedAttribute.id} does not fulfill the requirements of an own shared Attribute.`);
+            }
```
jkoenig134, line 86: can we also validate this, so that we don't

when accepting?

jkoenig134 (changes requested):
I think you missed adding all this to the DataViewExpander

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
Milena-Czierlinski: Would you throw an error or return a failed validation result? This should never happen, since getSharedVersionsOfRepositoryAttribute queries for Attributes with a specific shareInfo.sourceAttribute, so it should never be undefined, but the compiler doesn't seem to know.

jkoenig134: ah, ok, then this is good as is! :)

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -78,6 +172,45 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+    private async performOwnSharedThirdPartyRelationshipAttributeSuccession(sharedPredecessorId: CoreId, sourceSuccessor: LocalAttribute, requestInfo: LocalRequestInfo) {
+        const successorParams = {
+            content: sourceSuccessor.content,
+            shareInfo: LocalAttributeShareInfo.from({
+                peer: requestInfo.peer,
+                requestReference: requestInfo.id,
+                sourceAttribute: sourceSuccessor.id
+            })
+        };
+        const { successor } = await this.consumptionController.attributes.succeedOwnSharedRelationshipAttribute(sharedPredecessorId, successorParams);
```
britsta, line 197: I understand the advantage of encapsulation, but it also seems a bit redundant to me to define the three functions `performOwnSharedIdentityAttributeSuccession`, `performOwnSharedThirdPartyRelationshipAttributeSuccession` and `performThirdPartyOwnedRelationshipAttributeSuccession`, which have essentially identical code, instead of calling the corresponding AttributesController.ts functions directly above. If the proposed refactoring takes place, consider instead defining a function `succeedOwnSharedThirdPartyRelationshipAttribute` in the AttributesController.ts that works like the `succeedOwnSharedRelationshipAttribute` function (so that it is still emphasised in which context the succession is performed).

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -99,19 +232,36 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+                    peer: requestInfo.peer,
+                    requestReference: requestInfo.id
+                })
+            });
+
+            if (responseItem.successorContent instanceof IdentityAttribute) {
+                const { predecessor, successor } = await this.consumptionController.attributes.succeedPeerSharedIdentityAttribute(responseItem.predecessorId, successorParams);
+                return new PeerSharedAttributeSucceededEvent(this.currentIdentityAddress.toString(), predecessor, successor);
+            }
+            await this.consumptionController.attributes.succeedThirdPartyOwnedRelationshipAttribute(responseItem.predecessorId, successorParams);
```
britsta, line 262: The RelationshipAttribute in question could also be owned by the Recipient of the Request. Therefore, I'd define a function `succeedPeerSharedThirdPartyRelationshipAttribute` function in the AttributesController.ts instead, that works like the succeedPeerSharedRelationshipAttribute function (so that it is still emphasised in which context the succession is performed) and that can be called here. From the perspective of the Sender of the Request, it does not play a role whether the ThirdPartyRelationshipAttribute is owned by the Recipient of the Request or a third party.

packages/runtime/test/consumption/iqlQuery.test.ts
```diff
@@ -45,17 +46,19 @@ describe("IQL Query", () => {
             }
         });
 
-        rLocalAttribute = response.value;
+        rLocalAttribute1 = response.value;
```
britsta, line 49: I think you can also write the value of the variable `response` directly here instead of defining an auxiliary variable if you do the same for `rLocalAttribute2`.

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -2109,10 +2191,8 @@ describe("AttributesController", function () {
                 }
             });
 
-            await TestUtil.expectThrowsAsync(
-                consumptionController.attributes.getSharedVersionsOfRepositoryAttribute(sharedIdentityAttribute.id),
-                "error.consumption.attributes.invalidPropertyValue"
-            );
+            const result = await consumptionController.attributes.getSharedVersionsOfRepositoryAttribute(sharedIdentityAttribute.id);
```
britsta, line 2194: Why should this function no longer throw an error if no RepositoryAttribute ID is used?

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -2133,10 +2213,8 @@ describe("AttributesController", function () {
                 }
             });
 
-            await TestUtil.expectThrowsAsync(
-                consumptionController.attributes.getSharedVersionsOfRepositoryAttribute(relationshipAttribute.id),
-                "error.consumption.attributes.invalidPropertyValue"
-            );
+            const result = await consumptionController.attributes.getSharedVersionsOfRepositoryAttribute(relationshipAttribute.id);
```
britsta, line 2216: Why should this function no longer throw an error if no RepositoryAttribute ID is used?

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -53,16 +75,88 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+                            predecessorId: predecessorOwnSharedAttribute.id
+                        });
+                    }
 
+                    const successorOwnSharedAttribute = await this.performThirdPartyOwnedRelationshipAttributeSuccession(
+                        predecessorOwnSharedAttribute.id,
+                        existingSourceAttribute,
+                        requestInfo
+                    );
+                    return AttributeSuccessionAcceptResponseItem.from({
```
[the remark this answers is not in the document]
Milena-Czierlinski: Regarding your third point: I came to the conclusion, that we don't need to be that specific. For the owner it doesn't make a difference, if it is an own shared RelationshipAttribute or an own shared ThirdPartyRelationshipAttribute. For the peer it is the same with peer shared RelationshipAttribute and peer shared ThirdPartyRelationshipAttribute. The only difference we have is if a RelationshipAttribute is owned by a third party. Also, I don't think the name `performThirdPartyOwnedThirdPartyRelationshipAttributeSuccession`, as it is super long and the only type of Attribute a third party can own is in fact a ThirdPartyRelationshipAttribute.

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
Milena-Czierlinski: I see your point. I did this to be closer to the original code, but maybe it doesn't help too much (besides explaining what we do by the means of the function name, instead of using a comma, which seems to be our coding style). Regarding your proposed `succeedOwnSharedThirdPartyRelationshipAttribute` I don't see the need to define a function that does exactly the same than an already existing function does (and as I explained above I also don't see the need to distinguish between own shared RelationshipAttributes and own shared ThirdPartyRelationshipAttributes).

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -78,6 +172,45 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+    private async performThirdPartyOwnedRelationshipAttributeSuccession(sharedPredecessorId: CoreId, sourceSuccessor: LocalAttribute, requestInfo: LocalRequestInfo) {
+        const successorParams = {
+            content: sourceSuccessor.content,
+            shareInfo: LocalAttributeShareInfo.from({
+                peer: requestInfo.peer,
+                requestReference: requestInfo.id,
+                sourceAttribute: sourceSuccessor.id
+            })
+        };
+        const { successor } = await this.consumptionController.attributes.succeedThirdPartyOwnedRelationshipAttribute(sharedPredecessorId, successorParams);
```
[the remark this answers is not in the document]
Milena-Czierlinski: see above

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
Milena-Czierlinski: The peers of a RelationshipAttribute always have the identical versions of that RelationshipAttribute. Succeeding a RelationshipAttribute always implies sharing the successor with the peer. Thus, a ReadAttributeRequestItem can never be answered with the successor of a RelationshipAttribute that wasn't shared before (except it is a ThirdPartyRelationshipAttribute).

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -53,16 +75,80 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+                            latestSharedAttribute.id,
+                            existingSourceAttribute,
+                            requestInfo
+                        );
+                        // return AttributeSuccessionAcceptResponseItem.from({
+                        //     result: ResponseItemResult.Accepted,
+                        //     successorId: successorOwnSharedAttribute.id,
+                        //     successorContent: successorOwnSharedAttribute.content,
+                        //     predecessorId: latestOwnSharedAttribute.id
+                        // });
```
jkoenig134, line 138: please remove the commented out code

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -959,16 +959,40 @@ export class AttributesController extends ConsumptionBaseController {
...
+        return false;
+    }
+
     public async getSharedVersionsOfRepositoryAttribute(id: CoreId, peers?: CoreAddress[], onlyLatestVersions = true): Promise<LocalAttribute[]> {
         let repositoryAttribute = await this.getLocalAttribute(id);
         if (typeof repositoryAttribute === "undefined") {
             throw TransportCoreErrors.general.recordNotFound(LocalAttribute, id.toString());
         }
 
-        if (!repositoryAttribute.isRepositoryAttribute(this.identity.address)) {
```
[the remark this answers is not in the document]
britsta: Alright! Apart from changing the name and refactoring, which will be done later, couldn't we already include a check at this point whether it is a RepositoryAttribute **or** a ThirdPartyRelationshipAttribute?

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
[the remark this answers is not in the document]
britsta: I agree that the name of the function would have a respectable length of 63 characters and that a RelationshipAttribute owned by a third party must be a ThirdPartyRelationshipAttribute. However, from a documentation perspective, for consistency with the existing nomenclature, I would still prefer the additional insertion of the five characters for 'Third'. But we don't have to decide that here- I'd like to discuss the nomenclature of the ThirdPartyRelationshipAttributes with you in general.

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
britsta: Yes, that's right, because of the extra definitions of the functions you can determine what they do by their names. My suggestion would probably only shift the problem with the redundant code sections to the AttributesController.ts, but not solve it if we still want to have descriptive names. Maybe you have another idea for this.

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
britsta: I have referred to a ThirdPartyRelationshipAttribute, which has as `shareInfo.sourceAttribute` a RelationshipAttribute from the context of the Relationship between the Recipient and a third party, which is owned by the Recipient. If Attribute succession for the RelationshipAttribute is performed between the Recipient and the third party, the Sender is not initially informed about this. However, the Recipient can send the successor in response to a ReadAttributeRequestItem from the Sender, so that Attribute succession can then be performed on the ThirdPartyRelationshipAttribute (which is owned by the Recipient). So I'm also talking about ThirdPartyRelationshipAttributes, but I wanted to emphasise that from the Sender's perspective, it doesn't matter whether it is owned by the Recipient or the third party, which is why I think it is necessary to rename the function.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -959,16 +959,40 @@ export class AttributesController extends ConsumptionBaseController {
         return attributeVersions;
     }
 
+    public async isAPredecessorOf(probedAttribute: LocalAttribute, referencedAttribute: LocalAttribute): Promise<boolean> {
+        while (referencedAttribute.succeeds) {
+            const predecessor = await this.getLocalAttribute(referencedAttribute.succeeds);
+            if (!predecessor) {
```
slavistan, line 965: `typeof predecessor !== "undefined"`

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -959,16 +959,40 @@ export class AttributesController extends ConsumptionBaseController {
...
+
+            referencedAttribute = predecessor;
+        }
+        return false;
+    }
+
+    public async isASuccessorOf(probedAttribute: LocalAttribute, referencedAttribute: LocalAttribute): Promise<boolean> {
+        while (referencedAttribute.succeededBy) {
+            const successor = await this.getLocalAttribute(referencedAttribute.succeededBy);
+            if (!successor) {
```
slavistan, line 979: `typeof successor !== "undefined"`

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -959,16 +959,40 @@ export class AttributesController extends ConsumptionBaseController {
         return attributeVersions;
     }
 
+    public async isAPredecessorOf(probedAttribute: LocalAttribute, referencedAttribute: LocalAttribute): Promise<boolean> {
+        while (referencedAttribute.succeeds) {
```
slavistan, line 963: `typeof ...`

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -959,16 +959,40 @@ export class AttributesController extends ConsumptionBaseController {
...
+
+            if (_.isEqual(predecessor, probedAttribute)) return true;
+
+            referencedAttribute = predecessor;
+        }
+        return false;
+    }
+
+    public async isASuccessorOf(probedAttribute: LocalAttribute, referencedAttribute: LocalAttribute): Promise<boolean> {
+        while (referencedAttribute.succeededBy) {
```
slavistan, line 977: `typeof ...`

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -959,16 +959,40 @@ export class AttributesController extends ConsumptionBaseController {
         return attributeVersions;
     }
 
+    public async isAPredecessorOf(probedAttribute: LocalAttribute, referencedAttribute: LocalAttribute): Promise<boolean> {
```
slavistan, line 962: I find the names `probedAttribute` and `referencedAttribute` hard to unterstand in terms of their order in the succession chain. Additionally, you could get rid of either `isAPredecessor` or `isASuccessor` altogether, as they're doing the same thing in reverse order.

How about you create a single function that does the job and give the arguments well-defined names:

```ts
public async isInSuccession(predecessor: LocalAttribute, successor: LocalAttribute): Promise<boolean> { /* ... */ }
```

Alternative names could be `isSequentialInSuccession` or something alike. Plus, as a final point, if you compare the attribute IDs instead of the attributes you can get rid of one hop of the traversal through the linked list.

packages/consumption/src/modules/requests/itemProcessors/readAttribute/AcceptReadAttributeRequestItemParameters.ts
```diff
@@ -31,7 +31,7 @@ export class AcceptReadAttributeRequestItemParameters extends Serializable {
         return typeof this.existingAttributeId !== "undefined";
     }
 
-    public isWithNewAttribute(): this is { newAttributeValue: IdentityAttribute | RelationshipAttribute } {
+    public isWithNewAttribute(): this is { newAttribute: IdentityAttribute | RelationshipAttribute } {
```
slavistan, line 34: How is the _is with new attribute_ to be parsed? As in "contains a new attribute"?

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -2109,10 +2191,8 @@ describe("AttributesController", function () {
                 }
             });
 
-            await TestUtil.expectThrowsAsync(
-                consumptionController.attributes.getSharedVersionsOfRepositoryAttribute(sharedIdentityAttribute.id),
-                "error.consumption.attributes.invalidPropertyValue"
-            );
+            const result = await consumptionController.attributes.getSharedVersionsOfRepositoryAttribute(sharedIdentityAttribute.id);
+            expect(result).toHaveLength(0);
```
slavistan, line 2195: This test expects a throw, judging by its name.

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -2133,10 +2213,8 @@ describe("AttributesController", function () {
                 }
             });
 
-            await TestUtil.expectThrowsAsync(
-                consumptionController.attributes.getSharedVersionsOfRepositoryAttribute(relationshipAttribute.id),
-                "error.consumption.attributes.invalidPropertyValue"
-            );
+            const result = await consumptionController.attributes.getSharedVersionsOfRepositoryAttribute(relationshipAttribute.id);
+            expect(result).toHaveLength(0);
```
slavistan, line 2217: This test "should throw", according to its name.

Help me out here. What changed so that the call to `getSharedVersionsOfRepositoryAttribute` doesn't raise an exception anymore?

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -36,14 +38,34 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+                }
+
+                const latestSharedVersionSourceAttribute = await this.consumptionController.attributes.getLocalAttribute(latestSharedVersion[0].shareInfo.sourceAttribute);
+                if (typeof latestSharedVersionSourceAttribute === "undefined") {
+                    throw new Error(`The Attribute ${latestSharedVersion[0].shareInfo.sourceAttribute} was not found.`);
+                }
+
+                if (await this.consumptionController.attributes.isASuccessorOf(latestSharedVersionSourceAttribute, foundAttribute)) {
+                    return ValidationResult.error(CoreErrors.requests.invalidRequestItem("You cannot share the predecessor of an already shared Attribute version."));
+                }
```
slavistan, line 67: Is there a reason you sometimes throw an error while at other times you return a `ValidationResult`?

packages/consumption/test/modules/attributes/AttributesController.test.ts
Milena-Czierlinski: Ah right, I need to rename it. We also allow other Attributes than RepositoryAttributes as input (function and use case will need renaming), since we didn't consider ThirdPartyRelationshipAttributes at the time of implementing it. Now, we need its functionality in the ReadAttributeRequestItemProcessor line 88, which should also work with ThirdPartyRelationshipAttributes in addition to RepositoryAttributes.

packages/consumption/src/modules/attributes/AttributesController.ts
Milena-Czierlinski: I vaguely see where this is coming from, but I fail to see how it simplifies things. Maybe let's have a chat about that.

packages/consumption/src/modules/requests/itemProcessors/readAttribute/AcceptReadAttributeRequestItemParameters.ts
Milena-Czierlinski: Accepting a ReadAttributeRequestItem, you either do so using the `AcceptReadAttributeRequestItemParametersWithExistingAttribute` or `AcceptReadAttributeRequestItemParametersWithNewAttribute`, depending on whether you want to share an existing Attribute (Repository/ThirdPartyRelationship) or you want to create a new one (own shared/peer shared pair). Hence, one of `existingAttributeId` or `newAttribute` of `AcceptReadAttributeRequestItemParameters` must be set. The method `isWithNewAttribute()` should return true, if `newAttribute` is set.

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
Milena-Czierlinski: I return a `ValidationResult`, if the user made a mistake (by entering a response that is not allowed), which we might expect and can handle. However, if there is something broken in the data which we do not expect to happen, I throw an error (so the user doesn't assume they made a mistake).

packages/consumption/src/modules/attributes/AttributesController.ts
slavistan: Sure, let's have a chat. I'll give you a call later.

Meanwhile, just to summarize:

>I fail to see how it simplifies things

- removing repetitive implementations => less duplication, fewer functions 
- unambiguous parameter names => better ease of use, less errors, a fulfilling life

packages/consumption/src/modules/attributes/AttributesController.ts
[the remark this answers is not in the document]
jkoenig134: > but I postponed that

you should do that **before** `feature/add-acceptresponseitems` is merged, then I'm ok with postponing :)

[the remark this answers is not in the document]
jkoenig134: in that case you should add a TODO in this PR so that it can't be forgotten in the other PR.

jkoenig134 (approved):

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -53,16 +75,64 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+                });
+            }
+
+            const latestSharedAttribute = latestSharedVersion[0];
+            if (typeof latestSharedAttribute.shareInfo?.sourceAttribute === "undefined") {
+                throw new Error(`The Attribute ${latestSharedAttribute.id} does not fulfill the requirements of an own shared Attribute.`);
+            }
+
+            if (latestSharedAttribute.shareInfo.sourceAttribute.toString() === existingSourceAttribute.id.toString()) {
+                // return new AttributeAlreadySharedResponseItem
```
slavistan, line 109: Something still missing here?

Milena-Czierlinski: Yes, this is what I will implement in another PR on this feature branch.

jkoenig134 (approved):

PR #81, opened 2024-03-27
Test/backbone mocking concept

Magnus-Kuhn:
demo.test.ts contains a small demo of msw intercepting an axios http request and returning a response with
* path wildcard
* path parameter
* using the request body
* typing the response body

No plan of merging this though.


tnotheis:
@Magnus-Kuhn What's the goal of the things you've done in this PR? Is it to replace the Backbone in our test runs? If so, should this only be a short-term solution, or are we planning to do it like this in the mid or even long term?

Magnus-Kuhn:
@tnotheis My impression was that we were looking for a long-term solution to not have the runtime wait for backbone development - I'm not sure if it should replace the backbone in all test runs or only if the backbone is not ready

tnotheis:
> @tnotheis My impression was that we were looking for a long-term solution to not have the runtime wait for backbone development - I'm not sure if it should replace the backbone in all test runs or only if the backbone is not ready

I asked because I am personally not a big fan of network layer fakes. I much prefer to have clean abstractions at the software level that can then be implemented by test doubles. 

For example, have an interface `IMessageClient` that can be implemented by `MessageClient` in production code, but also by some `FakeMessageClient` for testing. The biggest advantange of this approach over the "network fakes" is that we have full type safety.

Magnus-Kuhn:
> For example, have an interface `IMessageClient` that can be implemented by `MessageClient` in production code, but also by some `FakeMessageClient` for testing. The biggest advantange of this approach over the "network fakes" is that we have full type safety.

In this case, some parts of the `MessageClient` would not be tested - and the type safety could be achieved on the http response level

tnotheis:
> the type safety could be achieved on the http response level

Not really. If the method signature of the clients change for whatever reason, the compiler doesn't tell us to update the http fakes. This might result in failing tests that are hard to track down.

> In this case, some parts of the MessageClient would not be tested

You're saying this as if it was a bad thing. ;) I typically like to test my code in isolation, so that I know exactly what causes the error.
But perhaps I have misunderstood the whole thing. Is it for writing integration tests? If so, then I'm okay with it as it is, I think. But if it is meant for unit testing the transport/consumption controllers, then we should rather go with the approach I suggested in my original comment.

Magnus-Kuhn:
> Is it for writing integration tests?

The runtime tests are mostly end-to-end-tests and I haven't heard of this changing.

jkoenig134:
The main reason why we started this ticket was that we don't want to wait for the backbone to finish features - thats correct.

But actually I think Timo has a point.

1. we don't want to test axios, but we do this when we mock away the http calls
2. what code do we test in our http controllers? I don't think there's any logic we should test

> The runtime tests are mostly end-to-end-tests

Partly true, we call them dirty integration tests b/c we didn't had a way to mock away the backbone until now

> and I haven't heard of this approach changing (apart from this mock which is more born out of necessity)

actually the mock should fix this in addition, why should we wait for long backbone requests when we have great mocks?

Magnus-Kuhn (commented):
The module mock is an implementation of https://jestjs.io/docs/mock-functions#mocking-partials. Here, the mocked module is not directly imported, but a module using the mocked module is

jkoenig134:
@Magnus-Kuhn maybe we do not need jest.mock and can use fakes like the FakeSyncClient

Magnus-Kuhn:
> like the FakeSyncClient

What's that?

jkoenig134:
https://github.com/nmshd/runtime/blob/main/packages/transport/test/testHelpers/FakeSyncClient.ts

PR #82, opened 2024-04-01
Fix/Wrong field name newDatawalletVersion

Dannyps:
While working on the backbone performance tests, it came to my attention that the field `newDatawalletVersion` of the `FinalizeDatawalletVersionUpgradeResponse` class should actually be named `newDatawalletModificationIndex`, as per the linked backbone code below:

https://github.com/nmshd/backbone/blob/main/Modules/Synchronization/src/Synchronization.Application/SyncRuns/Commands/FinalizeSyncRun/FinalizeDatawalletVersionUpgradeSyncRunResponse.cs#L7


jkoenig134:
I don't think this is right, you changed the code for `FinalizeDatawalletVersionUpgrade` but showed the backbone code for `FinalizeExternalEventSync` which is another functionality and correctly implemented some lines below. @tnotheis could you have a look?

tnotheis:
Where do you see `FinalizeExternalEventSync`? 

The way I see it is: what sense would it make to return the new datawallet version as part of the `FinalizeDatawalletVersionUpgradeResponse`?

PR #88, opened 2024-04-05
Feature/delete account

jkoenig134:


packages/app-runtime/test/runtime/Offboarding.test.ts
Milena-Czierlinski, line 1: The tests are dependent of each other. Maybe it would be cleaner to combine the first three into a single test case and then have the other three followed by a common setup?

jkoenig134: in that specific case I see the whole test suite as one flow.

Milena-Czierlinski: I can understand that the way it is right now. However, I fear the next person to touch it, e.g. to add a test, will need to clean it up then. So, shouldn't we do it properly right away? I think it would be great if we did our best to avoid a large test refactoring like we just had in the runtime.

jkoenig134: okely schmokely :)

packages/app-runtime/test/runtime/Offboarding.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+        await services2.transportServices.account.syncDatawallet();
+        await services1.transportServices.account.syncDatawallet();
+    });
+
+    afterAll(async function () {
+        await runtime.stop();
+    });
+
+    test("delete account", async function () {
```
Milena-Czierlinski, line 38: Just one tiny thing: I'd like to check that there were 2 accounts before to make it really obvious that after calling deleteAccount one account is gone. 

Then it looks good to me.

jkoenig134: but then I'd test that onboard account creates one account... IMO that's out of scope in the offboarding

Milena-Czierlinski: okely schmokely :)

Milena-Czierlinski (approved):

PR #90, opened 2024-04-09
Refactor/adapt runtime to new relationships api

Magnus-Kuhn:


Magnus-Kuhn:
app-runtime has not been changed, this is a TODO

jkoenig134:
@Magnus-Kuhn the issues in the app runtime are blocking the pipelines, I'd suggest to update the runtime before any reviews.

jkoenig134:
@tnotheis as you know the new routes the best would you be willing to skim this PR?

Magnus-Kuhn:
the branch is poorly named, this is everything excluding the app-runtime - I just couldn't be bothered renaming it again

jkoenig134 (changes requested):
some ideas that could lead to bigger changes. I'll pause review for now

packages/transport/src/modules/relationships/local/CachedRelationship.ts
```diff
@@ -29,9 +35,9 @@ export class CachedRelationship extends CoreSerializable implements ICachedRelat
     @serialize()
     public lastMessageReceivedAt?: CoreDate;
 
-    public get creationChange(): RelationshipChange {
-        return this.changes[0];
-    }
+    @validate({ nullable: true })
+    @serialize()
+    public auditLog?: AuditLogEntry[];
```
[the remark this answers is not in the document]
jkoenig134: cc @tnotheis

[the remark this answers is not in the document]
tnotheis: Correct. It's never null and it always has at least one entry.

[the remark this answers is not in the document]
jkoenig134: great, then I got this right.

packages/content/src/relationships/RelationshipCreationRequestContent.ts
```diff
@@ -0,0 +1,27 @@
...
+    "@type": "RelationshipCreationRequestContent";
+    response: ResponseJSON;
+}
+
+export interface IRelationshipCreationRequestContent extends ISerializable {
+    response: IResponse;
+}
+
+@type("RelationshipCreationRequestContent")
+export class RelationshipCreationRequestContent extends Serializable implements IRelationshipCreationRequestContent {
```
jkoenig134, line 15: I think this should be called `RelationshipCreationContent` as there is no request anymore

packages/transport/src/core/backbone/RESTClientAuthenticate.ts
```diff
@@ -68,7 +68,7 @@ export class RESTClientAuthenticate extends RESTClient {
         });
     }
 
-    public override async put<T>(path: string, data: any, config: AxiosRequestConfig = {}): Promise<ClientResult<T>> {
+    public override async put<T>(path: string, data: any = {}, config: AxiosRequestConfig = {}): Promise<ClientResult<T>> {
```
jkoenig134, line 71: I'd vote for undoing this

packages/transport/src/modules/relationships/local/AuditLog.ts
```diff
@@ -0,0 +1,50 @@
...
+    @validate()
+    @serialize()
+    public newStatus: RelationshipStatus;
+
+    public static from(value: IAuditLogEntry): AuditLogEntry {
+        return this.fromAny({ ...value, oldStatus: value.oldStatus ?? undefined });
+    }
+}
+
+export class AuditLog {
```
jkoenig134, line 42: I'd vote for naming this RelationshipAuditLog (and RelationshipAuditLogEntry)

jkoenig134: don't forget to rename the file when changing

packages/transport/src/modules/relationships/local/AuditLog.ts
```diff
@@ -0,0 +1,50 @@
...
+    public static fromBackboneAuditLog(backboneAuditLog: BackboneAuditLog): AuditLogEntry[] {
+        const auditLog: AuditLogEntry[] = [];
+        backboneAuditLog.forEach((entry) =>
+            auditLog.push(AuditLogEntry.from({ ...entry, createdAt: CoreDate.from(entry.createdAt), createdBy: CoreAddress.from(entry.createdBy) }))
+        );
```
jkoenig134, line 47: why not using `.map()` ? Additionally `createdAt: CoreDate.from(entry.createdAt), createdBy: CoreAddress.from(entry.createdBy)` should not be nessecary, this is ts-servals job

packages/transport/src/modules/sync/externalEventProcessors/ExternalEventProcessorRegistry.ts
```diff
@@ -2,16 +2,15 @@ import { TransportError } from "../../../core";
...
 
 export class ExternalEventProcessorRegistry {
     private readonly processors = new Map<string, ExternalEventProcessorConstructor>();
     public constructor() {
         this.registerProcessor("MessageReceived", MessageReceivedExternalEventProcessor);
         this.registerProcessor("MessageDelivered", MessageDeliveredExternalEventProcessor);
-        this.registerProcessor("RelationshipChangeCreated", RelationshipChangeCreatedExternalEventProcessor);
-        this.registerProcessor("RelationshipChangeCompleted", RelationshipChangeCompletedExternalEventProcessor);
+        this.registerProcessor("RelationshipCreated", RelationshipCreatedOrChangedExternalEventProcessor);
+        this.registerProcessor("RelationshipStatusChanged", RelationshipCreatedOrChangedExternalEventProcessor);
```
jkoenig134, line 13: same processor for different events? could explode

packages/runtime/src/useCases/transport/relationships/GetRelationshipByAddress.ts
```diff
@@ -24,7 +24,7 @@ export class GetRelationshipByAddressUseCase extends UseCase<GetRelationshipByAd
     }
 
     protected async executeInternal(request: GetRelationshipByAddressRequest): Promise<Result<RelationshipDTO>> {
-        const relationship = await this.relationshipsController.getRelationshipToIdentity(CoreAddress.from(request.address));
+        const relationship = await this.relationshipsController.getRelationshipToIdentity(CoreAddress.from(request.address), undefined);
```
jkoenig134, line 27:
```suggestion
        const relationship = await this.relationshipsController.getRelationshipToIdentity(CoreAddress.from(request.address));
```

packages/transport/src/modules/relationships/RelationshipsController.ts
jkoenig134, line 1: I have some naming issues here. I think `Request` should be renamed to `CreationContent` on all places as there is no Request anymore when talking about relationships.

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -423,218 +335,141 @@ export class RelationshipsController extends TransportController {
...
     }
+    private async encryptAcceptanceContent(relationship: Relationship) {
+        const publicResponseCrypto = await this.secrets.getPublicResponse(relationship.relationshipSecretId);
 
-    private async encryptRevokeContent(relationship: Relationship, content: ICoreSerializable) {
-        const responseContent = RelationshipCreationChangeResponseContentWrapper.from({
+        const responseContent = RelationshipCreationResponseContentWrapper.from({
             relationshipId: relationship.id,
-            content: content
+            content: ""
```
jkoenig134, line 407: you removed the content, why is that?

jkoenig134 (changes requested):

packages/runtime/src/types/transport/RelationshipDTO.ts
```diff
@@ -11,11 +10,28 @@ export enum RelationshipStatus {
...
+
+export interface AuditLogDTO extends Array<AuditLogEntryDTO> {}
+
 export interface RelationshipDTO {
     id: string;
     template: RelationshipTemplateDTO;
     status: RelationshipStatus;
     peer: string;
     peerIdentity: IdentityDTO;
-    changes: RelationshipChangeDTO[];
```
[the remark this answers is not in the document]
Magnus-Kuhn: since the changes are not written in the runtime, the type extension can be pushed to where this actually happens - I think it's cleaner

packages/content/src/relationships/RelationshipCreationRequestContent.ts
Magnus-Kuhn: I prefer `RelationshipTemplateRequestsResponses` - that is the only thing it is used for

packages/runtime/src/types/transport/RelationshipDTO.ts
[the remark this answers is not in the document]
jkoenig134: wait what? 😆 The main point of DTOs is to prevent backwards incompatible changes.

you have to "mock" this changes array inside the dto and create it in the mapper. (only containing the creation change)

packages/transport/src/modules/sync/externalEventProcessors/ExternalEventProcessorRegistry.ts
Magnus-Kuhn: the edge case that a status change event arrives before its creation event means that there is some duplication going on - so I went all out on it :)

packages/transport/src/modules/relationships/RelationshipsController.ts
Magnus-Kuhn: to stick with the original plan of having no acceptanceContent

packages/transport/src/modules/sync/externalEventProcessors/ExternalEventProcessorRegistry.ts
jkoenig134: @tnotheis you should look into that

jkoenig134: shouldn't happen imo

packages/transport/src/modules/relationships/RelationshipsController.ts
jkoenig134: why don't you remove it everywhere then?

packages/content/src/relationships/RelationshipCreationRequestContent.ts
Magnus-Kuhn: I mean it's used to send the responses to the requests of a template

jkoenig134: but that's not where the name should come from. The name should indicate how and where it is transferred (see RelationshipTemplateContent)

Magnus-Kuhn: I expect confusion that the transport creationContent is the more generic Serializable

jkoenig134: but that's EXACTLY how this works. the content defined in content is our best guess what you can send in this content field. Btw - same with RelationshipTemplate.

packages/transport/src/modules/relationships/local/AuditLog.ts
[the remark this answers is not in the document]
Magnus-Kuhn: I would have preferred an AuditLog class instead of an AuditLogEntry class, but don't know how to do this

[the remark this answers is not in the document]
jkoenig134: that would break how we work with arrays and Serializables, please don't do that.

packages/transport/src/modules/relationships/RelationshipsController.ts
Magnus-Kuhn: I only wanted to keep as much as necessary for the secret transfer - did I miss something?

packages/transport/src/core/backbone/RESTClientAuthenticate.ts
Magnus-Kuhn: why?

packages/runtime/src/types/transport/RelationshipDTO.ts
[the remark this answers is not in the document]
Magnus-Kuhn: What about the DVO?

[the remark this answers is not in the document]
jkoenig134: DVOs are only used by the app and we can adapt to changes way easier there than the connector users.

packages/transport/src/core/backbone/RESTClientAuthenticate.ts
jkoenig134: b/c defaults here can lead to unexpected behaviour.

jkoenig134: just pass {} if you **really** want to send an empty object

packages/transport/src/modules/relationships/RelationshipsController.ts
jkoenig134: Maybe I miss something. We should talk about this.

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -137,8 +136,7 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
         }
 
         // checking for an active relationship is not secure as in the meantime the relationship could have been accepted
-        const isFromNewRelationship = parsedParams.responseSource instanceof RelationshipChange && parsedParams.responseSource.type === "Creation";
-
+        const isFromNewRelationship = parsedParams.responseSource instanceof Relationship;
```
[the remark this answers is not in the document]
Magnus-Kuhn: Alright, but I don't know how this can occur

packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -92,6 +92,11 @@ class RelationshipTemplates {
     }
 }
 
+class Relationships {
+    public faultyRelationshipAuditLog() {
+        return new ApplicationError("error.runtime.relationships.faultyRelationshipAuditLog", "The relationship status change sequence indicated by the audit log is incorrect.");
+    }
+}
 class Messages {
```
jkoenig134, line 100: missing newline

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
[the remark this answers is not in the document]
jkoenig134: wrong usage of the api

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -423,218 +335,141 @@ export class RelationshipsController extends TransportController {
...
-        relationship.status = backboneResponse.status;
-
-        await this.relationships.update(relationshipDoc, relationship);
-
-        this.eventBus.publish(new RelationshipChangedEvent(this.parent.identity.address.toString(), relationship));
-
-        return relationship;
+        return await this.updatePendingRelationshipWithPeerResponse(relationshipDoc);
     }
+    private async encryptAcceptanceContent(relationship: Relationship) {
```
jkoenig134, line 402: missing newline

packages/runtime/src/useCases/transport/relationships/RelationshipMapper.ts
```diff
@@ -19,39 +19,53 @@ export class RelationshipMapper {
...
+                        createdBy: dtoWithoutChanges.auditLog[0].createdBy,
+                        createdByDevice: "DVC00000000000000001",
+                        content: dtoWithoutChanges.creationContent
+                    },
+                    status: this.getStatus(dtoWithoutChanges.auditLog),
+                    type: RelationshipChangeType.Creation
+                }
+            ]
         };
     }
```
jkoenig134, line 41: why don't you do this inline?

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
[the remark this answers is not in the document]
jkoenig134: ah, wait, I think I misread the old check, could be an error here. I think we use the type "Relationship" **only** for new relationships and therefore only checking the type is safe, right?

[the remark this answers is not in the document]
Magnus-Kuhn: I agree - also I still am figuring out how to use VSCode for PRs, hence me not seeing what has changed

packages/runtime/src/types/transport/RelationshipDTO.ts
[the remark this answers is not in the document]
Magnus-Kuhn: and regarding backwards compatibility, the response source switching from relationship change to relationship is not an issue?

packages/transport/src/modules/relationships/local/AuditLog.ts
Magnus-Kuhn: I feel like enmeshed is suffering from overly long names enough - my vote is for keeping it like this until a different audit log pops up

packages/runtime/src/useCases/transport/relationships/RelationshipMapper.ts
Magnus-Kuhn: didn't want to have super long dot-chains, but I see they aren't that long anyway

packages/transport/src/modules/relationships/local/AuditLog.ts
Magnus-Kuhn: I didn't know about `.map` XD - and I get an error without `createdAt, createdBy`

jkoenig134: In that case I would argue just mapping everything by hand

Magnus-Kuhn: I don't understand what you mean

packages/transport/src/modules/relationships/local/AuditLog.ts
Magnus-Kuhn: does it need a type annotation though? I can't think of an instance where the audit log would not be inside its relationship

jkoenig134: @stnmtz has a strong opinion about Serializables without type annotationy, so yes - it does.

packages/transport/src/modules/relationships/local/AuditLog.ts
jkoenig134:
```
const auditLog = backboneAuditLog.map((entry) => {
            return AuditLogEntry.from({
                createdAt: CoreDate.from(entry.createdAt),
                createdBy: CoreAddress.from(entry.createdBy),
                createdByDevice: CoreId.from(entry.createdByDevice),
                reason: entry.reason,
                oldStatus: entry.oldStatus,
                newStatus: entry.newStatus
            });
        });

        return _.orderBy(auditLog, ["createdAt"], ["asc"]);
```

would be like this. btw - you missed `createdByDevice` in AuditLogEntry

packages/runtime/src/modules/RequestModule.ts
```diff
@@ -226,8 +217,8 @@ export class RequestModule extends RuntimeModule {
             return;
         }
 
-        const creationChangeContent = RelationshipCreationChangeRequestContent.from({ response: request.response!.content });
-        const createRelationshipResult = await services.transportServices.relationships.createRelationship({ templateId, content: creationChangeContent });
+        const creationContent = RelationshipCreationContent.from({ response: request.response!.content });
+        const createRelationshipResult = await services.transportServices.relationships.createRelationship({ templateId, creationContent: creationContent });
```
jkoenig134, line 221:
```suggestion
        const createRelationshipResult = await services.transportServices.relationships.createRelationship({ templateId, creationContent });
```

packages/runtime/src/useCases/transport/relationships/RelationshipMapper.ts
```diff
@@ -19,39 +19,53 @@ export class RelationshipMapper {
...
-            createdAt: change.createdAt.toString(),
-            content: change.content?.toJSON()
+            ...dtoWithoutChanges,
+            changes: [
+                {
+                    id: "RCH00000000000000001",
+                    request: {
+                        createdAt: dtoWithoutChanges.auditLog[0].createdAt,
+                        createdBy: dtoWithoutChanges.auditLog[0].createdBy,
+                        createdByDevice: "DVC00000000000000001",
```
jkoenig134, line 33:
```suggestion
                        createdAt: relationship.cache.auditLog[0].createdAt.toString(),
                        createdBy: relationship.cache.auditLog[0].createdBy.toString(),
                        createdByDevice: relationship.cache.auditLog[0].createdByDevice.toString(),
```

packages/transport/src/modules/relationships/local/RelationshipAuditLog.ts
```diff
@@ -0,0 +1,61 @@
...
+            return RelationshipAuditLogEntry.from({
+                createdAt: CoreDate.from(entry.createdAt),
+                createdBy: CoreAddress.from(entry.createdBy),
+                createdByDevice: CoreId.from(entry.createdByDevice),
+                reason: entry.reason,
+                oldStatus: entry.oldStatus,
+                newStatus: entry.newStatus
+            });
+        });
+        return _.orderBy(auditLog, ["createdAt"], ["asc"]);
```
jkoenig134, line 59: newline

jkoenig134 (changes requested):

packages/consumption/test/modules/requests/testHelpers/TestObjectFactory.ts
```diff
@@ -61,18 +58,13 @@ export class TestObjectFactory {
             cache:
                 properties?.cache ??
                 CachedRelationship.from({
-                    changes: [
+                    auditLog: [
```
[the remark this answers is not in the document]
Magnus-Kuhn: there is a good amount of cheating going on in this factory - there then also should be a cachedAt date. I also wanted to ask if cache/cachedAt should be set to mandatory - for relationships they are, but this would be inconsistent with other types

packages/consumption/test/modules/requests/testHelpers/TestObjectFactory.ts
```diff
@@ -64,6 +65,14 @@ export class TestObjectFactory {
...
                             reason: AuditLogEntryReason.Creation,
+                            newStatus: RelationshipStatus.Pending
+                        },
+
+                        {
+                            createdAt: CoreDate.from("2020-01-02T00:00:00.000Z"),
+                            createdBy: CoreAddress.from("id1"),
+                            createdByDevice: CoreId.from("DVC1"),
+                            reason: AuditLogEntryReason.AcceptanceOfCreation,
                             newStatus: RelationshipStatus.Active
```
jkoenig134, line 77: you missed the oldStatus

packages/transport/src/modules/relationships/local/CachedRelationship.ts
```diff
@@ -17,9 +19,13 @@ export class CachedRelationship extends CoreSerializable implements ICachedRelat
...
-    @validate()
-    @serialize({ type: RelationshipChange })
-    public changes: RelationshipChange[];
+    @validate({ nullable: true })
+    @serialize()
+    public creationContent?: Serializable;
+
+    @validate({ nullable: true })
+    @serialize()
+    public acceptanceContent?: Serializable;
```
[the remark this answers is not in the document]
Magnus-Kuhn: Agreed

packages/transport/src/modules/relationships/RelationshipsController.ts
jkoenig134: discussed - we do not send an custom acceptance content currently, but we need the acceptance content for key exchange.

packages/transport/src/modules/relationships/local/CachedRelationship.ts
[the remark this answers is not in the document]
jkoenig134: yep! As we do not use the acceptanceContent to transfer a custom content (only key material is transferred) we also don't save this in the cache.

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -87,13 +85,14 @@ export class RelationshipsController extends TransportController {
...
     @log()
-    private async updateCacheOfExistingRelationshipInDb(id: string, response?: BackboneGetRelationshipsResponse) {
+    private async updateExistingRelationshipInDb(id: string, response?: BackboneGetRelationshipsResponse) {
         const relationshipDoc = await this.relationships.read(id);
         if (!relationshipDoc) throw CoreErrors.general.recordNotFound(Relationship, id);
 
         const relationship = Relationship.from(relationshipDoc);
 
         await this.updateCacheOfRelationship(relationship, response);
+        relationship.status = response!.status;
```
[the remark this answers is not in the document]
Magnus-Kuhn: if there is no response, the recordNotFound error is thrown (response and relationshipDoc belong to the same relationship)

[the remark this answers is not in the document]
jkoenig134: not every function caller knows this. response is optional so

a) check for it before using it
b) make it required

packages/transport/src/modules/relationships/local/CachedRelationship.ts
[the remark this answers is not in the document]
tnotheis: Ah okay, makes sense.

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -87,12 +87,12 @@ export class RelationshipsController extends TransportController {
     @log()
     private async updateExistingRelationshipInDb(id: string, response?: BackboneGetRelationshipsResponse) {
         const relationshipDoc = await this.relationships.read(id);
-        if (!relationshipDoc) throw CoreErrors.general.recordNotFound(Relationship, id);
+        if (!response || !relationshipDoc) throw CoreErrors.general.recordNotFound(Relationship, id);
```
jkoenig134, line 90: this is no typecheck, if you do it like this you invalidate the usage of this method without a response

packages/transport/src/modules/relationships/RelationshipSecretController.ts
```diff
@@ -88,7 +88,7 @@ export class RelationshipSecretController extends SecretController {
     }
 
     @log()
-    public async getPublicResponse(relationshipSecretId: CoreId): Promise<CryptoRelationshipPublicResponse> {
+    public async getPublicAcceptanceContentCrypto(relationshipSecretId: CoreId): Promise<CryptoRelationshipPublicResponse> {
```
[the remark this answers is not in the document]
jkoenig134: the name should correctly show what the function does. I don't know what the crypto wants to say

packages/transport/src/modules/sync/externalEventProcessors/ExternalEventProcessorRegistry.ts
jkoenig134: discussed with @tnotheis. RelationshipCreated event will be removed. We leave it here for now.

packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -1641,9 +1598,8 @@ export class DataViewExpander {
             attributeMap: attributesByType,
             items: expandedAttributes,
             nameMap: stringByType,
-            changes: changes,
-            changeCount: changes.length,
-            templateId: relationship.template.id
+            templateId: relationship.template.id,
+            auditLog: relationship.auditLog
```
stnmtz, line 1602: I fear the auditLog might need to be expanded also? E.g. there are addresses within which could be expanded

packages/runtime/src/useCases/transport/relationships/RelationshipMapper.ts
```diff
@@ -19,39 +19,49 @@ export class RelationshipMapper {
                 publicKey: relationship.peer.publicKey.toBase64(false),
                 realm: relationship.peer.realm
             },
-            changes: relationship.cache.changes.map((c) => this.toRelationshipChangeDTO(c))
+            auditLog: relationship.cache.auditLog.map((entry) => this.toAuditLogEntryDTO(entry)),
+            creationContent: relationship.cache.creationContent?.toJSON(),
+            changes: [
```
stnmtz, line 24: why changes here?

packages/runtime/test/dataViews/RelationshipDVO.test.ts
```diff
@@ -101,30 +84,13 @@ describe("RelationshipDVO", () => {
...
-        expect(change.request.content).toBe(dto.changes[0].request.content);
-        expect(change.isOwn).toBe(true);
-        expect(change.response!.type).toBe("RelationshipChangeResponseDVO");
-        expect(change.response!.createdAt).toBe(dto.changes[0].response!.createdAt);
-        expect(change.response!.createdBy).toBe(dto.changes[0].response!.createdBy);
-        expect(change.response!.createdByDevice).toBe(dto.changes[0].response!.createdByDevice);
-        expect(change.response!.content).toBe(dto.changes[0].response!.content);
-        expect(change.date).toBe(dto.changes[0].response!.createdAt);
+
         expect(dvo.relationship!.templateId).toBe(dto.template.id);
```
stnmtz, line 94: As written above, I guess we need to expand the AuditLogEntries to AuditLogEntryDVOs and thus need to test it here also

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -221,299 +228,172 @@ export class RelationshipsController extends TransportController {
...
 
-        if (!template) throw CoreErrors.general.recordNotFound(RelationshipTemplate, templateId.toString());
-        if (!template.cache) throw this.newCacheEmptyError(RelationshipTemplate, template.id.toString());
-        if (!change.request.content) throw this.newEmptyOrInvalidContentError(change);
+        const serializedAcceptanceContent = acceptanceContent.serialize();
+        const buffer = CoreUtil.toBuffer(serializedAcceptanceContent);
 
-        const secretId = await TransportIds.relationshipSecret.generate();
-        const requestCipher = RelationshipCreationChangeRequestCipher.fromBase64(change.request.content);
-        await this.secrets.createTemplatorSecrets(secretId, template.cache, requestCipher.publicRequestCrypto);
```
stnmtz, line 501: This seems wrong, why is this removed?

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -523,118 +403,35 @@ export class RelationshipsController extends TransportController {
...
-            throw CoreErrors.general.recordNotFound(RelationshipChange, change.id.toString());
-        }
-
-        if (queriedChange.status !== RelationshipChangeStatus.Pending) {
-            throw CoreErrors.relationships.wrongChangeStatus(queriedChange.status);
-        }
-
-        let encryptedContent;
-        if (content) {
-            encryptedContent =
```
stnmtz, line 540: This seems wrong, why is this removed?

packages/transport/src/modules/relationships/RelationshipsController.ts
Magnus-Kuhn: what was removed?

packages/transport/src/modules/relationships/RelationshipsController.ts
Magnus-Kuhn: what exactly is removed?

packages/runtime/src/dataViews/DataViewExpander.ts
Magnus-Kuhn: IDK - don't know how exactly the DVO is used. ping @jkoenig134

packages/runtime/src/useCases/transport/relationships/RelationshipMapper.ts
Magnus-Kuhn: to be backwards-compatible

packages/runtime/test/lib/testUtils.ts
```diff
@@ -302,23 +291,21 @@ export async function establishRelationshipWithContents(
     transportServices1: TransportServices,
     transportServices2: TransportServices,
     templateContent: RelationshipTemplateContentJSON | RelationshipTemplateContent | IRelationshipTemplateContent,
-    requestContent: RelationshipCreationChangeRequestContentJSON | RelationshipCreationChangeRequestContent | IRelationshipCreationChangeRequestContent
+    requestContent: any
```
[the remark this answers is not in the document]
Magnus-Kuhn: in the transport library it's a serializable, and the idea is to take any and turn it into a JSONWrapper there if it's nothing special

packages/runtime/test/lib/testUtilsWithInactiveModules.ts
```diff
@@ -121,16 +121,16 @@ export async function exchangeTemplateAndReceiverSendsResponse(
...
-        const content = RelationshipCreationChangeRequestContent.from({ response: decidedRequest.response!.content as unknown as IResponse });
-        const result = await rRuntimeServices.transport.relationships.createRelationship({ content, templateId });
+        const creationContent = RelationshipCreationContent.from({ response: decidedRequest.response!.content as unknown as IResponse });
+        const result = await rRuntimeServices.transport.relationships.createRelationship({ creationContent, templateId });
 
         expect(result).toBeSuccessful();
 
         relationship = result.value;
-
         const rRelationshipChange = result.value.changes[0];
```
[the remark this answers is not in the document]
Magnus-Kuhn: for backwards-compatibility

packages/runtime/src/dataViews/DataViewExpander.ts
jkoenig134: yeah, would be great if we expand addresses in the audit log.

packages/transport/src/modules/relationships/local/RelationshipAuditLog.ts
```diff
@@ -0,0 +1,62 @@
...
+    @serialize()
+    public createdBy: CoreAddress;
+
+    @validate()
+    @serialize()
+    public createdByDevice: CoreId;
+
+    @validate()
+    @serialize()
+    public reason: AuditLogEntryReason;
```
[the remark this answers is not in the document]
Magnus-Kuhn: really? I looked for another enum to copy from and the DeviceType enum doesn't have a custom validator

packages/transport/test/modules/relationships/RelationshipsCustomContent.test.ts
```diff
@@ -33,18 +33,15 @@ describe("Relationships Custom Content", function () {
...
+        const relSenderRequest = relSender[0].cache!.creationContent;
 
-        expect(relRecipientRequest).toBeInstanceOf(RelationshipChangeRequest);
-        expect(relSenderRequest).toBeInstanceOf(RelationshipChangeRequest);
-
-        expect(relRecipientRequest.content).toBeInstanceOf(JSONWrapper);
-        const recipientToken = relRecipientRequest.content as JSONWrapper;
-        expect(relSenderRequest.content).toBeInstanceOf(JSONWrapper);
-        const senderToken = relSenderRequest.content as JSONWrapper;
+        expect(relRecipientContent).toBeInstanceOf(JSONWrapper);
```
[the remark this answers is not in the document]
Magnus-Kuhn: your suggestion is used in `expect((recipientToken.toJSON() as any).content).toBe("TestToken");`

packages/transport/src/modules/relationships/RelationshipSecretController.ts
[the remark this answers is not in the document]
Magnus-Kuhn: @stnmtz we wanted your opinion on this matter specifically - there is a naming conflict with the crypto library using request/response which are not a thing in transport anymore. Right now the conflict is happening in the SecretsController where I've made the changes, and the question is if there is a better alternative

packages/runtime/src/dataViews/DataViewExpander.ts
Magnus-Kuhn: which parts of the standard peer address expansion do you need? A full expansion would cause a loop peerAddress -> relationship -> peerAddress

jkoenig134:
Some things that should be changed before merging
- https://github.com/nmshd/backbone/pull/615 will remove the RelationshipCreated external event, please rename everything to RelationshipStatusChanged as you will get this too when the R was created.
- https://github.com/nmshd/backbone/pull/613 AcceptanceContent is now named CreationResponseContent

 I think all of this is available under the version `6.0.0-alpha.1` when #113 is merged it will be really easy for you to upgrade the local/test backbone to this version.

jkoenig134:
Additionally - @stnmtz what do you think about the acceptance content, should we re-add it now, or add it in the future if something requires it?

packages/transport/src/modules/sync/externalEventProcessors/RelationshipCreatedOrChangedExternalEventProcessor.ts
```diff
@@ -3,10 +3,10 @@ import { Relationship } from "../../relationships/local/Relationship";
 import { BackboneExternalEvent } from "../backbone/BackboneExternalEvent";
 import { ExternalEventProcessor } from "./ExternalEventProcessor";
 
-export class RelationshipChangeCompletedExternalEventProcessor extends ExternalEventProcessor {
+export class RelationshipCreatedOrChangedExternalEventProcessor extends ExternalEventProcessor {
     public override async execute(externalEvent: BackboneExternalEvent): Promise<Relationship | undefined> {
-        const payload = externalEvent.payload as { changeId: string };
-        const relationship = await this.accountController.relationships.applyChangeById(payload.changeId);
+        const payload = externalEvent.payload as { relationshipId: string };
+        const relationship = await this.accountController.relationships.applyIncomingEvent(payload.relationshipId);
```
jkoenig134, line 9: I think we should rename `applyIncomingEvent` to `applyRelationshipStatusChangedEvent`

tnotheis:
> @tnotheis would you please ping us when `6.0.0-alpha.2` is published?

ping

The image with version `6.0.0-alpha.2` does now exist.

jkoenig134 (dismissed):

tnotheis (dismissed):

packages/transport/src/modules/relationships/RelationshipsController.ts
jkoenig134: I think it's only re-positioned and renamed, that's thy GH shows a big diff.

packages/runtime/test/lib/testUtils.ts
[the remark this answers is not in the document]
jkoenig134: yeah, but here we don't want to use JSONWrappers ..

packages/runtime/test/dataViews/RelationshipDVO.test.ts
jkoenig134: I'd postpone this to 
1. another PR
2. when we need it

packages/runtime/src/dataViews/DataViewExpander.ts
jkoenig134: as below, can be postponed. We don't need it in the App.

packages/transport/test/modules/relationships/RelationshipsCustomContent.test.ts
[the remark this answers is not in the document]
jkoenig134: @Magnus-Kuhn is right, we don't have the wrapper "Request" anymore, so the content itself is of type JSONWrapper

jkoenig134 (dismissed):

jkoenig134 (dismissed):

jkoenig134 (dismissed):

packages/transport/src/modules/relationships/backbone/BackboneRelationship.ts
```diff
@@ -0,0 +1,15 @@
...
+export interface BackboneRelationship {
+    id: string;
+    relationshipTemplateId: string;
+    from: string;
+    to: string;
+
+    createdAt: string;
+    status: RelationshipStatus;
+    auditLog: BackboneRelationshipAuditLog;
+    creationContent?: string;
```
stnmtz, line 13: May not be optional in Backbone

jkoenig134 (approved):

stnmtz (approved):

PR #93, opened 2024-04-09
Feature/Return AcceptResponseItem if attribute to delete doesn't exit

Milena-Czierlinski:
This might occur if the Attribute, that is requested to be deleted, was deleted after receiving the Request and before responding to it.


packages/consumption/src/modules/requests/itemProcessors/deleteAttribute/DeleteAttributeRequestItemProcessor.ts
```diff
@@ -40,7 +40,16 @@ export class DeleteAttributeRequestItemProcessor extends GenericRequestItemProce
         return ValidationResult.success();
     }
 
-    public override canAccept(_requestItem: DeleteAttributeRequestItem, params: AcceptDeleteAttributeRequestItemParametersJSON, _requestInfo: LocalRequestInfo): ValidationResult {
+    public override async canAccept(
+        requestItem: DeleteAttributeRequestItem,
+        params: AcceptDeleteAttributeRequestItemParametersJSON,
+        _requestInfo: LocalRequestInfo
+    ): Promise<ValidationResult> {
```
britsta, line 47: Perhaps we should still expect the `deletionDate` entered to be a valid one, which means inserting the check
```
const deletionDate = CoreDate.from(params.deletionDate);        
if (!deletionDate.dateTime.isValid) {
       return ValidationResult.error(CoreErrors.requests.invalidAcceptParameters("The deletionDate is invalid."));
} 
```
before the check
```
const attribute = await this.consumptionController.attributes.getLocalAttribute(requestItem.attributeId);
if (typeof attribute === "undefined") {
       return ValidationResult.success();
}
```

Milena-Czierlinski: Since we don't process it, I think it could be useful to also allow for free texts (see test).

jkoenig134 (approved):

PR #99, opened 2024-04-11
Feature/Detach ThirdPartyRelationshipAttributes

Milena-Czierlinski:
Deleting a RelationshipAttribute, that is the sourceAttribute of a ThirdPartyRelationshipAttribute, their respective sourceAttribute property will be unset.

### DIFF
--- a/packages/consumption/src/modules/attributes/AttributesController.ts
+++ b/packages/consumption/src/modules/attributes/AttributesController.ts
@@ -878,37 +878,37 @@ export class AttributesController extends ConsumptionBaseController {
         await this.attributes.delete({ id: id });
     }
 
-    public async deleteRepositoryAttribute(sourceAttribute: LocalAttribute): Promise<void> {
-        const validationResult = await this.validateSourceAttributeDeletion(sourceAttribute);
+    public async executeFullAttributeDeletionProcess(attribute: LocalAttribute): Promise<void> {
+        const validationResult = await this.validateFullAttributeDeletionProcess(attribute);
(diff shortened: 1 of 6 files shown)


packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -878,37 +878,37 @@ export class AttributesController extends ConsumptionBaseController {
         await this.attributes.delete({ id: id });
     }
 
-    public async deleteRepositoryAttribute(sourceAttribute: LocalAttribute): Promise<void> {
-        const validationResult = await this.validateSourceAttributeDeletion(sourceAttribute);
+    public async executeFullAttributeDeletionProcess(attribute: LocalAttribute): Promise<void> {
```
jkoenig134, line 881: 1. is there a "not full" attribute deletion process?
2. why don't we simply call it deleteAttribute? I think we can abstract away from the user that there is triggered a "process".

jkoenig134 (changes requested):
One small thing that confuses me, rest LGTM.

packages/consumption/src/modules/attributes/AttributesController.ts
Milena-Czierlinski: We already have a deleteAttribute function, which "only" deletes the Attribute with respective child Attributes (and fires an event). In contrast, this "full process" also deletes the predecessors, detaches all respective shared versions, and detaches the successor.

jkoenig134 (approved):

PR #100, opened 2024-04-12
Feature/Delete third party owned RelationshipAttribute use case

Milena-Czierlinski:
The thirdPartyOwned adjustments in the LocalAttribute were copied from #76 .
The ThirdPartyOwnedRelationshipAttributeDeletedByPeerNotificationItemProcessor does the same like the PeerSharedAttributeDeletedByPeerNotificationItemProcessor, but @britsta and I discussed that it still makes sense to have a dedicated ThirdPartyOwnedRelationshipAttributeDeletedByPeerNotificationItem, to be more precise and consistent with the naming (which will also be applied in the documentation), and to be flexible if we want to adjust the handling of the deletion of PeerSharedAttributes compared to ThirdPartyOwnedRelationshipAttributes.

### DIFF
--- a/packages/consumption/src/modules/attributes/local/LocalAttribute.ts
+++ b/packages/consumption/src/modules/attributes/local/LocalAttribute.ts
@@ -134,9 +134,6 @@ export class LocalAttribute extends CoreSynchronizable implements ILocalAttribut
 
     public isOwnSharedAttribute(ownAddress: CoreAddress, peerAddress?: CoreAddress): this is OwnSharedIdentityAttribute | OwnSharedRelationshipAttribute {
         let isOwnSharedAttribute = this.isShared() && this.isOwnedBy(ownAddress);
-        if (!isOwnSharedAttribute) {
-            return isOwnSharedAttribute;
-        }
 
         if (typeof peerAddress !== "undefined") {
             isOwnSharedAttribute &&= this.shareInfo!.peer.equals(peerAddress);
@@ -146,13 +143,8 @@ export class LocalAttribute extends CoreSynchronizable implements ILocalAttribut
 
     public isPeerSharedAttribute(peerAddress?: CoreAddress): this is PeerSharedIdentityAttribute | PeerSharedRelationshipAttribute {
         let isPeerSharedAttribute = this.isShared() && this.isOwnedBy(this.shareInfo.peer);
-        if (!isPeerSharedAttribute) {
-            return isPeerSharedAttribute;
-        }
 
-        if (this.isIdentityAttribute()) {
-            isPeerSharedAttribute &&= typeof this.shareInfo!.sourceAttribute === "undefined";
-        }
+        isPeerSharedAttribute &&= typeof this.shareInfo!.sourceAttribute === "undefined";
 
         if (typeof peerAddress !== "undefined") {
             isPeerSharedAttribute &&= this.isOwnedBy(peerAddress);
@@ -162,9 +154,6 @@ export class LocalAttribute extends CoreSynchronizable implements ILocalAttribut
 
     public isThirdPartyOwnedAttribute(ownAddress: CoreAddress, thirdPartyAddress?: CoreAddress): this is ThirdPartyOwnedRelationshipAttribute {
         let isThirdPartyOwnedAttribute = this.isShared() && !this.isOwnedBy(ownAddress) && !this.isOwnedBy(this.shareInfo.peer);
-        if (!isThirdPartyOwnedAttribute) {
-            return isThirdPartyOwnedAttribute;
-        }
 
         if (typeof thirdPartyAddress !== "undefined") {
             isThirdPartyOwnedAttribute &&= this.isOwnedBy(thirdPartyAddress);
--- a/packages/consumption/src/modules/attributes/local/LocalAttribute.ts
+++ b/packages/consumption/src/modules/attributes/local/LocalAttribute.ts
@@ -57,7 +57,7 @@ export type PeerSharedRelationshipAttribute = LocalAttribute & {
 
 export type ThirdPartyOwnedRelationshipAttribute = LocalAttribute & {
     content: RelationshipAttribute;
-    shareInfo: LocalAttributeShareInfo & { sourceAttribute: CoreId };
+    shareInfo: LocalAttributeShareInfo;
 };
 
 export type RepositoryAttribute = LocalAttribute & {
--- a/packages/consumption/src/modules/attributes/local/LocalAttribute.ts
+++ b/packages/consumption/src/modules/attributes/local/LocalAttribute.ts
@@ -55,6 +55,11 @@ export type PeerSharedRelationshipAttribute = LocalAttribute & {
     shareInfo: LocalAttributeShareInfo & { sourceAttribute: undefined };
 };
 
+export type ThirdPartyOwnedRelationshipAttribute = LocalAttribute & {
+    content: RelationshipAttribute;
+    shareInfo: LocalAttributeShareInfo & { sourceAttribute: CoreId };
+};
+
 export type RepositoryAttribute = LocalAttribute & {
     content: IdentityAttribute;
     shareInfo: undefined;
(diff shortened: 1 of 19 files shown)


jkoenig134 (approved):

packages/consumption/test/modules/notifications/itemProcessors/attributeDeleted/ThirdPartyOwnedRelationshipAttributeDeletedByPeerNotificationItemProcessor.test.ts
```diff
@@ -0,0 +1,203 @@
...
+        /* Manually trigger and verify rollback. */
+        const rollbackResult = await processor.rollback(notificationItem, notification);
+        expect(rollbackResult).toBeUndefined();
+    });
+
+    test("detects spoofing attempts", async function () {
+        /* A naughty peer is trying to delete attributes shared
+         * not with him, but with another peer. This must be
+         * caught by the validation. */
+        const differentThirdPartyOwnedRelationshipAttribute = await consumptionController.attributes.createAttributeUnsafe({
```
britsta, line 159: IMO the word `different` can be omitted when naming the variable `differentThirdPartyOwnedRelationshipAttribute.`, as only one ThirdPartyOwnedRelationshipAttribute plays a role in this test.

packages/consumption/test/modules/notifications/itemProcessors/attributeDeleted/ThirdPartyOwnedRelationshipAttributeDeletedByPeerNotificationItemProcessor.test.ts
```diff
@@ -0,0 +1,203 @@
...
+                value: {
+                    "@type": "ProprietaryString",
+                    value: "0815",
+                    title: "Customer ID"
+                },
+                owner: CoreAddress.from("thirdParty"),
+                confidentiality: RelationshipAttributeConfidentiality.Public
+            }),
+            shareInfo: {
+                peer: CoreAddress.from("otherPeer"),
```
britsta, line 171: IMO the peer `"otherPeer"` can be renamed just as `"peer"` when defining the property `shareInfo.peer`, as only the peer `"naughtyPeer"` and this peer here play a role in this test.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -1756,4 +1758,105 @@ describe("DeleteAttributeUseCases", () => {
...
+
+    describe(DeleteThirdPartyOwnedRelationshipAttributeAndNotifyPeerUseCase.name, () => {
+        let peerSharedRelationshipAttribute: LocalAttributeDTO;
+        let thirdPartyOwnedRelationshipAttribute: LocalAttributeDTO;
+        beforeEach(async () => {
+            peerSharedRelationshipAttribute = await executeFullCreateAndShareRelationshipAttributeFlow(services3, services1, {
+                content: {
+                    value: {
+                        "@type": "ProprietaryString",
+                        value: "My amazing string",
```
britsta, line 1770: I have never seen such an amazing string!

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -1756,4 +1758,105 @@ describe("DeleteAttributeUseCases", () => {
...
+
+            thirdPartyOwnedRelationshipAttribute = await executeFullRequestAndShareThirdPartyRelationshipAttributeFlow(
+                services1,
+                services2,
+                requestParams,
+                peerSharedRelationshipAttribute.id
+            );
+        });
+
+        test("should delete a third party owned RelationshipAttribute as the sender", async () => {
```
britsta, line 1798: IMHO: Normally, we refer to the Sender of a Request as the Sender and the Recipient of a Request as the Recipient. Here the designations are used the other way round, because we refer to the Attributes instead of the Request used, which makes sense because it is about the Attributes and not the Request here. However, I'd reformulate to "should delete a third party owned RelationshipAttribute as the sender **of it**" or try to find a synonym for the Sender to avoid confusion. 

Same for the tests below (with Recipient or Sender).

packages/runtime/test/consumption/attributes.test.ts
jkoenig134: actually we use descriptive names for strings like this (in this case `aProprietaryString` should be used).

packages/consumption/test/modules/notifications/itemProcessors/attributeDeleted/ThirdPartyOwnedRelationshipAttributeDeletedByPeerNotificationItemProcessor.test.ts
Milena-Czierlinski: The variable name is chosen to stress the fact, that it is a ThirdPartyRelationshipAttribute in the context of a different Relationship (as you have to look thoroughly for the difference and we don't use comments).

packages/consumption/test/modules/notifications/itemProcessors/attributeDeleted/ThirdPartyOwnedRelationshipAttributeDeletedByPeerNotificationItemProcessor.test.ts
Milena-Czierlinski: The variable name is chosen to stress the fact, that it is a ThirdPartyRelationshipAttribute in the context of a different Relationship (as you have to look thoroughly for the difference and we don't use comments).

packages/runtime/test/consumption/attributes.test.ts
Milena-Czierlinski: :(

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -151,14 +160,26 @@ export class LocalAttribute extends CoreSynchronizable implements ILocalAttribut
         return isPeerSharedAttribute;
     }
 
+    public isThirdPartyOwnedAttribute(ownAddress: CoreAddress, thirdPartyAddress?: CoreAddress): this is ThirdPartyOwnedRelationshipAttribute {
+        let isThirdPartyOwnedAttribute = this.isShared() && !this.isOwnedBy(ownAddress) && !this.isOwnedBy(this.shareInfo.peer);
+        if (!isThirdPartyOwnedAttribute) {
```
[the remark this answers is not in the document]
Milena-Czierlinski: This is implemented according to `isOwnSharedAttribute` and `isPeerSharedAttribute`. I think the idea is that we can return early, if the requirements aren't met in the first place. I doubt this saves much time/resources, however, it seems to be our coding style.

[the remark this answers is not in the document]
britsta: It is not clear to me why this should return earlier, because in both implementation variants you would have to perform at least one check (in one case `if (!isThirdPartyOwnedAttribute)` and in the other case `if (typeof thirdPartyAddress !== "undefined")`). If `this.isShared() && !this.isOwnedBy(ownAddress) && !this.isOwnedBy(this.shareInfo.peer)` is `true`, two checks would currently have to be made, which is one unneccessary one. Only if `this.isShared() && !this.isOwnedBy(ownAddress) && !this.isOwnedBy(this.shareInfo.peer)` is `false` and `thirdPartyAddress` is defined, the variable `isThirdPartyOwnedAttribute` is currently prevented from being updated unnecessarily (from `false` to `false`). 

But you're right that this doesn't save much time/resources. And if it's our coding style, it doesn't need to be adapted here either.

packages/consumption/test/modules/notifications/itemProcessors/attributeDeleted/ThirdPartyOwnedRelationshipAttributeDeletedByPeerNotificationItemProcessor.test.ts
britsta: All right, maybe you can think of an adjective to emphasize this more, because it wasn't quite clear to me, but if not it is good like this 👍

britsta (approved):
LGTM

PR #101, opened 2024-04-15
Feature/AttributeSuccessionAcceptResponseItem for ProposeAttributeRequestItemProcessor

Milena-Czierlinski:
Basically, what has already been implemented for the ReadAttributeRequestItemProcessor.

### DIFF
--- a/packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
+++ b/packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
@@ -1,4 +1,5 @@
 import {
+    AttributeSuccessionAcceptResponseItem,
     IdentityAttribute,
     ProposeAttributeAcceptResponseItem,
     ProposeAttributeRequestItem,
@@ -10,6 +11,7 @@ import {
 } from "@nmshd/content";
 import { CoreAddress, CoreId, CoreErrors as TransportCoreErrors } from "@nmshd/transport";
 import { CoreErrors } from "../../../../consumption/CoreErrors";
+import { AttributeSuccessorParams, LocalAttributeShareInfo, PeerSharedAttributeSucceededEvent } from "../../../attributes";
 import { LocalAttribute } from "../../../attributes/local/LocalAttribute";
 import { ValidationResult } from "../../../common/ValidationResult";
 import { GenericRequestItemProcessor } from "../GenericRequestItemProcessor";
@@ -71,19 +73,35 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
         let attribute = parsedParams.attribute;
 
         if (parsedParams.isWithExistingAttribute()) {
-            const localAttribute = await this.consumptionController.attributes.getLocalAttribute(parsedParams.attributeId);
+            const foundAttribute = await this.consumptionController.attributes.getLocalAttribute(parsedParams.attributeId);
 
-            if (!localAttribute) {
+            if (typeof foundAttribute === "undefined") {
                 return ValidationResult.error(TransportCoreErrors.general.recordNotFound(LocalAttribute, requestInfo.id.toString()));
             }
 
-            attribute = localAttribute.content;
+            const latestSharedVersion = await this.consumptionController.attributes.getSharedVersionsOfRepositoryAttribute(parsedParams.attributeId, [requestInfo.peer], true);
+            if (latestSharedVersion.length > 0) {
+                if (typeof latestSharedVersion[0].shareInfo?.sourceAttribute === "undefined") {
+                    throw new Error(`The Attribute ${latestSharedVersion[0].id} does not fulfill the requirements of an own shared Attribute.`);
+                }
+
+                const latestSharedVersionSourceAttribute = await this.consumptionController.attributes.getLocalAttribute(latestSharedVersion[0].shareInfo.sourceAttribute);
+                if (typeof latestSharedVersionSourceAttribute === "undefined") {
+                    throw new Error(`The Attribute ${latestSharedVersion[0].shareInfo.sourceAttribute} was not found.`);
+                }
+
+                if (await this.consumptionController.attributes.isSubsequentInSuccession(foundAttribute, latestSharedVersionSourceAttribute)) {
+                    return ValidationResult.error(CoreErrors.requests.invalidAcceptParameters("You cannot share the predecessor of an already shared Attribute version."));
+                }
+            }
+
+            attribute = foundAttribute.content;
         }
 
         const ownerIsEmpty = attribute!.owner.equals("");
         const ownerIsCurrentIdentity = attribute!.owner.equals(this.currentIdentityAddress);
         if (!ownerIsEmpty && !ownerIsCurrentIdentity) {
-            return ValidationResult.error(CoreErrors.requests.invalidRequestItem("The given Attribute belongs to someone else. You can only share own Attributes."));
+            return ValidationResult.error(CoreErrors.requests.invalidAcceptParameters("The given Attribute belongs to someone else. You can only share own Attributes."));
         }
 
         return ValidationResult.success();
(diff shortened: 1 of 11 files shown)


packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
```diff
@@ -93,15 +111,60 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
...
-            sharedLocalAttribute = await this.createNewAttribute(parsedParams.attribute!, requestInfo);
+            const existingSourceAttribute = await this.consumptionController.attributes.getLocalAttribute(parsedParams.attributeId);
+            if (typeof existingSourceAttribute === "undefined") {
+                throw TransportCoreErrors.general.recordNotFound(LocalAttribute, parsedParams.attributeId.toString());
+            }
+
+            const latestSharedVersion = await this.consumptionController.attributes.getSharedVersionsOfRepositoryAttribute(parsedParams.attributeId, [requestInfo.peer], true);
+
+            if (latestSharedVersion.length === 0) {
+                sharedLocalAttribute = await this.copyExistingAttribute(existingSourceAttribute.id, requestInfo);
```
slavistan, line 127: What's the point of `copyExistingAttribute`? It wraps a single call to `createSharedLocalAttributeCopy` and is called only once. Might as well yeet it and insert its implementation here.

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
```diff
@@ -93,15 +111,60 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
...
+                });
+            }
+
+            const latestSharedAttribute = latestSharedVersion[0];
+            if (typeof latestSharedAttribute.shareInfo?.sourceAttribute === "undefined") {
+                throw new Error(`The Attribute ${latestSharedAttribute.id} does not fulfill the requirements of an own shared Attribute.`);
+            }
+
+            if (latestSharedAttribute.shareInfo.sourceAttribute.toString() === existingSourceAttribute.id.toString()) {
+                // return new AttributeAlreadySharedResponseItem
```
slavistan, line 141: Missing implementation. Do we need a dedicated response item here?

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
```diff
@@ -93,15 +111,60 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
...
+            }
+
+            const predecessorSourceAttribute = await this.consumptionController.attributes.getLocalAttribute(latestSharedAttribute.shareInfo.sourceAttribute);
+            if (typeof predecessorSourceAttribute === "undefined") {
+                throw TransportCoreErrors.general.recordNotFound(LocalAttribute, latestSharedAttribute.shareInfo.sourceAttribute.toString());
+            }
+
+            if (await this.consumptionController.attributes.isSubsequentInSuccession(predecessorSourceAttribute, existingSourceAttribute)) {
+                let successorSharedAttribute: LocalAttribute;
+                if (existingSourceAttribute.isIdentityAttribute()) {
```
slavistan, line 151: Why not use `.isOwnSharedIdentityAttribute()`?

packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -572,9 +572,16 @@ export class DataViewExpander {
...
                 let proposedValueOverruled = false;
                 if (responseItemDVO && responseItemDVO.result === ResponseItemResult.Accepted) {
-                    const proposeAttributeResponseItem = responseItemDVO as ProposeAttributeAcceptResponseItemDVO;
-                    if (JSON.stringify(proposeAttributeResponseItem.attribute.content.value) !== JSON.stringify(proposeAttributeRequestItem.attribute.value)) {
-                        proposedValueOverruled = true;
+                    if (responseItemDVO.type === "AttributeSuccessionAcceptResponseItemDVO") {
+                        const attributeSuccessionResponseItem = responseItemDVO as AttributeSuccessionAcceptResponseItemDVO;
+                        if (JSON.stringify(attributeSuccessionResponseItem.successor.content.value) !== JSON.stringify(proposeAttributeRequestItem.attribute.value)) {
+                            proposedValueOverruled = true;
+                        }
```
slavistan, line 579: Use lodash's `isEqual()` instead of JSON.stringify() for deep comparisons of objects. `JSON.stringify()` will break if the order of keys differs in the operands. 

Also, you can yeet the if-clause.

```ts
proposedValueOverrule = !_.isEqual(attributeSuccessionResponseItem.successor.content.value, proposedAttributeRequestItem.attribute.value)
```

Same thing below.

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
```diff
@@ -71,19 +73,35 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
...
+                if (typeof latestSharedVersion[0].shareInfo?.sourceAttribute === "undefined") {
+                    throw new Error(`The Attribute ${latestSharedVersion[0].id} does not fulfill the requirements of an own shared Attribute.`);
+                }
+
+                const latestSharedVersionSourceAttribute = await this.consumptionController.attributes.getLocalAttribute(latestSharedVersion[0].shareInfo.sourceAttribute);
+                if (typeof latestSharedVersionSourceAttribute === "undefined") {
+                    throw new Error(`The Attribute ${latestSharedVersion[0].shareInfo.sourceAttribute} was not found.`);
+                }
+
+                if (await this.consumptionController.attributes.isSubsequentInSuccession(foundAttribute, latestSharedVersionSourceAttribute)) {
```
J-Gann, line 93: What if the proposed e.g. "foundAttribute" is the "latestSharedVersionSourceAttribute"? I belive "isSubsequentInSuccession" returns false hence not throwing an error.

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
Milena-Czierlinski: This will be implemented in a different PR on this feature branch. The idea is to allow the Recipient to answer with an already shared Attribute, if no successor was shared in the meantime, e.g. to confirm that the Attribute, the Sender has of the Recipient and proposed to them, is still valid. In the ResponseItem we want to share the ID of the shared Attribute, so that the App can directly display it. Maybe, we can discuss whether we want to return this ID. If not, we could in fact just use the AcceptResponseItem.

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
Milena-Czierlinski: Actually, `existingSourceAttribute` must be a RepositoryAttribute, not an own shared IdentityAttribute. But you are right, this can be formulated more precisely.

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
slavistan: Using helper functions with descriptive names to avoid evil comments is prefectly fine. Just in this particular instance it's a downgrade compared to a simple call to `createSharedLocalAttributeCopy` since all you did is to burden yourself with the task of finding another descriptive name for a function that already exists. _Create a shared attribute copy_ is a more accurate description of what's going on than _copy existing attribute_, which suggested to me that we're duplicating the attribute for some reason.

My 2 cents, nothing dramatic. Keep it, if you feel strongly about it.

packages/consumption/test/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.test.ts
```diff
@@ -590,12 +593,79 @@ describe("ProposeAttributeRequestItemProcessor", function () {
...
+            const successorOwnSharedIdentityAttribute = await consumptionController.attributes.getLocalAttribute((result as AttributeSuccessionAcceptResponseItem).successorId);
+            expect(successorOwnSharedIdentityAttribute).toBeDefined();
+            expect(successorOwnSharedIdentityAttribute!.shareInfo).toBeDefined();
+            expect(successorOwnSharedIdentityAttribute!.shareInfo!.peer.toString()).toStrictEqual(incomingRequest.peer.toString());
+            expect(successorOwnSharedIdentityAttribute!.shareInfo!.sourceAttribute).toStrictEqual(successorRepositoryAttribute.id);
+            expect(successorOwnSharedIdentityAttribute!.succeeds).toStrictEqual(predecessorOwnSharedIdentityAttribute.id);
+
+            const updatedPredecessorOwnSharedIdentityAttribute = await consumptionController.attributes.getLocalAttribute(predecessorOwnSharedIdentityAttribute.id);
+            expect(updatedPredecessorOwnSharedIdentityAttribute!.succeededBy).toStrictEqual(successorOwnSharedIdentityAttribute!.id);
+        });
```
britsta, line 669: In the case of the AttributeSuccessionAcceptResponseItem we have also added a test for the applyIncomingResponseItem function in the ReadAttributeRequestItemProcessor.test.ts, perhaps such a test may also be added for the ProposeAttributeRequestItemProcessor.

PR #102, opened 2024-04-15
Refactor/remove easily inferrable type annotations

jkoenig134:
I had this stashed and why not PR it :) These type annotations shouldn't be in our codebase - now they are gone :D

Background of this are the [inlay hints](https://code.visualstudio.com/docs/typescript/typescript-editing#_inlay-hints) that are supported by all major IDEs and lead to less reading for people that don't want these types and people who want the types can simply enable these.


packages/transport/src/modules/secrets/SecretController.ts
```diff
@@ -74,15 +73,15 @@ export class SecretController extends TransportController {
...
+        const plainBuffer = CoreBuffer.fromUtf8(plainString);
 
-        const nonce: number = await this.increaseNonce();
-        const encryptionKey: CryptoSecretKey = await CoreCrypto.deriveKeyFromBase(await this.getBaseKey(), nonce, SecretController.secretContext);
+        const nonce = await this.increaseNonce();
+        const encryptionKey = await CoreCrypto.deriveKeyFromBase(await this.getBaseKey(), nonce, SecretController.secretContext);
 
-        const cipher: CryptoCipher = await CoreCrypto.encrypt(plainBuffer, encryptionKey);
-        const createdAt: CoreDate = CoreDate.utc();
-        const secretContainerInterface: ISecretContainerCipher = {
```
britsta, line 85: It may not be a problem, but when I hover the cursor over the `secretContainerInterface` variable, I no longer see that it is an `ISecretContainerCipher`.

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -289,21 +289,21 @@ export class MessageController extends TransportController {
                 throw CoreErrors.general.recordNotFound(Relationship, recipient.toString());
             }
 
-            const signature: ICryptoSignature = await this.secrets.sign(relationship.relationshipSecretId, plaintextBuffer);
```
britsta, line 292: It may not be a problem, but when I hover the cursor over the `signature` variable, I no longer see that it is an `ICryptoSignature` but a `CryptoSignature`.

packages/transport/src/modules/devices/DeviceSecretController.ts
```diff
@@ -61,22 +51,22 @@ export class DeviceSecretController extends TransportController {
...
 
-        const encryptionKey: CryptoSecretKey = await CoreCrypto.deriveKeyFromBase(this.getBaseKey(), 1, DeviceSecretController.secretContext);
+        const encryptionKey = await CoreCrypto.deriveKeyFromBase(this.getBaseKey(), 1, DeviceSecretController.secretContext);
 
-        const cipher: ICryptoCipher = await CoreCrypto.encrypt(plainBuffer, encryptionKey);
-        const date: CoreDate = CoreDate.utc();
-        const secretContainerInterface: ISecretContainerCipher = {
+        const cipher = await CoreCrypto.encrypt(plainBuffer, encryptionKey);
+        const date = CoreDate.utc();
+        const secretContainerInterface = {
```
britsta, line 61: It may not be a problem, but when I hover the cursor over the `secretContainerInterface` variable, I no longer see that it is an `ISecretContainerCipher`.

packages/transport/src/modules/devices/DeviceSecretController.ts
```diff
@@ -89,15 +79,15 @@ export class DeviceSecretController extends TransportController {
         const secretObj = await this.secrets.get(name);
         if (!secretObj) return;
 
-        const baseKey: CryptoSecretKey = this.getBaseKey();
-        const secret: SecretContainerCipher = SecretContainerCipher.from(secretObj);
-        const decryptionKey: CryptoSecretKey = await CoreCrypto.deriveKeyFromBase(baseKey, 1, DeviceSecretController.secretContext);
-        const plainBuffer: ICoreBuffer = await CoreCrypto.decrypt(secret.cipher, decryptionKey);
```
britsta, line 95: It may not be a problem, but when I hover the cursor over the `plainBuffer` variable, I no longer see that it is an `ICoreBuffer` but a `CoreBuffer`.

packages/transport/src/modules/messages/MessageController.ts
jkoenig134: in this case we get an even stronger type than before

packages/transport/src/modules/devices/DeviceSecretController.ts
jkoenig134: good catch!

packages/transport/src/modules/devices/DeviceSecretController.ts
jkoenig134: CoreBuffer is way stricter, should be safer now.

packages/transport/src/modules/secrets/SecretController.ts
jkoenig134: good catch!

britsta (approved):
LGTM 👍

PR #106, opened 2024-04-16
Feature/pass profile name in devicesharedsecret

jkoenig134:
The feature introduces the possibility to pass the profile name from one device to another to keep them the same.

The user can overwrite this, but it will override the default name of the created profile.


jkoenig134:
@Milena-Czierlinski sure thing :)

PR #108, opened 2024-04-18
Feature/Add AttributeAlreadySharedAcceptResponseItem

Milena-Czierlinski:


Milena-Czierlinski:
In principle, this could also be offered for the ShareAttributeRequestItem and CreateAttributeRequestItem to make clear, that the Attribute was already shared. Then, we would need to add AcceptResponseParameters for those RequestItems. I'm not sure if this is really necessary, since the same can be achieved simply rejecting the RequestItem and optionally stating the reason in the message.

PR #109, opened 2024-04-18
Feature/Rename getSharedVersionsOfAttribute

Milena-Czierlinski:
Since this poses a breaking change, I'll merge `feature/add-acceptresponseitems` into `release/v5`.

### DIFF
--- a/packages/consumption/src/modules/attributes/AttributesController.ts
+++ b/packages/consumption/src/modules/attributes/AttributesController.ts
@@ -973,7 +973,6 @@ export class AttributesController extends ConsumptionBaseController {
         return false;
     }
 
-    // TODO: rename this and the respective runtime use case
     public async getSharedVersionsOfAttribute(id: CoreId, peers?: CoreAddress[], onlyLatestVersions = true): Promise<LocalAttribute[]> {
         let repositoryAttribute = await this.getLocalAttribute(id);
         if (typeof repositoryAttribute === "undefined") {
--- a/packages/consumption/src/modules/attributes/AttributesController.ts
+++ b/packages/consumption/src/modules/attributes/AttributesController.ts
@@ -974,7 +974,7 @@ export class AttributesController extends ConsumptionBaseController {
     }
 
     // TODO: rename this and the respective runtime use case
-    public async getSharedVersionsOfRepositoryAttribute(id: CoreId, peers?: CoreAddress[], onlyLatestVersions = true): Promise<LocalAttribute[]> {
+    public async getSharedVersionsOfAttribute(id: CoreId, peers?: CoreAddress[], onlyLatestVersions = true): Promise<LocalAttribute[]> {
         let repositoryAttribute = await this.getLocalAttribute(id);
         if (typeof repositoryAttribute === "undefined") {
             throw TransportCoreErrors.general.recordNotFound(LocalAttribute, id.toString());
--- a/packages/runtime/src/useCases/consumption/attributes/GetSharedVersionsOfAttribute.ts
+++ b/packages/runtime/src/useCases/consumption/attributes/GetSharedVersionsOfAttribute.ts
@@ -6,7 +6,7 @@ import { LocalAttributeDTO } from "../../../types";
 import { AddressString, AttributeIdString, RuntimeErrors, SchemaRepository, SchemaValidator, UseCase } from "../../common";
 import { AttributeMapper } from "./AttributeMapper";
 
-export interface GetSharedVersionsOfRepositoryAttributeRequest {
+export interface GetSharedVersionsOfAttributeRequest {
     attributeId: AttributeIdString;
     peers?: AddressString[];
     /**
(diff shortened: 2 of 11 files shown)


packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -973,8 +973,7 @@ export class AttributesController extends ConsumptionBaseController {
         return false;
     }
 
-    // TODO: rename this and the respective runtime use case
-    public async getSharedVersionsOfRepositoryAttribute(id: CoreId, peers?: CoreAddress[], onlyLatestVersions = true): Promise<LocalAttribute[]> {
+    public async getSharedVersionsOfAttribute(id: CoreId, peers?: CoreAddress[], onlyLatestVersions = true): Promise<LocalAttribute[]> {
         let repositoryAttribute = await this.getLocalAttribute(id);
```
britsta, line 977: We can then also adapt this variable name here, right?

britsta (commented):
Also note that the documentation of the corresponding use case https://enmeshed.eu/use-case-consumption-get-shared-versions-of-a-repositoryattribute needs to be adapted then, after this PR is published

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -973,25 +973,24 @@ export class AttributesController extends ConsumptionBaseController {
         return false;
     }
 
-    // TODO: rename this and the respective runtime use case
-    public async getSharedVersionsOfRepositoryAttribute(id: CoreId, peers?: CoreAddress[], onlyLatestVersions = true): Promise<LocalAttribute[]> {
-        let repositoryAttribute = await this.getLocalAttribute(id);
-        if (typeof repositoryAttribute === "undefined") {
+    public async getSharedVersionsOfAttribute(id: CoreId, peers?: CoreAddress[], onlyLatestVersions = true): Promise<LocalAttribute[]> {
+        let sourceAttribute = await this.getLocalAttribute(id);
+        if (typeof sourceAttribute === "undefined") {
```
jkoenig134, line 978: don't we have to throw if the attribute is already shared? before we checked that it's a repository attribute

additionally: does it make sense to call this function for RelationshipAttributes? In that case it would only return third party attributes, right?

packages/runtime/src/useCases/consumption/attributes/GetSharedVersionsOfAttribute.ts
```diff
@@ -30,7 +30,7 @@ export class GetSharedVersionsOfRepositoryAttributeUseCase extends UseCase<GetSh
         super(validator);
     }
 
-    protected async executeInternal(request: GetSharedVersionsOfRepositoryAttributeRequest): Promise<Result<LocalAttributeDTO[]>> {
+    protected async executeInternal(request: GetSharedVersionsOfAttributeRequest): Promise<Result<LocalAttributeDTO[]>> {
```
jkoenig134, line 33: you have renamed the UseCase, but you still check if it's a repository attribute

packages/runtime/test/consumption/attributes.test.ts
jkoenig134, line 1: there are still tests that indicate that the "new" usecase is only for repository attributes.

jkoenig134:
I think we should keep the UseCase that's only for repositoryAttributes and add a new one which acts for all attributes. Then it would not be a breaking change which is cool.

packages/consumption/src/modules/attributes/AttributesController.ts
Milena-Czierlinski: > don't we have to throw if the attribute is already shared? before we checked that it's a repository attribute
Now we simply return an empty list, if there are no shared versions of the Attribute (in accordance to e.g. getAttributes)

> additionally: does it make sense to call this function for RelationshipAttributes? In that case it would only return third party attributes, right?
This is exactly the desired behavior. The need for this arouse in the parent branch (e.g. https://github.com/nmshd/runtime/pull/76/files#diff-e79d6068574bbea96206a547db0e7d1a54f4b45b3efc1cd453e8d0fb5ff31d59R91), where we need to check if a successor of a RelationshipAttribute was already shared with a third party, before we may share a predecessor as Response to a ReadAttributeRequestItem. But we'll also make use of it within the Attribute deletion, where we need to get all ThirdPartyRelationshipAttributes of a RelationshipAttribute to unset their `shareInfo.sourceAttribute` field. (This is already implemented in the AttributesController, but uses the old name getSharedVersionsOfRepositoryAttribute to receive possible predecessors. So if I am not mistaken, we should currently have a bug here: In case of a succeeded ThirdPartyRelationshipAttribute, an error should be thrown trying to delete the predecessor ThirdPartyRelationshipAttribute. But I think before this parent branch is merged, it isn't even possible to succeed ThirdPartyRelationshipAttributes, so this case hasn't occured so far.)

PR #112, opened 2024-04-23
Feature/housenumber validation

RuthDiG:
I added a validation for the housenumber which didn't exist before, so you can only write housenumbers with numbers or numbers and one added letter with or without a blank. 
Then you can insert values like '', '1/3', '1-4', '1 / 3', '1 - 4' or for france you can insert values like 
'3bis', '3ter', '3quater' (also '3 bis', '3 ter', '3 quater' ). 
Also works '5 1/2' or also '5 1/2 b'. 
(https://de.wikipedia.org/wiki/Hausnummer)


packages/content/test/attributes/HouseNumber.test.ts
```diff
@@ -0,0 +1,63 @@
+import { ParsingError } from "@js-soft/ts-serval";
+import { CountryAlpha2, StreetAddress } from "../../src";
+
+describe("creation of RepositoryAttributes of Attribute Value Type Streetadress", function () {
+    test("can create a RepositoryAttribute of Attribute Value Type Streetadress", function () {
```
jkoenig134, line 5: typos in StreetAddress

jkoenig134:
@RuthDiG you mention all possibilities for france and germany, but does this match all house numbers in the world?

packages/content/test/attributes/HouseNumber.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        });
+        expect(validStreetAdress.recipient.toString()).toBe("Hugo Becker");
+        expect(validStreetAdress.street.toString()).toBe("Luisenstr.");
+        expect(validStreetAdress.houseNo.toString()).toBe("38-40");
+        expect(validStreetAdress.zipCode.toString()).toBe("76646");
+        expect(validStreetAdress.city.toString()).toBe("Bruchsal");
+        expect(validStreetAdress.country.toString()).toBe(CountryAlpha2.DE);
+        expect(validStreetAdress.state?.toString()).toBe("Baden-Württemberg");
+    });
+    test("returns an error when trying to create an Attribute Value Type Streetadress with an invalid value for houseNo.", function () {
```
britsta, line 23: I'd write in the test's name briefly more precisely why the value is invalid (houseNo consists of letters only).

Same for the other test below.

RuthDiG:
> @RuthDiG you mention all possibilities for france and germany, but does this match all house numbers in the world?
@jkoenig134 I have now looked over the possibilities worldwide and I have to add the following possibilities: 
- Numbers including 0 in the beginning 
- (Uruguay) 312 bis bis bis 
- 12+1 
- Word Number, Word. Number, Word Number, (or something like this)
- 3-13-2 3F Word
- Flat 3/L 1
- Lt. 9,2 o/d  
- 2, 1/esq.
- 125/III
In Venezuela, however, they only write the name of the house without a number, but that was exactly that what the BMBF didn't want to be possible. So I will implement only the rest now, okay?

jkoenig134:
I talked to Thomas and maybe it's not a good idea to validate this in the runtime ... when we do this in the App we can add other validations (like when DACH -> then must start with a number)

We shouldn't block valid addresses (like Venezuela) here.

RuthDiG:
> I talked to Thomas and maybe it's not a good idea to validate this in the runtime ... when we do this in the App we can add other validations (like when DACH -> then must start with a number)
> 
> We shouldn't block valid addresses (like Venezuela) here.

Okay so you wouldn't do any validation here and I should delete this PR?

jkoenig134:
@stnmtz what's your final opinion on that?

RuthDiG:
> @stnmtz what's your final opinion on that?

@stnmtz @jkoenig134 I thought about the Birthdate-Validation of @britsta there is also a validation that regards the month and the day if you are tipping 30.02.2013, so I can do in the runtime the same with country and the housenumber and there can then be afterwards additional validations in the app.
Furthermore I think later on it would also be nice or necessary to have a validation  if a street and the corresponding number really exist.

PR #117, opened 2024-04-29
Feature/relationship termination

Magnus-Kuhn:


packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -398,7 +410,7 @@ export class RelationshipsController extends TransportController {
     }
 
     @log()
-    private async completeStateTransition(targetStatus: RelationshipStatus, id: CoreId) {
+    private async executeNonDeletingOperation(operation: RelationshipOperation, id: CoreId) {
```
jkoenig134, line 413: I don't like the new method name... and you should not rename `RelationshipAuditLogEntryReason` to `RelationshipOperation` and therefore just define a completely new concept that nobody knows.

packages/runtime/src/useCases/transport/identity/CheckIdentity.ts
```diff
@@ -47,25 +47,30 @@ export class CheckIdentityUseCase extends UseCase<CheckIdentityRequest, CheckIde
...
+                case RelationshipStatus.Terminated:
+                    return Result.ok({
+                        peer: true,
+                        relationshipTerminated: true,
+                        relationship: relationshipDTO
+                    });
+                default:
+                    return Result.ok({
+                        unknown: true
+                    });
```
jkoenig134, line 72:
```suggestion
                    return Result.ok({
                        peer: true,
                        relationship: relationshipDTO
                    });
```

jkoenig134: I mean ... it found a relationship so it is definitely not unknown.

packages/consumption/test/modules/requests/RequestsIntegrationTest.ts
```diff
@@ -89,13 +89,22 @@ export class RequestsTestsContext {
             context.mockEventBus,
             { address: CoreAddress.from("anAddress") },
             {
-                getActiveRelationshipToIdentity: () => Promise.resolve(context.relationshipToReturnFromGetActiveRelationshipToIdentity)
+                getRelationshipToIdentity: () => Promise.resolve(context.relationshipToReturnFromGetRelationshipToIdentity)
```
jkoenig134, line 92: you can only work with requests if there is an active relationship... I don't really get why you renamed / changed all of this.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -401,6 +417,14 @@ export class IncomingRequestsController extends ConsumptionBaseController {
         }
     }
 
+    private async checkRelationshipStatus(peer: CoreAddress) {
+        const relationship = await this.relationshipResolver.getRelationshipToIdentity(peer);
+        // no relationship is for a template request on new relationship
+        if (relationship && relationship.status !== RelationshipStatus.Active) {
+            throw CoreErrors.requests.noMatchingRelationship(peer.toString());
```
jkoenig134, line 424: the error message is wrong. It says pending or active, but you only check for active here.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -35,7 +47,10 @@ export class IncomingRequestsController extends ConsumptionBaseController {
         private readonly processorRegistry: RequestItemProcessorRegistry,
         parent: ConsumptionController,
         private readonly eventBus: EventBus,
-        private readonly identity: { address: CoreAddress }
+        private readonly identity: { address: CoreAddress },
+        private readonly relationshipResolver: {
+            getRelationshipToIdentity(id: CoreAddress): Promise<Relationship | undefined>;
```
jkoenig134, line 52: If you only want an active relationship, please rename this to getActiveRelationshipToIdentity

jkoenig134 (changes requested):

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -40,14 +41,20 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
...
     }
 
     public async canCreate(params: ICanCreateOutgoingRequestParameters): Promise<ValidationResult> {
         const parsedParams = CanCreateOutgoingRequestParameters.from(params);
+        if (parsedParams.peer) {
+            const relationship = await this.relationshipResolver.getRelationshipToIdentity(parsedParams.peer);
+            if (!relationship || !(relationship.status === RelationshipStatus.Pending || relationship.status === RelationshipStatus.Active)) {
+                return ValidationResult.error(CoreErrors.requests.noMatchingRelationship(parsedParams.peer.toString()));
+            }
+        }
```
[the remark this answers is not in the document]
Magnus-Kuhn: This is for a request from a relationship template on a new relationship. I also considered putting this in the `create`-method, but it is a validation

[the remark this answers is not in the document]
jkoenig134: I see. Let this comment open please, I have to think about why we didn't add this earlier (or maybe also removed it, looks like the `relationshipResolver` was unused...)

packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
```diff
@@ -258,8 +258,6 @@ describe("OutgoingRequestsController", function () {
             });
 
             test("create an outgoing request from relationship creation with an active relationship", async function () {
-                await Given.anActiveRelationshipToIdentity();
```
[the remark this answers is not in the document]
Magnus-Kuhn: the canCreate validation requires a relationship for any creation, which is why the relationship is not created in the individual tests but put into the context at initialisation

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
Magnus-Kuhn: no relationship would also be an acceptable result

jkoenig134: yeah - but you want active or none. Not just any one.

packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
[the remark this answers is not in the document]
jkoenig134: But what if you want to have no active relationship for a test?

packages/runtime/src/useCases/transport/identity/CheckIdentity.ts
Magnus-Kuhn: what is the peer vs. unknown distinction used for? The remaining cases are rejected and revoked

jkoenig134: unknown means you have never heard of defined the address, if you find a relationship you defenitely have contacted the relationship before.

packages/runtime/src/useCases/transport/identity/CheckIdentity.ts
jkoenig134, line 1: @Milena-Czierlinski you said that this is used nowhere? Maybe we should really delete it to avoid other confusions.

packages/transport/src/modules/relationships/RelationshipsController.ts
Magnus-Kuhn: I didn't want to put an AuditLog in there - then I suggest changing the name RelationshipAuditLogEntryReason everywhere.

And completeStateTransition won't be enough once the reactivation is implemented (of course, the nonDeleting only matters when the decomposition is implemented, I could remove that)

jkoenig134: What's the matter with using another method for the reactivation? (or think about that for the reactivation)

Please undo all that for now that it doesn't get confusing.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
Magnus-Kuhn: getActiveRelationship would not distinguish between no relationship and inactive relationship

jkoenig134: It does not have to, we don't care. Actually you don't distinguish about that below.

packages/transport/src/modules/relationships/RelationshipsController.ts
Magnus-Kuhn: this method does the updating + event issuing - we'd end up duplicating this

jkoenig134: ok, let's think about this when we implement it. undo it for now.

Magnus-Kuhn: alright

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
jkoenig134: ok, I'll also think about this again!

packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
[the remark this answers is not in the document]
Magnus-Kuhn: in that case we'd have to come up with something... - with all the issues my current feeling is that the best approach might be renaming the canDecide etc. and applying the relationship status checking separately to the onNewRelationship vs. to other cases

packages/runtime/src/useCases/transport/identity/CheckIdentity.ts
Milena-Czierlinski: Yes, I agree.

jkoenig134: @stnmtz any vetos?

packages/consumption/test/modules/requests/RequestsIntegrationTest.ts
```diff
@@ -104,6 +113,7 @@ export class RequestsTestsContext {
             return originalCanCreate.call(context.outgoingRequestsController, params);
         };
 
+        context.relationshipToReturnFromGetRelationshipToIdentity = TestObjectFactory.createRelationship();
```
jkoenig134, line 116: I'd highly vote for undoing that and putting it everywhere required. Just assuming for every test that there is a relationship breaks the whole concept.

@tnotheis what's your opinion on that?

packages/consumption/test/modules/requests/RequestsIntegrationTest.ts
jkoenig134: got it now.

packages/consumption/test/modules/requests/RequestsIntegrationTest.ts
tnotheis: Yeah, I agree with @jkoenig134.

packages/consumption/test/modules/requests/RequestsIntegrationTest.ts
```diff
@@ -53,7 +53,7 @@ export class RequestsTestsContext {
     public outgoingRequestsController: OutgoingRequestsController;
     public currentIdentity: CoreAddress;
     public mockEventBus = new MockEventBus();
-    public relationshipToReturnFromGetActiveRelationshipToIdentity: Relationship | undefined;
+    public relationshipToReturnFromGetRelationshipToIdentity: Relationship | undefined;
```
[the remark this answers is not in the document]
Magnus-Kuhn: Depends on if the tests for the relationship status are integration- or e2e-tests. The test for creating an outgoing request on new relationship is e2e-style and I have created the other relationship tests as e2e-runtime tests.

packages/consumption/test/modules/requests/RequestsIntegrationTest.ts
Magnus-Kuhn: alright

packages/consumption/test/modules/requests/RequestsIntegrationTest.ts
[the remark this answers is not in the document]
jkoenig134: e2e tests are not enough. We definitely also need these tests here.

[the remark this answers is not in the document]
jkoenig134: As an explanation why: it is way easier to track where errors come from in these tests than in full-fledged end2end tests. End2end tests are nice to have, but these tests saved the whole requests handling multiple times.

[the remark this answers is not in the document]
Magnus-Kuhn: Why are requests usually created the standard way rather than using the test object factory? If I wanted to put the relationship individually into each test where it is required, it would be confusing why it sometimes isn't there.

However, my favorite approach would be to do the relationship setup in the beforeEach of each test suite - it is more realistic anyway.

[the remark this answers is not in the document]
Magnus-Kuhn: by calling the create controller method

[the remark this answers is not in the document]
jkoenig134: > by calling the create controller method

I think it's easier that way

> However, my favorite approach would be to do the relationship setup in the beforeEach of each test suite - it is more realistic anyway.

I think this is ok if every test shares what kind of relationship should be returned in a define declaration.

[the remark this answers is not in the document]
Magnus-Kuhn: What is a define declaration?
Also - would putting the current relationship termination tests into consumption / transport mean that the runtime tests aren't necessary anymore? There's no substantial logic there. I'm generally not clear about the overall testing approach.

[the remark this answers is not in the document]
jkoenig134: sorry, i meant describe.

no, transport tests doesn't remove the need for runtime tests.

packages/consumption/src/consumption/CoreErrors.ts
```diff
@@ -268,6 +268,10 @@ class Requests {
         return new CoreError("error.consumption.requests.invalidRequestItem", message);
     }
 
+    public noMatchingRelationship(address: string) {
+        return new CoreError("error.consumption.requests.noMatchingRelationship", `An active or pending Relationship with the given address '${address}' does not exist.`);
+    }
```
jkoenig134, line 273: we have to completely re-think this error and maybe introduce multiple new ones:

- for the IncomingRequestController we accept no relationship and an active relationship - these are the things we can answer to
- for the OutgoingRequestController I think this is fitting

Two possibilities to fix this:
1. provide two errors describing the two different things the controllers need
2. pass the complete message to the method

I'd highly vote for 2. but the error code has to be improved significantly. 

It should describe what does `noMatchingRelationship` mean? Not matching for what operation?

packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
```diff
@@ -258,8 +270,6 @@ describe("OutgoingRequestsController", function () {
             });
 
             test("create an outgoing request from relationship creation with an active relationship", async function () {
```
jkoenig134, line 272: now the title states that this function should be called using an active relationship but in beforeEach you run `await Given.aPendingRelationshipToIdentity();` which sets a pending relationship instead of an active.

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -281,3 +291,83 @@ describe("Attributes for the relationship", () => {
...
+        expect(result).toBeAnError(/.*/, "error.transport.messages.noMatchingRelationship");
+    });
+
+    test("should not decide a request", async () => {
+        await syncUntilHasRelationships(services2.transport);
+        const incomingRequest = (await services2.eventBus.waitForEvent(IncomingRequestReceivedEvent)).data;
+
+        const canAcceptResult = (await services2.consumption.incomingRequests.canAccept({ requestId: incomingRequest.id, items: [{ accept: true }] })).value;
+        expect(canAcceptResult.isSuccess).toBe(false);
+        expect(canAcceptResult.code).toBe("error.consumption.requests.noMatchingRelationship");
```
jkoenig134, line 343: please use the utility method toBeAnError instead of doing everything manually. Same for the other occurences.

Magnus-Kuhn: the `canAccept/Reject` return validation results though

jkoenig134: you can make the function also accept these.

packages/transport/test/testHelpers/TestUtil.ts
```diff
@@ -381,6 +381,14 @@ export class TestUtil {
         return [acceptedRelationshipFromSelf, acceptedRelationshipPeer];
     }
 
+    public static async terminateRelationship(from: AccountController, to: AccountController): Promise<Relationship[]> {
```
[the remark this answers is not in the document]
Magnus-Kuhn: I like this - I'll also change the method above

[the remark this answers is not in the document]
jkoenig134: even better!

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -338,9 +338,10 @@ describe("RelationshipTermination", () => {
         await syncUntilHasRelationships(services2.transport);
         const incomingRequest = (await services2.eventBus.waitForEvent(IncomingRequestReceivedEvent)).data;
 
-        const canAcceptResult = (await services2.consumption.incomingRequests.canAccept({ requestId: incomingRequest.id, items: [{ accept: true }] })).value;
-        expect(canAcceptResult.isSuccess).toBe(false);
-        expect(canAcceptResult.code).toBe("error.consumption.requests.noMatchingRelationship");
+        const canAcceptResult = await services2.consumption.incomingRequests.canAccept({ requestId: incomingRequest.id, items: [{ accept: true }] });
+        expect(canAcceptResult).toBeAnError(/.*/, "error.consumption.requests.noMatchingRelationship");
+        // expect(canAcceptResult.isSuccess).toBe(false);
+        // expect(canAcceptResult.code).toBe("error.consumption.requests.noMatchingRelationship");
```
jkoenig134, line 344: commented out code?

Magnus-Kuhn: I didn't mean to commit that - whoops.
But if you've seen this: this test doesn't work, the result is successful. I've not yet found a way to make this work - and the requests tests also do it my previous way

jkoenig134: Ok, i'd say return it as it was and we'll do a refactoring afterwards.

packages/consumption/src/consumption/CoreErrors.ts
Magnus-Kuhn: if you want to answer the question about the operation inside error code, you're in option 1 - I also think option 2 fits better to the existing errors. I think I wouldn't even mention the correct statuses in the message because of the special cases, just the current wrong status.

jkoenig134: Great!

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -51,8 +51,17 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
...
-                return ValidationResult.error(CoreErrors.requests.noMatchingRelationship(parsedParams.peer.toString()));
+            if (!relationship) {
+                return ValidationResult.error(
+                    CoreErrors.requests.missingRelationship(`You cannot create a request to ${parsedParams.peer.toString()} since you are not in a relationship.`)
+                );
+            }
+            if (!(relationship.status === RelationshipStatus.Pending || relationship.status === RelationshipStatus.Active)) {
+                return ValidationResult.error(
+                    CoreErrors.requests.wrongRelationshipStatus(
+                        `You cannot create a request to ${parsedParams.peer.toString()} since the relationship is in status ${relationship.status}.`
```
jkoenig134, line 62:
```suggestion
                        `You cannot create a request to ${parsedParams.peer.toString()} since the relationship is in status '${relationship.status}'.`
```

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -423,7 +423,9 @@ export class IncomingRequestsController extends ConsumptionBaseController {
         const relationship = await this.relationshipResolver.getRelationshipToIdentity(peer);
         // no relationship is for a template request on new relationship
         if (relationship && relationship.status !== RelationshipStatus.Active) {
-            return ValidationResult.error(CoreErrors.requests.noMatchingRelationship(peer.toString()));
+            return ValidationResult.error(
+                CoreErrors.requests.wrongRelationshipStatus(`You cannot decide a request from ${peer.toString()} since the relationship is in status ${relationship.status}.`)
```
jkoenig134, line 427:
```suggestion
                CoreErrors.requests.wrongRelationshipStatus(`You cannot decide a request from ${peer.toString()} since the relationship is in status '${relationship.status}'.`)
```

jkoenig134: Maybe also put the address in 's

jkoenig134: `since the relationship` 

-> I'd say `since you have a relationship with the peer in status ... but you need an Active relationship.`

jkoenig134: Also the error has `You cannot decide a request` but the method name does not tell anything about that this method can only be called for decide. Maybe inline this.

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
jkoenig134: same things as in Incoming

packages/transport/src/core/CoreErrors.ts
```diff
@@ -44,8 +44,8 @@ class Messages {
         );
     }
 
-    public noMatchingRelationship(address: string) {
-        return new CoreError("error.transport.messages.noMatchingRelationship", `A Relationship with the given address '${address}' does not exist.`);
+    public missingOrInactiveRelationship(address: string) {
+        return new CoreError("error.transport.messages.noMatchingRelationship", `An active Relationship with the given address '${address}' does not exist.`);
```
jkoenig134, line 48:
```suggestion
        return new CoreError("error.transport.messages.missingOrInactiveRelationship", `An active Relationship with the given address '${address}' does not exist.`);
```

jkoenig134: also the message seems to be incorrect

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
Magnus-Kuhn: I guess the ' for consistency, although I don't like them - it's clear without them. And going to active is always the only thing an end-user can do if the status is wrong. I even feel that depending on the UI this may be more confusing if the users only see Accept/Reactivate instead of "Switch to active" as options.

packages/transport/src/core/CoreErrors.ts
Magnus-Kuhn: what's incorrect about it?

jkoenig134: nothing, I misread it.

jkoenig134 (approved):
LGTM

packages/transport/test/end2end/End2End.test.ts
```diff
@@ -420,6 +420,51 @@ describe("RelationshipTest: Revoke", function () {
...
+    test("should terminate a relationship between two accounts", async function () {
+        const relationshipId = (await TestUtil.addRelationship(from, to)).acceptedRelationshipFromSelf.id;
+
+        const terminatedRelationshipFromSelf = await from.relationships.terminate(relationshipId);
+        expect(terminatedRelationshipFromSelf.id.toString()).toStrictEqual(relationshipId.toString());
+        expect(terminatedRelationshipFromSelf.status).toStrictEqual(RelationshipStatus.Terminated);
+        expect(terminatedRelationshipFromSelf.cache?.auditLog).toHaveLength(3);
+        expect(terminatedRelationshipFromSelf.cache!.auditLog[2].newStatus).toBe(RelationshipStatus.Terminated);
+
+        // Get terminated relationship
```
Milena-Czierlinski, line 456: Can this be omitted?

packages/transport/test/end2end/End2End.test.ts
```diff
@@ -420,6 +420,51 @@ describe("RelationshipTest: Revoke", function () {
...
+    let to: AccountController;
+
+    beforeAll(async function () {
+        connection = await TestUtil.createDatabaseConnection();
+        transport = TestUtil.createTransport(connection);
+
+        await transport.init();
+
+        const accounts = await TestUtil.provideAccounts(transport, 2);
+        from = accounts[0];
```
Milena-Czierlinski, line 437: Perhaps the names can be chosen better, such if when you read them, you know they describe accounts.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -401,6 +419,17 @@ export class IncomingRequestsController extends ConsumptionBaseController {
         }
     }
 
+    private async checkRelationshipStatus(peer: CoreAddress): Promise<ValidationResult> {
+        const relationship = await this.relationshipResolver.getRelationshipToIdentity(peer);
+        // no relationship is for a template request on new relationship
```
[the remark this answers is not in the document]
Magnus-Kuhn: template request on new relationship: a request from the on new relationship section of a relationship template

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -401,6 +419,17 @@ export class IncomingRequestsController extends ConsumptionBaseController {
...
+        if (relationship && relationship.status !== RelationshipStatus.Active) {
```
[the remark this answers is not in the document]
Magnus-Kuhn: a decent amount of the transport library does it the other way - I guess we can try to agree on something?!

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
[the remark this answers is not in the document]
Milena-Czierlinski: And what about the "no relationship is..." part? What does that sentence mean?

packages/transport/test/end2end/End2End.test.ts
Magnus-Kuhn: you essentially propose refactoring the transport library :)

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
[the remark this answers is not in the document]
jkoenig134: maybe we should go away from that ugly typeof thing. Please leave these things as is in this PR.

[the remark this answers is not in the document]
Milena-Czierlinski: I remember tons of these comments in my PRs and you arguing in favor of it. 😅 I don't feel strongly about this, though. But maybe we can discuss this shortly in the next sync, so that we have a guide line for the following PRs.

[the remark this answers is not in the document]
jkoenig134: and I definitely had reasons about that. But opinions change.

> But maybe we can discuss this shortly in the next sync

Yes, please!

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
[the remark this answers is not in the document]
jkoenig134: Yeah, @Magnus-Kuhn please make sure everyone is able to understand this. I suggest 

`It is save to decide an incoming Request when no Relationship is found as this is the case when the Request origins from onNewRelationship of the RelationshipTemplateContent`

option b would be to remove the comment, I don't think that its **really** necessary here.

packages/consumption/test/modules/requests/IncomingRequestsController.test.ts
```diff
@@ -384,7 +384,7 @@ describe("IncomingRequestsController", function () {
             expect(validationResult.items[1].items[2].isError()).toBe(true);
         });
 
-        test("throws on terminated relationship", async function () {
+        test("returns 'error' on terminated relationship", async function () {
```
jkoenig134, line 387: now this sounds like the method returns a string 'error' ...

Magnus-Kuhn: I'm just using the style of this file

jkoenig134 (dismissed):

packages/transport/test/end2end/End2End.test.ts
Magnus-Kuhn: I've decided to follow the existing style

Magnus-Kuhn: I don't mind those comments, I like them for structure, but it seems I'm outvoted. Now I'm curious about your opinion on the SyncController tests.

jkoenig134 (approved):

PR #119, opened 2024-04-30
Refactor/typesafe external events

sebbi08:


packages/transport/src/modules/sync/backbone/BackboneExternalEvent.ts
```diff
@@ -1,8 +1,35 @@
...
+    @validate()
+    public index: number;
+    @serialize()
+    @validate()
+    public createdAt: string;
+    @serialize()
+    @validate()
+    public syncErrorCount: number;
+    @serialize()
+    @validate()
```
jkoenig134, line 30: where are all the newlines? :D

packages/transport/src/modules/sync/SyncController.ts
```diff
@@ -410,14 +411,15 @@ export class SyncController extends TransportController {
         const changedItems = new ChangedItems();
 
         for (const externalEvent of externalEvents) {
+            const externalEventObject = BackboneExternalEvent.fromAny(externalEvent);
```
slavistan, line 414: Does this require a `try {} catch {}` block, as below?

packages/transport/src/modules/sync/externalEventProcessors/MessageDeliveredExternalEventProcessor.ts
```diff
@@ -1,11 +1,19 @@
+import { serialize, validate } from "@js-soft/ts-serval";
+import { CoreSerializable } from "../../../core/CoreSerializable";
 import { MessageDeliveredEvent } from "../../../events";
 import { Message } from "../../messages/local/Message";
 import { BackboneExternalEvent } from "../backbone/BackboneExternalEvent";
 import { ExternalEventProcessor } from "./ExternalEventProcessor";
 
+export class MessageDeliveredExternalEventData extends CoreSerializable {
```
jkoenig134, line 8: why is this exported? And why CoreSerializable instead of Serializable?

same for all other EventDatas

packages/transport/src/modules/sync/externalEventProcessors/ExternalEventProcessor.ts
```diff
@@ -1,10 +1,12 @@
 import { EventBus } from "@js-soft/ts-utils";
+import { CoreSerializable } from "../../../core";
 import { AccountController } from "../../accounts/AccountController";
 import { Message } from "../../messages/local/Message";
 import { Relationship } from "../../relationships/local/Relationship";
 import { BackboneExternalEvent } from "../backbone/BackboneExternalEvent";
 
 export type ExternalEventProcessorConstructor = new (eventBus: EventBus, accountController: AccountController) => ExternalEventProcessor;
+export type ExternalEventProcessorDataConstructor = new (...any: any[]) => CoreSerializable;
```
jkoenig134, line 9: I don't see where this is used

packages/transport/src/modules/sync/backbone/SyncClient.ts
jkoenig134, line 1: as said in the other file - undo pls

jkoenig134 (changes requested):

packages/transport/src/modules/sync/externalEventProcessors/MessageDeliveredExternalEventProcessor.ts
sebbi08: I didnt know there is `Serializable`. What is the difference between `CoreSerializable` and `Serializable`

packages/transport/src/modules/sync/SyncController.ts
jkoenig134: It really does. in case of a deserialization error we have to send the backbone that the external event could not be processed.

packages/transport/src/modules/sync/externalEventProcessors/MessageDeliveredExternalEventProcessor.ts
jkoenig134: I think CoreSerializable adds some functionalities on serval's Serializable. IDK where and if we even need that.

packages/transport/src/modules/sync/SyncController.ts
sebbi08: Is there already an API to do that ?

jkoenig134: see below in the catch block

sebbi08: ah ok than i would just move the parsing in the try catch below

packages/transport/src/modules/sync/data/ExternalEvent.ts
```diff
@@ -0,0 +1,26 @@
...
+    public index: number;
+    @serialize()
+    @validate()
+    public createdAt: string;
+    @serialize()
+    @validate()
+    public syncErrorCount: number;
+    @serialize()
+    @validate()
+    public payload: object;
```
jkoenig134, line 22: as in my other comment - where are the newlines gone? :D

packages/transport/src/modules/sync/SyncController.ts
```diff
@@ -410,14 +411,15 @@ export class SyncController extends TransportController {
         const changedItems = new ChangedItems();
 
         for (const externalEvent of externalEvents) {
+            const externalEventObject = ExternalEvent.fromAny(externalEvent);
```
jkoenig134, line 414: shoudn't this be inside the try catch?

jkoenig134 (approved):

PR #120, opened 2024-05-02
Feature/identity deletion

sebbi08:


packages/runtime/src/events/transport/IdentityDeletionProcessStatusChangedEvent.ts
```diff
@@ -0,0 +1,10 @@
+import { IdentityDeletionProcessDTO } from "../../types/transport/IdentityDeletionProcessDTO";
+import { DataEvent } from "../DataEvent";
+
+export class IdentityDeletionProcessStatusChangedEvent extends DataEvent<IdentityDeletionProcessDTO> {
+    public static readonly namespace = "tansport.identityDeletionProcessStatusChangedEvent";
+
+    public constructor(eventTargetAddress: string, data: IdentityDeletionProcessDTO) {
```
slavistan, line 7: Just a question for my own understanding: What's the point and purpose of `eventTargetAddress`?

packages/transport/src/events/IdentityDeletionProcessStatusChangedEvent.ts
```diff
@@ -0,0 +1,10 @@
+import { IdentityDeletionProcess } from "../modules";
+import { TransportDataEvent } from "./TransportDataEvent";
+
+export class IdentityDeletionProcessStatusChangedEvent extends TransportDataEvent<IdentityDeletionProcess> {
+    public static readonly namespace = "tansport.identityDeletionProcessStatusChangedEvent";
```
slavistan, line 5: t**r**ansport

packages/transport/src/modules/accounts/IdentityDeletionProcessController.ts
```diff
@@ -0,0 +1,89 @@
...
+    public override async init(): Promise<IdentityDeletionProcessController> {
+        await super.init();
+
+        this.identityDeletionProcessCollection = await this.parent.getSynchronizedCollection("IdentityDeletionProcesses");
+        return this;
+    }
+
+    public async getIdentityDeletionProcess(identityDeletionProcessId: string): Promise<IdentityDeletionProcess | undefined> {
+        const identityDeletionProcess = await this.identityDeletionProcessCollection.findOne({ id: identityDeletionProcessId });
+        return identityDeletionProcess ? IdentityDeletionProcess.from(identityDeletionProcess) : undefined;
```
slavistan, line 30: `a ? b : undefined` is equivalent to `a ?? b`, right?

packages/transport/src/modules/accounts/data/IdentityDeletionProcess.ts
```diff
@@ -0,0 +1,116 @@
...
+    public static from(value: IIdentityDeletionProcess | IdentityDeletionProcessJSON): IdentityDeletionProcess {
+        return this.fromAny(value);
+    }
+}
+
+/**
+ * Fragen:
+ *
+ * TODO: Wie erhält App Benachrichtigung über ausstehenden DeletionProcess (Admin UI)?
+ */
```
slavistan, line 116: Can be removed; the runtime will receive an external event which we map to a runtime event for the app to react to.

packages/runtime/test/lib/testUtils.ts
```diff
@@ -74,46 +80,50 @@ export async function syncUntil(transportServices: TransportServices, until: (sy
...
+    filter: (r: SyncEverythingResponse[T][0]) => boolean
+): Promise<SyncEverythingResponse[T][0]> {
+    const syncResult = await syncUntil(transportServices, (syncResult) => syncResult[key].some((r) => filter(r)));
+    return syncResult[key].filter(filter)[0];
+}
+
+async function syncUntilHasMany<T extends keyof SyncEverythingResponse>(
+    transportServices: TransportServices,
+    key: T,
+    expectedNumberOfMessages = 1
```
jkoenig134, line 96: messages?

packages/runtime/test/lib/testUtils.ts
```diff
@@ -52,7 +54,7 @@ import { TestRuntimeServices } from "./RuntimeServiceProvider";
 import { TestNotificationItem } from "./TestNotificationItem";
 
 export async function syncUntil(transportServices: TransportServices, until: (syncResult: SyncEverythingResponse) => boolean): Promise<SyncEverythingResponse> {
-    const finalSyncResult: SyncEverythingResponse = { messages: [], relationships: [] };
+    const finalSyncResult: any = {};
```
jkoenig134, line 57: avoid any

packages/runtime/test/lib/testUtils.ts
```diff
@@ -615,3 +625,36 @@ export async function waitForEvent<TEvent>(
...
+export async function startIdentityDeletionProcessFromBackboneAdminApi(transportService: TransportServices, accountAddress: string): Promise<IdentityDeletionProcessDTO> {
+    const adminApiClient = await getBackboneAdminApiClient();
+    const deletionProcess = await adminApiClient.post<{ result: { id: string } }>(`http://localhost:8091/api/v1/Identities/${accountAddress}/DeletionProcesses`);
+
+    await syncUntilHasIdentityDeletionProcess(transportService, deletionProcess.data.result.id);
+
+    const activeIdentityDeletionProcess = await transportService.identityDeletionProcesses.getIdentityDeletionProcess({});
+
+    return activeIdentityDeletionProcess.value;
+}
```
jkoenig134, line 660: I'd prefer to have this in another file

packages/runtime/test/lib/testUtils.ts
```diff
@@ -615,3 +625,36 @@ export async function waitForEvent<TEvent>(
...
+    return adminClient;
+}
+
+export async function startIdentityDeletionProcessFromBackboneAdminApi(transportService: TransportServices, accountAddress: string): Promise<IdentityDeletionProcessDTO> {
+    const adminApiClient = await getBackboneAdminApiClient();
+    const deletionProcess = await adminApiClient.post<{ result: { id: string } }>(`http://localhost:8091/api/v1/Identities/${accountAddress}/DeletionProcesses`);
+
+    await syncUntilHasIdentityDeletionProcess(transportService, deletionProcess.data.result.id);
+
+    const activeIdentityDeletionProcess = await transportService.identityDeletionProcesses.getIdentityDeletionProcess({});
```
jkoenig134, line 657: this is exactly a UseCase where I want to see a specific UseCase for getting the active IdentityDeletionProcess.

packages/transport/src/modules/accounts/IdentityController.ts
```diff
@@ -3,9 +3,13 @@ import { CoreBuffer, CryptoSignature, CryptoSignaturePrivateKey, CryptoSignature
 import { ControllerName, CoreAddress, CoreCrypto, CoreErrors, TransportController } from "../../core";
 import { AccountController } from "../accounts/AccountController";
 import { DeviceSecretType } from "../devices/DeviceSecretController";
-import { Identity, Realm } from "./data/Identity";
+import { SynchronizedCollection } from "../sync/SynchronizedCollection";
+import { Identity } from "./data/Identity";
+import { Realm } from "./data/Realm";
 
 export class IdentityController extends TransportController {
+    public identityDeletionProcessCollection: SynchronizedCollection;
```
jkoenig134, line 11: I think this is unused.

packages/transport/src/modules/accounts/IdentityController.ts
```diff
@@ -3,9 +3,13 @@ import { CoreBuffer, CryptoSignature, CryptoSignaturePrivateKey, CryptoSignature
 import { ControllerName, CoreAddress, CoreCrypto, CoreErrors, TransportController } from "../../core";
 import { AccountController } from "../accounts/AccountController";
 import { DeviceSecretType } from "../devices/DeviceSecretController";
-import { Identity, Realm } from "./data/Identity";
+import { SynchronizedCollection } from "../sync/SynchronizedCollection";
+import { Identity } from "./data/Identity";
+import { Realm } from "./data/Realm";
```
jkoenig134, line 8: Why was Realm moved?

packages/transport/src/modules/accounts/IdentityDeletionProcessController.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public async getWaitingIdentityDeletionProcess(): Promise<IdentityDeletionProcess | undefined> {
+        const identityDeletionProcess = await this.identityDeletionProcessCollection.findOne({
+            status: IdentityDeletionProcessStatus.WaitingForApproval
+        });
+
+        return identityDeletionProcess ? IdentityDeletionProcess.from(identityDeletionProcess) : undefined;
+    }
```
jkoenig134, line 52: I would vote for just one method that let's you pass a IdentityDeletionProcessStatus. Maybe allow more than one status by spread operator, would be easy to simplify the part in `packages/runtime/src/useCases/transport/identityDeletionProcesses/GetIdentityDeletionProcess.ts` where you query two times which could be one operation.

packages/transport/src/modules/accounts/IdentityDeletionProcessController.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    private async updateIdentityDeletionProcess(identityDeletionProcess: IdentityDeletionProcess): Promise<void> {
+        const oldIdentityDeletion = await this.identityDeletionProcessCollection.findOne({ id: identityDeletionProcess.id.toString() });
+        await this.identityDeletionProcessCollection.update(oldIdentityDeletion, identityDeletionProcess);
```
jkoenig134, line 56: Please be VERY cautious with the SynchronizedCollection. You have to work _**ANYTIME**_ with CoreSynchronizable Objects - typescript doesn't detect this error, but if you would have tested this with synchronization enabled that should have blown up.

jkoenig134: (to fix this pls call getIdentityDeletionProcess in the line before)

and make sure that the process is defined.

packages/transport/src/modules/accounts/IdentityDeletionProcessController.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    public async getWaitingIdentityDeletionProcess(): Promise<IdentityDeletionProcess | undefined> {
+        const identityDeletionProcess = await this.identityDeletionProcessCollection.findOne({
+            status: IdentityDeletionProcessStatus.WaitingForApproval
+        });
+
+        return identityDeletionProcess ? IdentityDeletionProcess.from(identityDeletionProcess) : undefined;
+    }
+
+    private async updateIdentityDeletionProcess(identityDeletionProcess: IdentityDeletionProcess): Promise<void> {
```
jkoenig134, line 54: the private method should be beyond the callee

packages/transport/src/modules/accounts/backbone/IdentityClient.ts
```diff
@@ -1,31 +1,11 @@
...
-    createdAt: string;
-}
+import { BackboneIdentityPostRequest, BackboneIdentityPostResponse } from "./BackbonePostIdentity";
 
 export class IdentityClient extends RESTClient {
     protected override _logDirective = RESTClientLogDirective.LogResponse;
 
-    public async createIdentity(value: PlatformIdentityPostRequest): Promise<ClientResult<PlatformIdentityPostResponse>> {
-        return await this.post<PlatformIdentityPostResponse>("/api/v1/Identities", value, {});
+    public async createIdentity(value: BackboneIdentityPostRequest): Promise<ClientResult<BackboneIdentityPostResponse>> {
```
jkoenig134, line 8: I think naming is a bit off here

should be `BackbonePostIdentityRequest` and `BackbonePostIdentityResponse`

packages/transport/src/modules/accounts/backbone/IdentityDeletionProcessClient.ts
```diff
@@ -0,0 +1,27 @@
...
+        return await this.put<IdentityDeletionProcessJSON>(`/api/v1/Identities/Self/DeletionProcesses/${identityDeletionProcessId}/Cancel`, {});
+    }
+
+    public async approveIdentityDeletionProcess(identityDeletionProcessId: string): Promise<ClientResult<IdentityDeletionProcessJSON>> {
+        return await this.put<IdentityDeletionProcessJSON>(`/api/v1/Identities/Self/DeletionProcesses/${identityDeletionProcessId}/Approve`, {});
+    }
+
+    public async rejectIdentityDeletionProcess(identityDeletionProcessId: string): Promise<ClientResult<IdentityDeletionProcessJSON>> {
+        return await this.put<IdentityDeletionProcessJSON>(`/api/v1/Identities/Self/DeletionProcesses/${identityDeletionProcessId}/Reject`, {});
+    }
```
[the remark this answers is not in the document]
jkoenig134: I'v written a very long explanation why this is required, but it got deleted.

Methods shouldn't "inherit" context from their classes as you do not call `ClassName.get()` - you call `{whereTheInstanceOfTheClassIsSaved}.get()`.

In this case the field where the client is saved could only be named client - `client.initiate()` would not be enough to explain what's going on.

You can see exactly the same concept on all clients and also controllers of the whole codebase.

packages/transport/src/modules/accounts/backbone/IdentityDeletionProcessClient.ts
jkoenig134, line 1: @tnotheis would you please review the client and requests / responses?

packages/transport/src/modules/accounts/data/IdentityDeletionProcess.ts
jkoenig134, line 1: please remove the comments, we do not use structural comments.

packages/transport/src/modules/sync/SyncController.ts
```diff
@@ -412,6 +412,10 @@ export class SyncController extends TransportController {
 
         for (const externalEvent of externalEvents) {
             try {
+                // TODO: remove once backbone dose not fire this event anymore
+                if (externalEvent.type === "IdentityDeletionProcessStatusChanged") {
+                    continue;
+                }
```
jkoenig134, line 418: actually we should handle this, or not?

jkoenig134: if not this should be done asap as I don't want to ship this TODO comment @tnotheis

packages/transport/src/modules/sync/externalEventProcessors/IdentityDeletionProcessStartedEventProcessor.ts
```diff
@@ -0,0 +1,28 @@
...
+    public override async execute(externalEvent: BackboneExternalEvent): Promise<IdentityDeletionProcess> {
+        const messageReceivedPayload = IdentityDeletionProcessStartedEventData.fromAny(externalEvent.payload);
+        const identityDeletionProcessJSONResponse = await this.accountController.identityDeletionProcess.identityDeletionProcessClient.getIdentityDeletionProcess(
+            messageReceivedPayload.deletionProcessId
+        );
+
+        const identityDeletionProcess = IdentityDeletionProcess.from(identityDeletionProcessJSONResponse.value);
+        await this.accountController.identityDeletionProcess.identityDeletionProcessCollection.create(identityDeletionProcess);
+
+        // TODO: is it intended that we publish an Identity...ChangedEvent, even though we receive an Identity...StartedEvent?
```
jkoenig134, line 23: yep

packages/transport/test/testHelpers/TestUtil.ts
```diff
@@ -385,17 +388,19 @@ export class TestUtil {
      * the `until` callback is met.
      */
     public static async syncUntil(accountController: AccountController, until: (syncResult: ChangedItems) => boolean): Promise<ChangedItems> {
-        const { messages, relationships } = await accountController.syncEverything();
-        const syncResult = new ChangedItems([...relationships], [...messages]);
-
+        const syncResult: any = new ChangedItems();
```
jkoenig134, line 391: any

packages/transport/test/testHelpers/TestUtil.ts
```diff
@@ -404,24 +409,36 @@ export class TestUtil {
         return syncResult;
     }
 
+    public static async syncUntilHas<T extends keyof IChangedItems>(accountController: AccountController, id: CoreId, key: T): Promise<ChangedItems[T]> {
+        const syncResult = await TestUtil.syncUntil(accountController, (syncResult) => syncResult[key].some((r) => r.id.equals(id)));
+        return syncResult[key];
+    }
+    public static async syncUntilHasMany<T extends keyof IChangedItems>(accountController: AccountController, key: T, expectedNumberOfMessages = 1): Promise<ChangedItems[T]> {
```
jkoenig134, line 416: newline

packages/transport/test/testHelpers/TestUtil.ts
```diff
@@ -404,24 +409,36 @@ export class TestUtil {
...
+        const syncResult = await TestUtil.syncUntil(accountController, (syncResult) => syncResult[key].length >= expectedNumberOfMessages);
+        return syncResult[key];
+    }
+
+    public static async syncUntilHasIdentityDeletionProcess(accountController: AccountController, id: CoreId): Promise<IdentityDeletionProcess[]> {
+        return await TestUtil.syncUntilHas(accountController, id, "identityDeletionProcesses");
+    }
+    public static async syncUntilHasIdentityDeletionProcesses(accountController: AccountController): Promise<IdentityDeletionProcess[]> {
```
jkoenig134, line 424: newline

packages/transport/test/testHelpers/TestUtil.ts
```diff
@@ -547,4 +564,41 @@ export class TestUtil {
...
+        );
+        await TestUtil.syncUntilHasIdentityDeletionProcess(account, CoreId.from(deletionProcess.data.result.id));
+
+        const activeIdentityDeletionProcess = await account.identityDeletionProcess.getWaitingIdentityDeletionProcess();
+        if (typeof activeIdentityDeletionProcess === "undefined") {
+            throw new Error("Identity deletion process not found.");
+        }
+
+        return activeIdentityDeletionProcess;
+    }
```
jkoenig134, line 603: please move to external file

packages/runtime/test/transport/identityDeletionProcess.test.ts
```diff
@@ -0,0 +1,152 @@
...
+const serviceProvider = new RuntimeServiceProvider();
+let transportService: TransportServices;
+let accountAddress: string;
+
+beforeEach(async () => {
+    const runtimeServices = await serviceProvider.launch(1);
+    transportService = runtimeServices[0].transport;
+    accountAddress = runtimeServices[0].address;
+}, 30000);
+afterEach(async () => await serviceProvider.stop());
```
jkoenig134, line 22: there **must** be a better cleanup than creating one identity per test case.

jkoenig134: for example you can reject an open Process in afterEach when on exists.

jkoenig134 (changes requested):

packages/transport/src/modules/sync/SyncController.ts
tnotheis: Well, an external event of this type is always created. The problem is that they are currently created for in case of ANY status change. But we only want them for cases where it was triggered by the support (and therefore where it is really external).

packages/app-runtime/test/lib/TestUtil.ts
```diff
@@ -300,19 +300,21 @@ export class TestUtil {
      * specified in the `until` callback is met.
      */
     public static async syncUntil(session: LocalAccountSession, until: (syncResult: SyncEverythingResponse) => boolean): Promise<SyncEverythingResponse> {
-        const syncResult = await session.transportServices.account.syncEverything();
-
-        const { messages, relationships } = syncResult.value;
-        const syncResponse = { relationships: [...relationships], messages: [...messages] };
+        const syncResponse: any = {};
```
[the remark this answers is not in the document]
sebbi08: I made the method generic so that in future it dose not need to be changed when we add new properties to `SyncEverythingResponse`. The any is used because otherwise there will be TS errors.

packages/runtime/src/useCases/transport/identityDeletionProcesses/GetIdentityDeletionProcess.ts
```diff
@@ -0,0 +1,45 @@
...
+        if (typeof request.id === "undefined") {
+            const activeIdentityDeletionProcess =
+                (await this.identityDeletionProcessController.getApprovedIdentityDeletionProcess()) ??
+                (await this.identityDeletionProcessController.getWaitingIdentityDeletionProcess());
+            if (typeof activeIdentityDeletionProcess === "undefined") {
+                return Result.fail(RuntimeErrors.identity.noActiveIdentityDeletionProcess());
+            }
+
+            return Result.ok(IdentityDeletionProcessMapper.toIdentityDeletionProcessDTO(activeIdentityDeletionProcess));
+        }
```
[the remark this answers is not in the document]
sebbi08: we made the parameter optional to combine both usecases

packages/runtime/src/useCases/transport/identityDeletionProcesses/GetIdentityDeletionProcess.ts
```diff
@@ -0,0 +1,45 @@
+import { Result } from "@js-soft/ts-utils";
+import { IdentityDeletionProcess, IdentityDeletionProcessController } from "@nmshd/transport";
+import { Inject } from "typescript-ioc";
+import { IdentityDeletionProcessDTO } from "../../../types/transport/IdentityDeletionProcessDTO";
+import { IdentityDeletionProcessIdString, RuntimeErrors, SchemaRepository, SchemaValidator, UseCase } from "../../common";
+import { IdentityDeletionProcessMapper } from "./IdentityDeletionProcessMapper";
+
+export interface GetIdentityDeletionProcessRequest {
+    id?: IdentityDeletionProcessIdString;
```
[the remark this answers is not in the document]
sebbi08: we made the parameter optional to combine both usecases

packages/runtime/test/lib/testUtils.ts
sebbi08: I made the method generic so that in future it dose not need to be changed when we add new properties to SyncEverythingResponse. The any is used because otherwise there will be TS errors.

jkoenig134: yeah - please still avoid any!

packages/transport/src/modules/accounts/IdentityController.ts
sebbi08: it was to prevent cyclic dependencies when we had a dependency from Identity to IdentiyDeletionProcess

packages/runtime/test/transport/identityDeletionProcess.test.ts
```diff
@@ -8,18 +8,34 @@ import {
...
     const runtimeServices = await serviceProvider.launch(1);
     transportService = runtimeServices[0].transport;
     accountAddress = runtimeServices[0].address;
 }, 30000);
-afterEach(async () => await serviceProvider.stop());
+
+afterAll(async () => {
+    return await serviceProvider.stop();
+});
+afterEach(async () => {
```
jkoenig134, line 27: missing newline

packages/runtime/test/lib/AdminApiClient.ts
```diff
@@ -0,0 +1,36 @@
...
+            "x-api-key": process.env.NMSHD_TEST_ADMIN_API_KEY!
+            /* eslint-enable @typescript-eslint/naming-convention */
+        }
+    });
+    return adminClient;
+}
+
+export async function startIdentityDeletionProcessFromBackboneAdminApi(transportService: TransportServices, accountAddress: string): Promise<IdentityDeletionProcessDTO> {
+    const adminApiClient = await getBackboneAdminApiClient();
+    const deletionProcess = await adminApiClient.post<{ result: { id: string } }>(`http://localhost:8091/api/v1/Identities/${accountAddress}/DeletionProcesses`);
```
jkoenig134, line 29:
```suggestion
    const deletionProcess = await adminApiClient.post<{ result: { id: string } }>(`/api/v1/Identities/${accountAddress}/DeletionProcesses`);
```

packages/runtime/test/lib/testUtils.ts
sebbi08: Ok but than we need to go back to the old implementation.
Or use `// @ts-expect-error`

jkoenig134: fo me the old one was more readable.

packages/transport/src/modules/sync/SyncController.ts
tnotheis: This is fixed with https://github.com/nmshd/backbone/pull/638.
I will create a new version as soon as it is merged.

tnotheis: No. We will only create external events in case of an action that was triggered by something external. So if the deletion process was started by the support or if it was cancelled by the support or our background job.
But I don't think you need one in case of an approval, do you?

sebbi08: Yea but on runtime side this would have no action to do

jkoenig134: eeehm, you have to update the local identity deletion process.

sebbi08: No because all those "changes" are triggered by us, and we already updated the process. Only the start over the Admin UI is not handled by us.

packages/runtime/test/transport/identityDeletionProcess.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+        test("should return an error trying to get an Identity deletion process without specifying an ID if there is none active", async function () {
+            const result = await transportService.identityDeletionProcesses.getIdentityDeletionProcess({});
+            expect(result).toBeAnError("No active identity deletion process found.", "error.runtime.identity.noActiveIdentityDeletionProcess");
+        });
+
+        test("should return an error trying to get an Identity deletion process specifying an unknown ID", async function () {
+            const unknownId = (await CoreId.generate("IDP")).toString();
+            const result = await transportService.identityDeletionProcesses.getIdentityDeletionProcess({ id: unknownId });
+            expect(result).toBeAnError("IdentityDeletionProcess not found. Make sure the ID exists and the record is not expired.", "error.runtime.recordNotFound");
```
britsta, line 101: I would prefer the spelling ‘IdentityDeletionProcess’ used here to be used in all other error messages too instead of ‘identity deletion process’. However, if you decide to keep ‘identity deletion process’ in the other error messages for now, I would at least consistently use ‘Identity deletion process’ instead of ‘identity deletion process’ (as you already do in the tests' names).

packages/runtime/test/transport/identityDeletionProcess.test.ts
```diff
@@ -0,0 +1,152 @@
...
+        });
+
+        test("should return an error trying to cancel an Identity deletion process if there is none active", async function () {
+            const result = await transportService.identityDeletionProcesses.cancelIdentityDeletionProcess();
+            expect(result).toBeAnError("No active identity deletion process found.", "error.runtime.identity.noActiveIdentityDeletionProcess");
+        });
+    });
+
+    describe(ApproveIdentityDeletionProcessUseCase.name, () => {
+        test("should approve an waiting Identity deletion process", async function () {
```
britsta, line 148: "a waiting" instead of "an waiting"

Would also avoid talking about a "waiting" IdentityDeletionProcess and instead clarify that it is about an Identity deletion process that is "waiting for approval".

packages/runtime/test/transport/identityDeletionProcess.test.ts
```diff
@@ -0,0 +1,152 @@
...
+        });
+
+        test("should return an error trying to approve an Identity deletion process if there is none active", async function () {
+            const result = await transportService.identityDeletionProcesses.approveIdentityDeletionProcess();
+            expect(result).toBeAnError("No identity deletion process waiting for approval found.", "error.runtime.identity.noWaitingIdentityDeletionProcess");
+        });
+    });
+
+    describe(RejectIdentityDeletionProcessUseCase.name, () => {
+        test("should reject an waiting Identity deletion process", async function () {
```
britsta, line 164: "a waiting" instead of "an waiting"

Would also avoid talking about a "waiting" IdentityDeletionProcess and instead clarify that it is about an Identity deletion process that is "waiting for approval".

packages/transport/test/modules/account/IdentityDeletionProcessController.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        await TestUtil.startIdentityDeletionProcessFromBackboneAdminApi(account);
+        await expect(TestUtil.startIdentityDeletionProcessFromBackboneAdminApi(account)).rejects.toThrow("Request failed with status code 400");
+    });
+
+    test("should not start an Identity deletion process from the Backbone admin API if one was already started by the user", async function () {
+        await account.identityDeletionProcess.initiateIdentityDeletionProcess();
+        await expect(TestUtil.startIdentityDeletionProcessFromBackboneAdminApi(account)).rejects.toThrow("Request failed with status code 400");
+    });
+
+    test("should get the waiting Identity deletion process", async function () {
```
britsta, line 83: Would avoid talking about a "waiting" IdentityDeletionProcess and instead clarify that it is about an Identity deletion process that is "waiting for approval".

Same for the variable name `waitingIdentityDeletionProcess` below.

packages/transport/test/modules/account/IdentityDeletionProcessController.test.ts
```diff
@@ -0,0 +1,116 @@
...
+        expect(result.id.toString()).toBe(startedIdentityDeletionProcess.id.toString());
+        expect(result.status).toBe(IdentityDeletionProcessStatus.Approved);
+
+        const approvedIdentityDeletionProcess = await account.identityDeletionProcess.getApprovedIdentityDeletionProcess();
+        expect(approvedIdentityDeletionProcess).toBeDefined();
+        expect(approvedIdentityDeletionProcess!.status).toBe(IdentityDeletionProcessStatus.Approved);
+        expect(approvedIdentityDeletionProcess!.id.toString()).toBe(result.id.toString());
+    });
+
+    test("should reject an Identity deletion proces that was started from the Backbone admin APIs", async function () {
```
britsta, line 106: "Identity deletion process" instead "Identity deletion proces", "Backbone admin API" instead of "Backbone admin APIs"

packages/transport/src/modules/sync/SyncController.ts
tnotheis: Version 5.3.0 does now exist and contains the changes mentioned above.

packages/transport/src/modules/accounts/backbone/IdentityDeletionProcessClient.ts
tnotheis: There are two things I noticed:

Theoretically, there is an endpoint to get all deletion processes (GET /api/v1/Identities/Self/DeletionProcesses). Just saying. Of course, if you don't need it, you don't have to call it.

The responses you currently define for the POST and PUT routes currently don't match the actual responses. That's because we always wanted to only return the properties that changed during the action. But actually, that's not even the case. Some properties seem to have been missed, while others are returned even though they didn't even change.
I will fix this hopefully today and deploy a new version. And I think instead of returning only changed properties, I will return the whole deletion process, so that it matches your responses.

packages/transport/src/modules/accounts/data/IdentityDeletionProcess.ts
```diff
@@ -0,0 +1,101 @@
...
+    id: string;
+    status: IdentityDeletionProcessStatus;
+    createdAt?: string;
+    createdByDevice?: string;
+    approvedAt?: string;
+    approvedByDevice?: string;
+    gracePeriodEndsAt?: string;
+    cancelledAt?: string;
+    cancelledByDevice?: string;
+    rejectedAt?: string;
```
britsta, line 16: As with the order of the fields of `IdentityDeletionProcessStatus` (the field `Rejected` is placed before the field `Approved`), I would also place the two fields `rejectedAt?` and `rejectedByDevice?` before `approvedAt?` here, to avoid confusion with the `cancelledAt?` and `cancelledByDevice?` fields. Same below and in the IdentityDeletionProcessMapper and the IdentityDeletionProcessDTO.

packages/transport/src/modules/sync/externalEventProcessors/IdentityDeletionProcessStartedEventProcessor.ts
```diff
@@ -0,0 +1,27 @@
+import { Serializable, serialize, validate } from "@js-soft/ts-serval";
+import { IdentityDeletionProcessStatusChangedEvent } from "../../../events";
+import { IdentityDeletionProcess } from "../../accounts/data/IdentityDeletionProcess";
+import { BackboneExternalEvent } from "../backbone/BackboneExternalEvent";
+import { ExternalEventProcessor } from "./ExternalEventProcessor";
+
+class IdentityDeletionProcessStartedEventData extends Serializable {
+    @serialize()
+    @validate()
+    public deletionProcessId: string;
```
britsta, line 10: "identityDeletionProcessId" instead of "deletionProcessId"

packages/runtime/src/types/transport/IdentityDeletionProcessDTO.ts
```diff
@@ -0,0 +1,15 @@
...
+    createdAt?: string;
+    createdByDevice?: string;
+    approvedAt?: string;
+    approvedByDevice?: string;
+    gracePeriodEndsAt?: string;
+    cancelledAt?: string;
+    cancelledByDevice?: string;
+    rejectedAt?: string;
+    rejectedByDevice?: string;
+    status: IdentityDeletionProcessStatus;
```
britsta, line 14: I would change the order of the fields, so that `status` is at the same place like in the IdentityDeletionProcess.

packages/transport/test/testHelpers/AdminApiClient.ts
```diff
@@ -0,0 +1,40 @@
...
+import { TestUtil } from "./TestUtil";
+
+export class AdminApiClient {
+    private static adminClient: Axios | undefined;
+
+    public static async getBackboneAdminApiClient(): Promise<Axios> {
+        if (AdminApiClient.adminClient) {
+            return AdminApiClient.adminClient;
+        }
+        const adminAPIBaseUrl = process.env.NMSHD_TEST_BASEURL_ADMIN_API!;
```
slavistan, line 12: The trailing bang `!` is unnecessary since you're checking for `null`/`undefined` in the next line.

packages/transport/test/testHelpers/AdminApiClient.ts
```diff
@@ -0,0 +1,40 @@
...
+        if (!adminAPIBaseUrl) throw new Error("Missing environment variable NMSHD_TEST_BASEURL_ADMIN_API");
+        const csrf = await axios.get(`${adminAPIBaseUrl}/api/v1/xsrf`);
+        AdminApiClient.adminClient = axios.create({
+            baseURL: adminAPIBaseUrl,
+            headers: {
+                /* eslint-disable @typescript-eslint/naming-convention */
+                cookie: csrf.headers["set-cookie"],
+                "x-xsrf-token": csrf.data,
+                "x-api-key": process.env.NMSHD_TEST_ADMIN_API_KEY!
```
slavistan, line 21: Since we throw an error if the above envvar `NMSHD_TEST_BASEURL_ADMIN_API` doesn't exist, I'd say we do th same for `NMSHD_TEST_ADMIN_API_KEY`.

sebbi08:
Na the API key is localy not required only optional

packages/transport/test/testHelpers/AdminApiClient.ts
sebbi08: Na the API key is localy not required only optional

packages/transport/src/modules/accounts/IdentityDeletionProcessController.ts
```diff
@@ -0,0 +1,90 @@
...
+        await super.init();
+
+        this.identityDeletionProcessCollection = await this.parent.getSynchronizedCollection("IdentityDeletionProcesses");
+        return this;
+    }
+
+    private async updateIdentityDeletionProcess(identityDeletionProcess: IdentityDeletionProcess): Promise<void> {
+        const oldIdentityDeletion = await this.getIdentityDeletionProcess(identityDeletionProcess.id.toString());
+        await this.identityDeletionProcessCollection.update(oldIdentityDeletion, identityDeletionProcess);
+        this.eventBus.publish(new IdentityDeletionProcessStatusChangedEvent(this.parent.identity.address.toString(), identityDeletionProcess));
```
britsta, line 31: I assume that when the underlying Backbone routes are called, it is ensured that a status change has actually taken place? I'm not sure if this is appropriate in this file, but perhaps you could still check here whether the status of `oldIdentityDeletionProcess` is different from the status of the new `identityDeletionProcess`.

packages/transport/src/modules/accounts/IdentityDeletionProcessController.ts
```diff
@@ -0,0 +1,90 @@
...
+    @log()
+    public override async init(): Promise<IdentityDeletionProcessController> {
+        await super.init();
+
+        this.identityDeletionProcessCollection = await this.parent.getSynchronizedCollection("IdentityDeletionProcesses");
+        return this;
+    }
+
+    private async updateIdentityDeletionProcess(identityDeletionProcess: IdentityDeletionProcess): Promise<void> {
+        const oldIdentityDeletion = await this.getIdentityDeletionProcess(identityDeletionProcess.id.toString());
```
britsta, line 29: `oldIdentityDeletionProcess` instead of `oldIdentityDeletion`.

packages/transport/test/testHelpers/AdminApiClient.ts
```diff
@@ -0,0 +1,40 @@
...
+                /* eslint-enable @typescript-eslint/naming-convention */
+            }
+        });
+        return AdminApiClient.adminClient;
+    }
+
+    public static async startIdentityDeletionProcessFromBackboneAdminApi(account: AccountController): Promise<IdentityDeletionProcess> {
+        const adminApiClient = await AdminApiClient.getBackboneAdminApiClient();
+        const deletionProcess = await adminApiClient.post<{ result: { id: string } }>(`/api/v1/Identities/${account.identity.address.toString()}/DeletionProcesses`);
```
britsta, line 30: `identityDeletionProcess` instead of `deletionProcess`.

packages/runtime/test/lib/AdminApiClient.ts
britsta, line 29: `identityDeletionProcess` instead of `deletionProcess`.

packages/transport/src/modules/sync/externalEventProcessors/IdentityDeletionProcessStartedEventProcessor.ts
britsta: Alright, I see!

packages/runtime/src/useCases/transport/identityDeletionProcesses/InitiateIdentityDeletionProcess.ts
```diff
@@ -0,0 +1,30 @@
...
+export class InitiateIdentityDeletionProcessUseCase extends UseCase<void, IdentityDeletionProcessDTO> {
+    public constructor(
+        @Inject private readonly identityDeletionProcessController: IdentityDeletionProcessController,
+        @Inject private readonly accountController: AccountController
+    ) {
+        super();
+    }
+
+    protected async executeInternal(): Promise<Result<IdentityDeletionProcessDTO>> {
+        const activeIdentityDeletionProcess =
```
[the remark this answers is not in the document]
sebbi08: No, when there is one IdentiyDeletionProcess with status Deleting the identity should not be able to anything else than cancel that deletion

packages/runtime/src/extensibility/facades/transport/IdentityDeletionProcessesFacade.ts
```diff
@@ -10,6 +10,7 @@ import {
     InitiateIdentityDeletionProcessUseCase,
     RejectIdentityDeletionProcessUseCase
 } from "../../../useCases";
+import { GetActiveIdentityDeletionProcessUseCase } from "../../../useCases/transport/identityDeletionProcesses/GetActiveIdentityDeletionProcess";
```
jkoenig134, line 13: missing export in barrel file?

packages/transport/test/testHelpers/AdminApiClient.ts
jkoenig134: I think it's totally ok in this context to say `deletionProcess`

packages/transport/src/modules/accounts/backbone/IdentityDeletionProcessClient.ts
tnotheis: The change is available with version 5.4.0

jkoenig134:
@sebbi08 you removed calls to `await this.accountController.syncDatawallet();` - why? You have to upload the cache change modifications.

packages/runtime/test/lib/AdminApiClient.ts
```diff
@@ -38,9 +39,11 @@ export async function cancelIdentityDeletionProcessFromBackboneAdminApi(
     transportService: TransportServices,
     accountAddress: string,
     identityDeletionProcessId: string
-): Promise<void> {
+): Promise<Result<IdentityDeletionProcessDTO, ApplicationError>> {
```
jkoenig134, line 42:
```suggestion
): Promise<Result<IdentityDeletionProcessDTO>> {
```

sebbi08:
> @sebbi08 you removed calls to `await this.accountController.syncDatawallet();` - why? You have to upload the cache change modifications.

We moved them from the use cases to the transport controller.
And we had to remove them in the external event handler because in the external event handler there is already a sync running and doing a data wallet sync there would wait indefinitely.

jkoenig134:
> We moved them from the use cases to the transport controller.

we don't do this anywhere else..

packages/runtime/test/lib/AdminApiClient.ts
```diff
@@ -45,5 +45,5 @@ export async function cancelIdentityDeletionProcessFromBackboneAdminApi(
 
     await syncUntilHasIdentityDeletionProcess(transportService, identityDeletionProcessId);
 
-    return await transportService.identityDeletionProcesses.getIdentityDeletionProcess({ id: identityDeletionProcessId });
+    return transportService.identityDeletionProcesses.getIdentityDeletionProcess({ id: identityDeletionProcessId });
```
jkoenig134, line 48: actually we want promises to fail where they were called and not where they are returned. Please re-add this await.

I don't know why our eslint rules aren't shouting.

packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -189,6 +189,27 @@ class Attributes {
     }
 }
 
+class Identity {
```
jkoenig134, line 192: should be `IdentityDeletionProcess`, I think this naming is from earlier

packages/runtime/src/useCases/transport/identityDeletionProcesses/CancelIdentityDeletionProcess.ts
```diff
@@ -0,0 +1,27 @@
...
+export class CancelIdentityDeletionProcessUseCase extends UseCase<void, IdentityDeletionProcessDTO> {
+    public constructor(
+        @Inject private readonly identityDeletionProcessController: IdentityDeletionProcessController,
+        @Inject private readonly accountController: AccountController
+    ) {
+        super();
+    }
+
+    protected async executeInternal(): Promise<Result<IdentityDeletionProcessDTO>> {
+        const activeIdentityDeletionProcess = await this.identityDeletionProcessController.getIdentityDeletionProcessByStatus(IdentityDeletionProcessStatus.Approved);
```
jkoenig134, line 17: see other comment, just remove the active

packages/transport/src/modules/accounts/IdentityDeletionProcessController.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    public async getIdentityDeletionProcesses(): Promise<IdentityDeletionProcess[]> {
+        return (await this.identityDeletionProcessCollection.find())
+            .map((identityDeletionProcess) => (identityDeletionProcess ? IdentityDeletionProcess.from(identityDeletionProcess) : undefined))
+            .filter((identityDeletionProcess) => !!identityDeletionProcess) as IdentityDeletionProcess[];
+    }
+
+    public async getIdentityDeletionProcessByStatus(
+        identityDeletionProcessStatus: IdentityDeletionProcessStatus | IdentityDeletionProcessStatus[]
+    ): Promise<IdentityDeletionProcess | undefined> {
```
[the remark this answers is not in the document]
jkoenig134: instead of `getIdentityDeletionProcessByStatus(["A", "B"])` you call it `getIdentityDeletionProcessByStatus("A", "B")` then

packages/transport/src/modules/accounts/data/IdentityDeletionProcess.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export interface IIdentityDeletionProcess {
+    id: CoreId;
+    cache?: CachedIdentityDeletionProcess;
+    cachedAt?: CoreDate;
+}
+
+@type("IdentityDeletionProcess")
+export class IdentityDeletionProcess extends CoreSynchronizable implements IIdentityDeletionProcess {
+    public override readonly technicalProperties = [nameof<IdentityDeletionProcess>((r) => r.id)];
```
jkoenig134, line 20: I think the ID is synced automatically.

packages/transport/src/modules/sync/externalEventProcessors/IdentityDeletionProcessStartedEventProcessor.ts
jkoenig134, line 1: see other processor, please move the actual logic to the controller

packages/transport/src/modules/accounts/data/IdentityDeletionProcess.ts
sebbi08: Yea, but without any synced property, the initial data wallet entry will not be created. And therefor no sync at all will happen.

packages/transport/src/modules/accounts/IdentityDeletionProcessController.ts
```diff
@@ -0,0 +1,114 @@
...
+            $or: identityDeletionProcessStatus.map((status) => {
+                return { "cache.status": status };
+            })
+        });
+        return identityDeletionProcess ? IdentityDeletionProcess.from(identityDeletionProcess) : undefined;
+    }
+
+    public async fetchCaches(ids: CoreId[]): Promise<{ id: CoreId; cache: CachedIdentityDeletionProcess }[]> {
+        if (ids.length === 0) return [];
+        const getIdentityDeletionPromises = ids.map((id) => this.identityDeletionProcessClient.getIdentityDeletionProcess(id.toString()));
```
britsta, line 104: "getIdentityDeletionProcesses" instead of "getIdentityDeletionPromises"

sebbi08: There is no await in the map, so at this point we still have the promises that resolve to the IdentityDeletionProcesses.

packages/transport/src/modules/accounts/data/IdentityDeletionProcess.ts
jkoenig134: It should create a cacheChange modification..?

packages/transport/src/modules/accounts/IdentityDeletionProcessController.ts
```diff
@@ -27,6 +27,23 @@ export class IdentityDeletionProcessController extends TransportController {
...
+    public async loadNewIdentityDeletionProcessFromBackbone(identityDeletionProcessId: string): Promise<IdentityDeletionProcess> {
+        const identityDeletionProcessJSONResponse = await this.identityDeletionProcessClient.getIdentityDeletionProcess(identityDeletionProcessId);
+
+        const newIdentityDeletionProcess = this.createIdentityDeletionProcessFromBackboneResponse(identityDeletionProcessJSONResponse);
+        await this.identityDeletionProcessCollection.create(newIdentityDeletionProcess);
+        this.eventBus.publish(new IdentityDeletionProcessStatusChangedEvent(this.parent.identity.address.toString(), newIdentityDeletionProcess));
+        return newIdentityDeletionProcess;
+    }
+
+    public async updateIdentityDeletionProcessById(identityDeletionProcess: string): Promise<IdentityDeletionProcess> {
```
jkoenig134, line 39: IMO this can just be called updateCache like in the other controllers that handle caching

packages/transport/src/modules/accounts/data/IdentityDeletionProcess.ts
jkoenig134: I see - it shouldn't. We aren't prepared for having no fields. But your workaround should be safe.

packages/app-runtime/test/lib/TestUtil.ts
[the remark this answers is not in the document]
jkoenig134: @sebbi08 this still seems to be unresolved.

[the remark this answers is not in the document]
sebbi08: yea missed the app runtime when reverting the changes

jkoenig134 (approved):
LGTM, there are some open discussions that should be resolved.

packages/runtime/test/transport/identityDeletionProcess.test.ts
sebbi08: we changed it to IdentityDeletionProcess everywhere it is better :)

jkoenig134 (approved):

packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -189,6 +189,23 @@ class Attributes {
...
+    public noActiveIdentityDeletionProcess() {
```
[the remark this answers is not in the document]
sebbi08: No active should mean that there is no process in the status `Approved` or `WaitingForApproval` this is only used in the `GetActiveIdentityDeletionProcessUseCase` now.

PR #122, opened 2024-05-07
Refactor/refactor validation for emailAdress and website

RuthDiG:
The emailAdress- Validation and the validation of the url had to be refactored and it should be possible to use German umlauts.


packages/content/test/attributes/EmailAddress.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        const invalidEMailAdressCall = () => {
+            EMailAddress.from({
+                value: "Hugo Becker@gmx.de"
+            });
+        };
+        expect(invalidEMailAdressCall).toThrow(
+            new ParsingError(
+                "EMailAddress",
+                "value",
+                "Value does not match regular expression /^[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:[.][a-z0-9!#$%&'*+/=?^_`{|}~-]+){0,}@(?:[a-z0-9ÄäÖöÜüß](?:[a-zÄäÖöÜüß0-9-]{0,61}[a-zÄäÖöÜüß0-9]){0,1}[.])+[a-zÄäÖöÜüß0-9](?:[a-zÄäÖöÜüß0-9-]{0,61}[a-zÄäÖöÜüß0-9])$/i"
```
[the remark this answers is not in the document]
Milena-Czierlinski: Library user. I suppose anyone seeing this error will be confused and we should try to make the error message more helpful.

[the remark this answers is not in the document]
jkoenig134: I mean this comes from serval, this would be a decision to change all serval classes using regexes.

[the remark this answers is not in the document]
Milena-Czierlinski: If these kind of errors are thrown everywhere we use regexes, then I vote for improving them in general. However, then this lies beyond the scope of this PR.

packages/content/test/attributes/Website.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        });
+        expect(validWebsite.value.toString()).toBe("https://inwind.it");
+    });
+    test("can create a RepositoryAttribute of Attribute Value Type Website with the German 'ä' in its domain", function () {
+        const validWebsite = Website.from({
+            value: "https://inwänd.it"
+        });
+        expect(validWebsite.value.toString()).toBe("https://inwänd.it");
+    });
+    test("can create a RepositoryAttribute of Attribute Value Type Website without www.", function () {
```
Milena-Czierlinski, line 17: You neither test websites starting with `www.` before. And in this test you omitted `https:`.

packages/content/test/attributes/Website.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        const validWebsite = Website.from({
+            value: "//google.it"
+        });
+        expect(validWebsite.value.toString()).toBe("//google.it");
+    });
+    test("can create a RepositoryAttribute of Attribute Value Type Website with a enhanced path", function () {
```
Milena-Czierlinski, line 23: a**n** enhanced

packages/content/test/attributes/Website.test.ts
```diff
@@ -0,0 +1,73 @@
...
+            value: "https://enmeshed.de/blog/meilenstein-enmeshed-als-komponente-ablage-in-mein-bildungsraum-geht-in-die-testphase-der-beta-version/"
+        });
+        expect(validWebsite.value.toString()).toBe(
+            "https://enmeshed.de/blog/meilenstein-enmeshed-als-komponente-ablage-in-mein-bildungsraum-geht-in-die-testphase-der-beta-version/"
+        );
+    });
+    test("returns an error when trying to create an Attribute Value Type Website with a blank in the value for Website", function () {
+        const invalidWebsiteCall = () => {
+            Website.from({
+                value: "Hugo https://inwind.it"
```
Milena-Czierlinski, line 34: Would and should this work if the blank wasn't in that place? `Hugohttps://inwind.it` I would have expected the blank somewhere within in the actual domain.

packages/content/test/attributes/Website.test.ts
RuthDiG: okay I will do this before.

packages/content/test/attributes/Website.test.ts
RuthDiG: It would never work if there would be a blank in any place. I was focused on the thing that would have been worked before and that was one of these things. But I can change the title of the test like: 
"returns an error when trying to create an Attribute Value Type Website which contains more than only the website"

Milena-Czierlinski: I actually preferred the name you had before, since it was more concrete.

Milena-Czierlinski (commented):
The tests look good to me, but I'm not really into the topic of regexes. Maybe somebody else can have a look, too.

packages/content/test/attributes/Website.test.ts
RuthDiG: I rechanged it now.

packages/content/test/attributes/EmailAddress.test.ts
```diff
@@ -0,0 +1,59 @@
+import { ParsingError } from "@js-soft/ts-serval";
+import { EMailAddress } from "../../src";
+
+describe("creation of RepositoryAttributes of Attribute Value Type EMailAddress", function () {
+    test("can create a RepositoryAttribute of Attribute Value Type EMailAddress", function () {
+        const validEMailAdress = EMailAddress.from({
```
britsta, line 6: `validEMailAddress` instead of `validEMailAdress`.

Same in other places below.

packages/content/test/attributes/EmailAddress.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        expect(validEMailAdress.value.toString()).toBe("peter123@inwind.it");
+    });
+    test("can create a RepositoryAttribute of Attribute Value Type EMailAddress with the German 'ä' in its domain", function () {
+        const validEMailAdress = EMailAddress.from({
+            value: "peter123@inwänd.it"
+        });
+        expect(validEMailAdress.value.toString()).toBe("peter123@inwänd.it");
+    });
+    test("returns an error when trying to create an Attribute Value Type EMailAddress with a blank in the value for EMailAddress", function () {
+        const invalidEMailAdressCall = () => {
```
britsta, line 18: `invalidEMailAddressCall` instead of `invalidEMailAdressCall`.

Same in other places below.

packages/content/test/attributes/EmailAddress.test.ts
```diff
@@ -0,0 +1,59 @@
...
+            EMailAddress.from({
+                value: "Becker@gmx"
+            });
+        };
+        expect(invalidEMailAdressCall).toThrow(
+            new ParsingError(
+                "EMailAddress",
+                "value",
+                "Value does not match regular expression /^[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:[.][a-z0-9!#$%&'*+/=?^_`{|}~-]+){0,}@(?:[a-z0-9ÄäÖöÜüß](?:[a-zÄäÖöÜüß0-9-]{0,61}[a-zÄäÖöÜüß0-9]){0,1}[.])+[a-zÄäÖöÜüß0-9](?:[a-zÄäÖöÜüß0-9-]{0,61}[a-zÄäÖöÜüß0-9])$/i"
```
britsta, line 41: Just a question out of interest: Why are the unicode terms, which originally appear in the regular expression (e.g. `\u00c4`), already replaced by the actual characters (e.g. `Ä`) in the tests for EMailAddress?

This seems not to be the case in the tests for the Website, where the unicode terms still occur in the regular expressions of the tests.

packages/content/test/attributes/EmailAddress.test.ts
```diff
@@ -0,0 +1,59 @@
+import { ParsingError } from "@js-soft/ts-serval";
+import { EMailAddress } from "../../src";
+
+describe("creation of RepositoryAttributes of Attribute Value Type EMailAddress", function () {
```
britsta, line 4: From now on we would like to establish the spelling "Attribute value type" instead of "Attribute Value Type". It would be great if you could adjust this everywhere in the test file for the EMailAddress and for the Website.

packages/content/test/attributes/Website.test.ts
```diff
@@ -0,0 +1,79 @@
...
+        };
+        expect(invalidWebsiteCall).toThrow(
+            new ParsingError(
+                "Website",
+                "value",
+                "Value does not match regular expression /^(?:|[A-Za-z]{3,9}:(?:\\/\\/)?|[:](?:\\/\\/)?|(?:\\/\\/))(?:(www[.])(?:[a-z0-9\\u00c4\\u00e4\\u00d6\\u00f6\\u00dc\\u00fc\\u00df](?:[A-Za-z\\u00c4\\u00e4\\u00d6\\u00f6\\u00dc\\u00fc\\u00df0-9-]{0,61}[A-Za-z\\u00c4\\u00e4\\u00d6\\u00f6\\u00dc\\u00fc\\u00df0-9]){0,1}[.])|((?!www.)(?:[a-z0-9\\u00c4\\u00e4\\u00d6\\u00f6\\u00dc\\u00fc\\u00df](?:[A-Za-z\\u00c4\\u00e4\\u00d6\\u00f6\\u00dc\\u00fc\\u00df0-9-]{0,61}[A-Za-z\\u00c4\\u00e4\\u00d6\\u00f6\\u00dc\\u00fc\\u00df0-9]){0,1}[.])))[A-Za-z\\u00c4\\u00e4\\u00d6\\u00f6\\u00dc\\u00fc\\u00df0-9](?:[A-Za-z\\u00c4\\u00e4\\u00d6\\u00f6\\u00dc\\u00fc\\u00df0-9-]{0,61}[A-Za-z\\u00c4\\u00e4\\u00d6\\u00f6\\u00dc\\u00fc\\u00df0-9])(\\:[0-9]+){0,}(?:[/][\\w\\u00c4\\u00e4\\u00d6\\u00f6\\u00dc\\u00fc\\u00df\\-_?#@!\\$&'\\(\\)\\*\\+,;=]{0,}){0,}$/"
+            )
+        );
+    });
+    test("returns an error when trying to create an Attribute Value Type Website with a - before the . in the value for Website", function () {
```
britsta, line 65: Make it clear that the crux is here that the `-` is **directly** before the `.`

packages/content/test/attributes/EmailAddress.test.ts
RuthDiG: In the RegExpConstructor you can insert a string than you have to use "" or a regexp there you use // and that's why in one case it translates the expression and in the other not, but you are right it should be traded in the same way and so I will change this.

packages/content/test/attributes/Website.test.ts
```diff
@@ -0,0 +1,93 @@
...
+            value: "https://enmeshed.de/blog/meilenstein-enmeshed-als-komponente-ablage-in-mein-bildungsraum-geht-in-die-testphase-der-beta-version/"
+        });
+        expect(validWebsite.value.toString()).toBe(
+            "https://enmeshed.de/blog/meilenstein-enmeshed-als-komponente-ablage-in-mein-bildungsraum-geht-in-die-testphase-der-beta-version/"
+        );
+    });
+    test("returns an error when trying to create an Attribute value type Website with a blank in the value for Website", function () {
+        const invalidWebsiteCall = () => {
+            Website.from({
+                value: "https://inwind.test it"
```
sebbi08, line 40: i think we are missing a test here with an encoded url like `https://inwind.test/%20it/`

RuthDiG: Good point, I changed this.

packages/content/test/attributes/EmailAddress.test.ts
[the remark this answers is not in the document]
RuthDiG: https://js-soft.atlassian.net/browse/NMSHDR-100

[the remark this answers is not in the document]
britsta: (Sidenote: As far as I know, we should avoid linking to our internal boards or customer boards on our public repositories.)

packages/content/test/attributes/EmailAddress.test.ts
britsta, line 1: Rename the file: "EMailAddress.test.ts" instead of "EmailAddress.test.ts".

packages/content/src/attributes/types/strings/AbstractURL.ts
```diff
@@ -8,8 +8,7 @@ export abstract class AbstractURL extends AbstractString {
         min: 3,
         max: 1024,
         regExp: new RegExp(
-            // eslint-disable-next-line no-useless-escape
-            /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=\+\$,\w]+@)?[A-Za-z0-9.-]+|(?:www.|[-;:&=\+\$,\w]+@)[A-Za-z0-9.-]+)((?:\/[\+~%\/.\w-_]*)?\??(?:[-\+=&;%@.\w_]*)#?(?:[\w]*))?)/
+            /^(?:|[A-Za-z]{3,9}:(?:[/][/])?|[:](?:[/][/])?|(?:[/][/]))(?:(www[.])(?:[A-Za-z0-9\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df](?:[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9-]{0,61}[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9]){0,1}[.])|((?!www[.])(?:[A-Za-z0-9\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df](?:[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9-]{0,61}[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9]){0,1}[.])))[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9](?:[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9-]{0,61}[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9])(?:[:][0-9]+){0,}(?:[/][A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9?#@!$&'()*+/,;=%-]{0,}){0,}$/
         )
     })
     public override value: string;
```
slavistan, line 14: 1. You can replace unicode escape sequences (`\u00c4`, ...) with literal characters (`Ä`, ...) in regular expressions. This will make this madness more legible.

    ```ts
    re = /[A-Za-z0-9ÄäÖöÜüß]/
    ```

2. You're using non-capturing groups `(?:...)`, which would be useful if you actually used capture groups somewhere to _substitute_ certain parts of the string matched by the regex. But since we're just matching and not substituting here just yeet the `?:` prefix and make everything a little more legible again.

3. _www_ has no special role in a domain. It's an old convention to use the _www_ subdomain to host one's public facing website (e.g. _www.ebay.com_). It's practically meaningless nowadays, so no need to give it any separate treatment.

Also, I think you're checking that there's no more than one _www_ subdomain in the url. This is perfectly legal, however. `www.foo.www.www.enmeshed.eu` is a perfectly fine domain. 

4. `{0,1}` is better expressed as `?` in a regex, e.g. `([a-z]{0,61})?` instead of `([a-z]{0,61}){0,1}`.

5. I assume `(\:[0-9]+)` is supposed to check for the port of a host. What's the leading backslash good for? I don't think you need it.

6. `{0,}` is better expressed as `*` in a regex, e.g. `([a-z]{0,61})*` instead of `([a-z]{0,61}){0,}`.

7. Is `[.]` supposed to match a literal period character '.'? Use the standard `\.` (a `.` inside a regex means "anything but a newline", hence you have to escape it).

8. Can you help me make sense of those segments of up to 61 characters? I don't quite understand what these are about.

packages/content/src/attributes/types/strings/AbstractURL.ts
```diff
@@ -19,7 +18,7 @@ export abstract class AbstractURL extends AbstractString {
             min: 3,
             max: 1024,
             pattern:
-                "/((([A-Za-z]{3,9}:(?:\\/\\/)?)(?:[-;:&=\\+\\$,\\w]+@)?[A-Za-z0-9.-]+|(?:www.|[-;:&=\\+\\$,\\w]+@)[A-Za-z0-9.-]+)((?:\\/[\\+~%\\/.\\w\\-_]*)?\\??(?:[-\\+=&;%@.\\w_]*)#?(?:[\\w]*))?)/"
+                "/^(?:|[A-Za-z]{3,9}:(?:[/][/])?|[:](?:[/][/])?|(?:[/][/]))(?:(www[.])(?:[A-Za-z0-9\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df](?:[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9-]{0,61}[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9]){0,1}[.])|((?!www[.])(?:[A-Za-z0-9\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df](?:[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9-]{0,61}[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9]){0,1}[.])))[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9](?:[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9-]{0,61}[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9])(?:[:][0-9]+){0,}(?:[/][A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9?#@!$&'()*+/,;=%-]{0,}){0,}$/"
         });
     }
```
slavistan, line 23: If you need a string version of your regex, just call `String(myRegexp)` instead of the hard-coded copy.

packages/content/src/attributes/types/strings/AbstractURL.ts
```diff
@@ -8,8 +8,7 @@
         min: 3,
         max: 1024,
         regExp: new RegExp(
-            // eslint-disable-next-line no-useless-escape
-            /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[-;:&=\+\$,\w]+@)?[A-Za-z0-9.-]+|(?:www.|[-;:&=\+\$,\w]+@)[A-Za-z0-9.-]+)((?:\/[\+~%\/.\w-_]*)?\??(?:[-\+=&;%@.\w_]*)#?(?:[\w]*))?)/
+            /^(?:|[A-Za-z]{3,9}:(?:[/][/])?|[:](?:[/][/])?|(?:[/][/]))(?:(www[.])(?:[A-Za-z0-9\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df](?:[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9-]{0,61}[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9]){0,1}[.])|((?!www[.])(?:[A-Za-z0-9\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df](?:[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9-]{0,61}[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9]){0,1}[.])))[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9](?:[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9-]{0,61}[A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9])(?:[:][0-9]+){0,}(?:[/][A-Za-z\u00c4\u00e4\u00d6\u00f6\u00dc\u00fc\u00df0-9?#@!$&'()*+/,;=%-]{0,}){0,}$/
```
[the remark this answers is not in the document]
slavistan: Your or-expression `(|[A-Za-z]{3,9}:(?:\/\/)?|[:](?:\/\/)?|(?:\/\/))` allows to match an empty string due to the leading `|`. The idiomatic way to express this is to use make the whole expression optional using a trailing `?`: `([A-Za-z]{3,9}:(?:\/\/)?|[:](?:\/\/)?|(?:\/\/))?`

packages/content/test/attributes/Website.test.ts
```diff
@@ -0,0 +1,105 @@
...
+            value: "https://inwänd.it"
+        });
+        expect(validWebsite.value.toString()).toBe("https://inwänd.it");
+    });
+    test("can create a RepositoryAttribute of Attribute value type Website without www.", function () {
+        const validWebsite = Website.from({
+            value: "//google.it"
+        });
+        expect(validWebsite.value.toString()).toBe("//google.it");
+    });
```
slavistan, line 28: That's not a valid website URL, is it?

packages/content/test/attributes/Website.test.ts
```diff
@@ -0,0 +1,105 @@
+import { ParsingError } from "@js-soft/ts-serval";
+import { Website } from "../../src";
+
+describe("creation of RepositoryAttributes of Attribute value type Website", function () {
```
slavistan, line 4: Most of this file consists of repetitive boilerplate, very low information density here.

Here's a suggestion: Create two separate test suites (`describe` blocks) in this file, one for valid and one for invalid websites. In each you define an array of URLs, which you test with a [parameterized jest test](https://medium.com/@Klaymant/parameterized-tests-with-typescript-jest-19f45a4a2925).

```ts
decribe("Test valid URLs", () => {
    const validUrls = [
        "www.google.com",
        "https://döner.cooking",
        // add all the weird URLs you want to test here.        
    ]
    
    test.each(validUrls)("url %s is recognized as valid", (url) => {
        const validWebsite = Website.from({value: url});
        expect(validWebsite.value.toString())toBe(url)
    })
})

// Same for invalid URLs
// ...
```

`test.each([...])` is basically equivalent to a for-loop running many tests with the added benefit that all the test results appear in the output individually, as if you had written them manually.

packages/content/src/attributes/types/strings/AbstractURL.ts
RuthDiG: 1. & 2. Thank you for the hints!
3. That wasn't my intension, I didn't want that you can write www.google for example, but for other reasons www.foo.www.www.enmeshed.eu didn't work because I overlooked it, but now it works when I will push the commit. 
4.&6. I know but I thought that someone who don't know what + and * means would understand it better but anyway it's shorter so I change it. 
5.&7. I used \: for escaping it but when you put it in []-brackets you don't need to escape and therefore in 7. I also put the . into [] 
By the way If you don't avoid that typescript deletes unnecessary escape characters you will have the problem that were there in further versions. 
8. Between the points are the domains and every domain can't have more than 63 characters and can't end or begin with a -

packages/content/src/attributes/types/strings/AbstractEMailAddress.ts
```diff
@@ -4,19 +4,22 @@ import { AbstractString } from "../AbstractString";
...
     })
     public override value: string;
 
     public static override get valueHints(): ValueHints {
         return super.valueHints.copyWith({
             min: 3,
-            max: 100,
-            pattern: "/^[A-Z0-9._%+-]+@[A-Z0-9.-]+.[A-Z]{2,}$/i"
+            max: 254,
+            pattern: JSON.stringify(AbstractEMailAddress.regExp, null, 2)
```
slavistan, line 22: Use `String(...)` here instead of `JSON.stringify` (as it was before we thought about using multiple regexes).

britsta (dismissed):

packages/content/src/attributes/types/strings/AbstractEMailAddress.ts
```diff
@@ -4,19 +4,22 @@ import { AbstractString } from "../AbstractString";
 
 export abstract class AbstractEMailAddress extends AbstractString {
     // from https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address
+    private static readonly regExp = new RegExp(
+        /^[A-Za-z0-9!#$%&'*+/=?^_`{|}~-]+(\.[A-Za-z0-9!#$%&'*+/=?^_`{|}~-]+)*@([A-Za-z0-9ÄäÖöÜüß]([A-Za-z0-9ÄäÖöÜüß-]{0,61}[A-Za-z0-9ÄäÖöÜüß])?\.)+[A-Za-z0-9ÄäÖöÜüß][A-Za-z0-9ÄäÖöÜüß-]{0,61}[A-Za-z0-9ÄäÖöÜüß]$/
```
slavistan, line 8: The part left of the '@' is forbids non-ascii characters. Add them like you did in the domain part of the email address. Don't forget to add a test.

packages/content/test/attributes/Website.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    test.each(validUrls)("URL %s is recognized as valid", (url) => {
+        const validWebsite = Website.from({ value: url });
+        expect(validWebsite.value.toString()).toBe(url);
+    });
+});
+
+describe("Test invalid URLs", () => {
+    const invalidUrls = ["google-.de", "www.google", "www.-google", "https://inwind.test it"];
+
```
slavistan, line 22: Add a test for an empty URL.

packages/content/src/attributes/types/strings/AbstractURL.ts
```diff
@@ -3,23 +3,23 @@ import { RenderHints, RenderHintsDataType, RenderHintsEditType, ValueHints } fro
 import { AbstractString } from "../AbstractString";
 
 export abstract class AbstractURL extends AbstractString {
+    private static readonly regExp = new RegExp(
+        /^([A-Za-z]+:\/\/)?(www\.([A-Za-z0-9ÄäÖöÜüß]([A-Za-zÄäÖöÜüß0-9-]{0,61}[A-Za-zÄäÖöÜüß0-9])?\.)+|((?!www[.])([A-Za-z0-9ÄäÖöÜüß]([A-Za-zÄäÖöÜüß0-9-]{0,61}[A-Za-zÄäÖöÜüß0-9])?[.]))([A-Za-z0-9ÄäÖöÜüß]([A-Za-zÄäÖöÜüß0-9-]{0,61}[A-Za-zÄäÖöÜüß0-9])?\.)*)[A-Za-zÄäÖöÜüß0-9]([A-Za-zÄäÖöÜüß0-9-]{0,61}[A-Za-zÄäÖöÜüß0-9])([:][0-9]+)*(\/[A-Za-zÄäÖöÜüß0-9?#@!$&'()*+,;=%-]*)*$/
```
slavistan, line 7: Here's an abridged version of your regex:

```js
/^([A-Za-z]+:\/\/)?((www\.)|(?!www\.))([A-Za-z0-9ÄäÖöÜüß]([A-Za-zÄäÖöÜüß0-9-]{0,61}[A-Za-zÄäÖöÜüß0-9])?\.)+([A-Za-z0-9ÄäÖöÜüß]([A-Za-zÄäÖöÜüß0-9-]{0,61}[A-Za-zÄäÖöÜüß0-9])?)(:[0-9]+)?(\/[A-Za-zÄäÖöÜüß0-9?#@!$&'()*+,;=%-]*)*$/
```

packages/content/test/attributes/Website.test.ts
RuthDiG: Should we also accept an empty emailaddress?

packages/content/src/attributes/types/strings/AbstractEMailAddress.ts
RuthDiG: It was said that we should allow non-ascii characters in the domains (you can see it in the ticket) that's why I didn't add them there. So should I?

packages/content/test/attributes/Website.test.ts
slavistan: I meant to _reject_ empty strings :-) Same for email addresses, too.

packages/content/test/attributes/EMailAddress.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    const validEMailAddresses = ["peter123@inwind.it", "peter123@inwänd.it"];
+
+    test.each(validEMailAddresses)("EMail %s is recognized as valid", (email) => {
+        const validEMailAddress = EMailAddress.from({ value: email });
+        expect(validEMailAddress.value.toString()).toBe(email);
+    });
+});
+
+describe("Test invalid EMailAddresses", () => {
+    const invalidEMailAddresses = ["Hugo Becker@gmx.de", "Becker@gmx", "Becker@gmx-.de", "Becker@gmx-.de", ".Becker@gmx.de", "test@.address", "test@test..address"];
```
slavistan, line 14: Test that empty email addresses are rejected.

packages/content/test/attributes/Website.test.ts
```diff
@@ -0,0 +1,45 @@
+import { ParsingError } from "@js-soft/ts-serval";
+import { Website } from "../../src";
+
+describe("Test valid URLs", () => {
+    const validUrls = [
```
slavistan, line 5: Test that empty strings are rejected.

packages/content/test/attributes/Website.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    ];
+
+    test.each(validUrls)("URL %s is recognized as valid", (url) => {
+        const validWebsite = Website.from({ value: url });
+        expect(validWebsite.value.toString()).toBe(url);
+    });
+});
+
+describe("Test invalid URLs", () => {
+    const invalidUrls = ["google-.de", "www.google", "www.-google", "https://inwind.test it"];
```
slavistan, line 21: Test that empty strings are rejected.

slavistan: Oh, my bad. It's further down below. All good 👍

slavistan (approved):

PR #123, opened 2024-05-07
Ci/do not include bundled libs

jkoenig134:
As we have no more app live that requires bundling we can get rid of huge packages and longer build times.


jkoenig134:
For context: we use the fabulous [esbuild](https://esbuild.github.io/) to bundle for our flutter app.

tnotheis (approved):

PR #125, opened 2024-05-10
Feature/Relationship decomposition

Magnus-Kuhn:


Magnus-Kuhn:
To not wait for the backbone to complete the pseudonymization any longer, I'm asking for a review now  since this only affects the messages. The tests for the messages are skipped.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -322,6 +322,11 @@ export class AttributesController extends ConsumptionBaseController {
         this.eventBus.publish(new AttributeDeletedEvent(this.identity.address.toString(), attribute));
     }
 
+    public async deleteAttributesSharedWithPeer(peer: ICoreAddress): Promise<void> {
```
slavistan, line 325: `peer` should be a `CoreAddress`, otherwise `.toString()` won't work unless peer is actually of type `CoreAddress`.

```ts
{ address: "id1234" }.toString()
// [object Object]
```

If you want to keep the interface type for `peer` use `peer.address` to get the actual address as a string.

jkoenig134:
Why do you scattered the name `BackboneRelationshipStatus` all over the codebase? I think the wording is not good.

jkoenig134:
After decomposing on the backbone you have to handle messages yourself, the backbone will not send external events.

packages/runtime/src/types/transport/RecipientDTO.ts
```diff
@@ -2,5 +2,5 @@ export interface RecipientDTO {
     address: string;
     receivedAt?: string;
     receivedByDevice?: string;
-    relationshipId: string;
+    relationshipId?: string;
```
slavistan, line 5: Why is this optional now? Would you mind giving me a little context?

jkoenig134: I think this is caused by the anonymized messages. We don't have relationships for such recipients.

slavistan: Ah, right. Cheers 👍

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -62,6 +62,25 @@ export class MessageController extends TransportController {
...
+        if (message.relationshipIds.length === 0) {
+            await this.messages.delete(message);
+            return;
+        }
+        await this.updateCacheOfMessage(message);
+
+        await this.messages.update(messageDoc, message);
+    }
+
+    public async decomposeMessagesOfRelationship(relationship: Relationship): Promise<void> {
```
slavistan, line 79: We _decompose_ a relationship but not messages. Why not something like `deletePeerOfRelationshipFrom(ReceipientsOf)Messages`?

packages/runtime/src/useCases/transport/messages/MessageMapper.ts
```diff
@@ -54,16 +82,49 @@ export class MessageMapper {
...
-            relationshipId: relationshipId.toString()
+            relationshipId: relationshipId?.toString()
         };
     }
+
+    private static isPseudonymizationRequired(recipients: MessageEnvelopeRecipient[], relationshipIds: CoreId[]): boolean {
+        return recipients.length === relationshipIds.length ? false : true;
+    }
+
+    private static async getPseudonymizedMessageContent(messageContent: any, recipients: MessageEnvelopeRecipient[], pseudonym: string): Promise<any> {
```
slavistan, line 113: Unnecessarily async

jkoenig134:
> I didn't like the backbone status DeletionProposed as runtime status name, so I changed it there into DecomposedByPeer

I don't get this explanation. We shouldn't just change how things are named.

jkoenig134:
Please move the refactoring (rename from RelationshipStatus to BackboneRelationship) out of this PR. It is to much noise.

packages/runtime/src/useCases/transport/messages/MessageMapper.ts
```diff
@@ -18,11 +18,25 @@ export class MessageMapper {
         };
     }
 
-    public static toMessageWithAttachmentsDTO(message: Message, attachments: File[]): MessageWithAttachmentsDTO {
+    public static async toMessageWithAttachmentsDTO(message: Message, attachments: File[]): Promise<MessageWithAttachmentsDTO> {
```
jkoenig134, line 21: I have strong doubts that this really should be async. Can we achieve this differently?

packages/runtime/src/useCases/transport/relationships/DecomposeRelationship.ts
```diff
@@ -0,0 +1,56 @@
...
+        const peer = relationship.peer;
+
+        // backbone call first so nothing is deleted in case it goes wrong
+        await this.relationshipsController.decompose(relationship.id);
+        await this.relationshipTemplateController.cleanupDuringRelationshipDecomposition(relationship.cache.template);
+        await this.messageController.deleteRelationshipFromMessages(relationship);
+        await this.incomingRequestsController.deleteRequestsFromPeer(peer.address);
+        await this.outgoingRequestsController.deleteRequestsToPeer(peer.address);
+        await this.notificationsController.deleteNotificationsWithPeer(peer.address);
+        await this.attributesController.deleteAttributesSharedWithPeer(peer.address);
```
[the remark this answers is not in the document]
Magnus-Kuhn: why is this an improvement?

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -293,6 +293,22 @@ export class RelationshipsController extends TransportController {
...
+        const result = await this.client.decomposeRelationship(relationshipId.toString());
+        if (result.isError) throw result.error;
+
+        const isSecretDeletionSuccessful = await this.secrets.deleteSecretForRelationship(relationship.relationshipSecretId);
+        if (!isSecretDeletionSuccessful) {
+            throw new TransportError("Decomposition failed to delete secrets");
+        }
+        await this.relationships.delete({ id: relationshipId });
+
+        this.eventBus.publish(new RelationshipDecomposedBySelfEvent(this.parent.identity.address.toString(), relationshipId));
```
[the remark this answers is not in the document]
Magnus-Kuhn: agreed, how is this done?

packages/runtime/src/useCases/transport/relationships/DecomposeRelationship.ts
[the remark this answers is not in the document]
jkoenig134: It moves knowledge from transport / consumption to .... transport and consumption.

packages/transport/src/modules/relationships/RelationshipsController.ts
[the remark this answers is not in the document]
jkoenig134: just log and not throw in ln 305

packages/runtime/src/useCases/transport/messages/MessageMapper.ts
jkoenig134: As said - the pseudonymization is NOT the task of the MessageMapper. It should already be done before in the DB.

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -62,6 +62,25 @@ export class MessageController extends TransportController {
         });
     }
 
+    public async deleteRelationshipFromMessage(messageId: CoreId, relationship: Relationship): Promise<void> {
```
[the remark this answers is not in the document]
jkoenig134: Why does the cache update not update the relationship ids? that would be nice.

jkoenig134:
@stnmtz as this now heavily changes the behaviour of messages I'd be happy about a review.

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -71,12 +72,30 @@ export class MessageController extends TransportController {
         const messageDoc = await this.messages.read(messageId.toString());
         const message = Message.from(messageDoc);
 
-        message.relationshipIds = message.relationshipIds.filter((id) => !id.equals(relationship.id));
-        if (message.relationshipIds.length === 0) {
+        // a received message only has one recipient (yourself) so it can be deleted without further looking into the recipients
+        if (!message.isOwn) {
```
jkoenig134, line 76: also delete here if there is only one recipient

packages/runtime/src/useCases/transport/relationships/DecomposeRelationship.ts
```diff
@@ -0,0 +1,44 @@
...
+        const relationship = await this.relationshipsController.getRelationship(CoreId.from(request.relationshipId));
+        if (!relationship) {
+            return Result.fail(RuntimeErrors.general.recordNotFound(Relationship));
+        }
+
+        if (!relationship.cache) {
+            return Result.fail(RuntimeErrors.general.cacheEmpty(Relationship, relationship.id.toString()));
+        }
+
+        await this.accountController.decomposeRelationshipAndCleanupData(relationship as Relationship & { cache: CachedRelationship });
```
jkoenig134, line 37: should be two methods

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -246,4 +248,10 @@ export class RelationshipTemplateController extends TransportController {
 
         return relationshipTemplate;
     }
+
+    public async cleanupTemplateOfDecomposedRelationship(template: RelationshipTemplate): Promise<void> {
+        if (!template.isOwn || template.cache?.maxNumberOfAllocations === 1) {
+            await this.templates.delete(template);
+        }
```
jkoenig134, line 255: maybe we should remove *all* templates of the peer - I don't care if this happens in another method or in this one (which should then be renamed)

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -72,7 +72,6 @@ export class MessageController extends TransportController {
         const messageDoc = await this.messages.read(messageId.toString());
         const message = Message.from(messageDoc);
 
-        // a received message only has one recipient (yourself) so it can be deleted without further looking into the recipients
```
jkoenig134: I think this got wrong while merging.

packages/runtime/src/useCases/transport/relationships/DecomposeRelationship.ts
```diff
@@ -34,7 +34,9 @@ export class DecomposeRelationshipUseCase extends UseCase<DecomposeRelationshipR
             return Result.fail(RuntimeErrors.general.cacheEmpty(Relationship, relationship.id.toString()));
         }
 
-        await this.accountController.decomposeRelationshipAndCleanupData(relationship as Relationship & { cache: CachedRelationship });
+        // backbone call first so nothing is deleted in case it goes wrong
+        await this.relationshipsController.decompose(relationship.id);
+        await this.accountController.cleanupDataOfDecomposedRelationship(relationship as Relationship & { cache: CachedRelationship });
```
jkoenig134, line 39: add newline between these calls

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
jkoenig134: why? We are decomposing *here* to get rid of all data.

packages/transport/src/modules/messages/MessageController.ts
jkoenig134: (we need that comment)

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
Magnus-Kuhn: just from the perspective that you delete something once it's useless - it depends on whether you want to create a relationship with an old template or if the history is of any use

jkoenig134: It's not only about sth to be useless, it's about that that Templates are containing sensitive data.

jkoenig134 (approved):

packages/consumption/src/modules/settings/SettingsController.ts
```diff
@@ -56,4 +56,11 @@ export class SettingsController extends ConsumptionBaseController {
     public async deleteSetting(setting: Setting): Promise<void> {
         await this.settings.delete(setting);
     }
+
+    public async deleteSettingsForRelationship(relationshipId: CoreId): Promise<void> {
+        const settings = await this.getSettings({ reference: relationshipId.toString(), scope: SettingScope.Relationship });
+        for (const setting of settings) {
+            await this.deleteSetting(setting);
```
Magnus-Kuhn, line 63: why this instead of Promise.all ? (similarly for relationship templates and tokens)

jkoenig134: don't want to parallelize db calls espacially if the datawallet is used

jkoenig134 (approved):

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -324,7 +324,7 @@ export class AttributesController extends ConsumptionBaseController {
 
     public async deleteAttributesExchangedWithPeer(peer: CoreAddress): Promise<void> {
         const attributes = await this.getLocalAttributes({ "shareInfo.peer": peer.toString() });
-        await Promise.all(attributes.map((attribute) => this.deleteAttributeUnsafe(attribute.id)));
+        attributes.forEach(async (attribute) => await this.deleteAttributeUnsafe(attribute.id));
```
jkoenig134, line 327: @Magnus-Kuhn please don't use .forEach for async operations

jkoenig134: these operations will just run in parallel and run into an unhandled promise rejection on an error

jkoenig134 (approved):

packages/transport/src/modules/files/FileController.ts
```diff
@@ -48,12 +48,18 @@ export class FileController extends TransportController {
...
+                return;
+            }
+
             const file = File.from(fileDoc);
 
             return { id: CoreId.from(f.id), cache: await this.decryptFile(f, file.secretKey) };
         });
 
-        return await Promise.all(decryptionPromises);
+        const chaches = await Promise.all(decryptionPromises);
```
stnmtz, line 61: caches ;)

packages/transport/src/modules/files/FileController.ts
```diff
@@ -48,12 +48,18 @@ export class FileController extends TransportController {
 
         const decryptionPromises = backboneFiles.map(async (f) => {
             const fileDoc = await this.files.read(f.id);
+            if (!fileDoc) {
+                this._log.error(`File '${f.id}' not found in local database and the cache can therefore not be applied.`);
```
stnmtz, line 52: Was heißt denn hier "cache can therefore not be applied"?

packages/transport/src/modules/files/FileController.ts
```diff
@@ -264,4 +270,12 @@ export class FileController extends TransportController {
 
         return decrypt;
     }
+
+    public async cleanupFilesOfDecomposedRelationship(peer: CoreAddress): Promise<void> {
+        const tokenDocs = await this.getFiles({ "cache.createdBy": peer.toString() });
```
stnmtz, line 275: What happens with files send to this Relationship? Or with files received from this relationship which were not created by this relationship? (I don't know if the latter is actually possible?)

stnmtz: - Naming: filesDocs?
- We should not clean up files so far, there still are some open questions, as we do not fetch all files with this query and might delete too many (as they are not linked to one relationship)

jkoenig134 (approved):

stnmtz (approved):

PR #126, opened 2024-05-13
Fix/Add identityDeletionProcesses to ExternalEventReceivedEvent

Milena-Czierlinski:
I used `SyncEverythingResponse` instead of manually adding `identityDeletionProcesses` in order to improve type safety for the future.

### DIFF
--- a/packages/app-runtime/src/events/ExternalEventReceivedEvent.ts
+++ b/packages/app-runtime/src/events/ExternalEventReceivedEvent.ts
@@ -1,15 +1,9 @@
-import { DataEvent, MessageDTO, RelationshipDTO } from "@nmshd/runtime";
+import { DataEvent, SyncEverythingResponse } from "@nmshd/runtime";
 
-export class ExternalEventReceivedEvent extends DataEvent<{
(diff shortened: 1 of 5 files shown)


packages/app-runtime/src/events/ExternalEventReceivedEvent.ts
jkoenig134, line 1: I'm not a big fan of re-using the SyncEverythingResponse type here..

Milena-Czierlinski: I don't think it's super elegant, either. But writing its components manually, leads to what just happened -  the adjustment here slipped through in our initial PR, since we didn't get support from the compiler. Do you have another suggestion, that stills ensures us this support? @sebbi08 proposed defining a new type that simply extends SyncEverythingResponse.

Milena-Czierlinski: Unfortunately, this doesn't yield the endeavored type safety.

jkoenig134 (approved):

PR #128, opened 2024-05-13
Ci/XSRF token being retrieved without the API key makes it invalid for subsequent calls

Dannyps:
As the subsequent calls carry the API key, this results in a claims-based user mismatch.

This PR fixes the current behaviour.

Btw, I'm not sure why there are two files, but I fixed both anyway.


Dannyps:
It seems that the pipelines won't run...
See #82 and #84

sebbi08 (dismissed):

tnotheis (dismissed):

sebbi08 (dismissed):

PR #130, opened 2024-05-14
Test/IdentityDeletionProcess events

Milena-Czierlinski:


packages/runtime/test/transport/identityDeletionProcess.test.ts
```diff
@@ -43,16 +50,37 @@ afterEach(async () => {
...
+        await expect(eventBus).toHavePublished(IdentityDeletionProcessStatusChangedEvent, (e) => e.data.status === IdentityDeletionProcessStatus.WaitingForApproval);
+
+        await transportService.identityDeletionProcesses.approveIdentityDeletionProcess();
+
+        const result = await cancelIdentityDeletionProcessFromBackboneAdminApi(transportService, accountAddress, initiatedIdentityDeletionProcess.id.toString());
+        expect(result).toBeSuccessful();
+        const identityDeletionProcess = result.value;
+        expect(identityDeletionProcess.status).toStrictEqual(IdentityDeletionProcessStatus.Cancelled);
+
+        await expect(eventBus).toHavePublished(IdentityDeletionProcessStatusChangedEvent, (e) => e.data.id === initiatedIdentityDeletionProcess.id);
```
sebbi08, line 67: You either need to reset the event bus in between create/approve and cancel or move the check for the status in the same callback, as the toHavPublished will check all the event since the last reset.

And as the event name and the IDP id are the same this check will always "succeed" with the event from the creation

sebbi08 (changes requested):

jkoenig134 (approved):

PR #131, opened 2024-05-15
Feature/Add toggle to disable automatic identiy creation

sebbi08:


packages/app-runtime/src/AppConfig.ts
```diff
@@ -24,7 +24,8 @@ export function createAppConfig(...configs: AppConfigOverwrite[]): AppConfig {
         accountsDbName: "accounts",
         transportLibrary: {
             realm: Realm.Prod,
-            datawalletEnabled: true
+            datawalletEnabled: true,
+            allowIdentityCreation: true
```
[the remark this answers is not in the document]
jkoenig134: yep, I think so.

packages/transport/src/core/Transport.ts
```diff
@@ -56,6 +58,7 @@ export class Transport {
     }
 
     private static readonly defaultConfig: IConfig = {
+        allowIdentityCreation: false,
```
jkoenig134, line 61: I would vote for leaving the functionality as-is by using true as a default and using false in the connector

packages/transport/src/modules/accounts/AccountController.ts
jkoenig134, line 1: I'm a bit confused that your vscode sorts imports completely different than mine. Is it up2date?

sebbi08: Yea also took a look into it, do you have some extension installed that alters the sorting ? 

Maybe we should add [this](https://eslint.org/docs/latest/rules/sort-imports) to our eslint config

jkoenig134: No, I use vscode default behaviour.

sebbi08: Hm weird, when I resave the file the imports go back to the right order. Like I had not saved the file before ^^

jkoenig134:
don't you want to publish all this?

jkoenig134:
version bump not added to lockfile.... `npm i` please

jkoenig134 (approved):

PR #133, opened 2024-05-21
Feature/relationship reactivation

Magnus-Kuhn:


packages/transport/src/modules/relationships/backbone/RelationshipClient.ts
```diff
@@ -26,6 +26,25 @@ export class RelationshipClient extends RESTClientAuthenticate {
...
+    public async rejectRelationshipActivation(relationshipId: string): Promise<ClientResult<BackbonePutRelationshipsResponse>> {
+        return await this.put<BackbonePutRelationshipsResponse>(`/api/v1/Relationships/${relationshipId}/Reactivate/Reject`, {});
+    }
+
+    public async revokeRelationshipActivation(relationshipId: string): Promise<ClientResult<BackbonePutRelationshipsResponse>> {
+        return await this.put<BackbonePutRelationshipsResponse>(`/api/v1/Relationships/${relationshipId}/Reactivate/Revoke`, {});
+    }
+    public async decomposeRelationship(relationshipId: string): Promise<ClientResult<void>> {
+        return await this.put<void>(`/api/v1/Relationships/${relationshipId}/Decompose`, {});
+    }
```
jkoenig134, line 46: this is a PR about reactivation, why is decomposition included?

packages/transport/src/modules/relationships/backbone/RelationshipClient.ts
```diff
@@ -26,6 +26,25 @@ export class RelationshipClient extends RESTClientAuthenticate {
...
+        return await this.put<BackbonePutRelationshipsResponse>(`/api/v1/Relationships/${relationshipId}/Reactivate/Accept`, {});
+    }
+
+    public async rejectRelationshipActivation(relationshipId: string): Promise<ClientResult<BackbonePutRelationshipsResponse>> {
+        return await this.put<BackbonePutRelationshipsResponse>(`/api/v1/Relationships/${relationshipId}/Reactivate/Reject`, {});
+    }
+
+    public async revokeRelationshipActivation(relationshipId: string): Promise<ClientResult<BackbonePutRelationshipsResponse>> {
+        return await this.put<BackbonePutRelationshipsResponse>(`/api/v1/Relationships/${relationshipId}/Reactivate/Revoke`, {});
+    }
```
jkoenig134, line 43:
```suggestion
    public async acceptRelationshipReactivation(relationshipId: string): Promise<ClientResult<BackbonePutRelationshipsResponse>> {
        return await this.put<BackbonePutRelationshipsResponse>(`/api/v1/Relationships/${relationshipId}/Reactivate/Accept`, {});
    }

    public async rejectRelationshipReactivation(relationshipId: string): Promise<ClientResult<BackbonePutRelationshipsResponse>> {
        return await this.put<BackbonePutRelationshipsResponse>(`/api/v1/Relationships/${relationshipId}/Reactivate/Reject`, {});
    }

    public async revokeRelationshipReactivation(relationshipId: string): Promise<ClientResult<BackbonePutRelationshipsResponse>> {
        return await this.put<BackbonePutRelationshipsResponse>(`/api/v1/Relationships/${relationshipId}/Reactivate/Revoke`, {});
    }
```

packages/runtime/src/extensibility/facades/transport/RelationshipsFacade.ts
```diff
@@ -68,6 +80,22 @@ export class RelationshipsFacade {
         return await this.terminateRelationshipUseCase.execute(request);
     }
 
+    public async reactivateRelationship(request: ReactivateRelationshipRequest): Promise<Result<RelationshipDTO>> {
```
Milena-Czierlinski, line 83: What do you think about renaming this (and the use case/underlying RelationshipsController function) to `initiateRelationshipReactivation`? This would yield a nice consistency with the other names and is in accordance to what we did for the Identity deletion.

packages/runtime/src/useCases/transport/relationships/AcceptRelationshipReactivation.ts
Milena-Czierlinski, line 1: I'd also check if the Relationship has the correct status. The way it is currently implemented, a transport error would be thrown. Instead, if the status is incorrect, we should return a runtime error as Result.

Same for the other use cases.

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -191,32 +192,63 @@ export class RelationshipsController extends TransportController {
         if (!relationship) {
             throw CoreErrors.general.recordNotFound("Relationship", relationshipId.toString());
         }
+        if (!relationship.cache) {
+            await this.updateCacheOfRelationship(relationship);
+        }
+        if (!relationship.cache) {
+            throw this.newCacheEmptyError(Relationship, relationshipId.toString());
+        }
```
Milena-Czierlinski, line 200: Why is it necessary to do this manually in all functions? I'd expect the cache to be updated once a corresponding external events is received.

If this is actually necessary, it should probably happen in a separate function, as this piece of code is repeated quite often. Also, the two identical conditions on the same level look funny to me. Should the second rather be moved inside of the body of the first?

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -191,32 +192,63 @@ export class RelationshipsController extends TransportController {
         if (!relationship) {
             throw CoreErrors.general.recordNotFound("Relationship", relationshipId.toString());
         }
+        if (!relationship.cache) {
+            await this.updateCacheOfRelationship(relationship);
+        }
+        if (!relationship.cache) {
+            throw this.newCacheEmptyError(Relationship, relationshipId.toString());
```
Milena-Czierlinski, line 199: Probably out of context, but I wonder why this error is defined in the TransportController and not the CoreErrors file of the transport lib, where all other errors are defined (with error codes).

packages/runtime/src/extensibility/facades/transport/RelationshipsFacade.ts
Magnus-Kuhn: I agree with renaming, would personally tend to `requestRelationshipReactivation` and using it everywhere (the audit log entra reason is also called like this)

packages/runtime/src/useCases/transport/relationships/AcceptRelationshipReactivation.ts
Magnus-Kuhn: I'm generally not clear about if the validation should happen in transport or runtime - the cache check, that I copied, is redundant

jkoenig134: typically we do both

packages/transport/src/modules/relationships/RelationshipsController.ts
Magnus-Kuhn: this is because the cache is optional, so we need to throw something in case the cache is undefined. I copied this sequence from the completeOperation method - personally I would consider omitting the cache update, but didn't feel like this was worth fighting for.

packages/runtime/src/useCases/transport/relationships/AcceptRelationshipReactivation.ts
Magnus-Kuhn: so you suggest copy-pasting the whole validation? This seems inefficient for just a bit of runtime saving.

packages/transport/src/modules/relationships/RelationshipsController.ts
Magnus-Kuhn: Another entry on the "we need to discuss our refactoring existing coding approach" list

packages/runtime/src/useCases/transport/relationships/AcceptRelationshipReactivation.ts
jkoenig134: No, only the most important checks. We do that everywhere.

packages/runtime/src/extensibility/facades/transport/RelationshipsFacade.ts
Milena-Czierlinski: This would then yield a `RequestRelationshipReactivationRequest`. For this reason and for symmetry with Identity deletion I'd prefere `initiate...`. Having said that, there are also other places where our naming is not super pretty. So, if you feel strongly about it, I can also be convinced of another name.

packages/transport/test/end2end/End2End.test.ts
```diff
@@ -455,6 +455,357 @@ describe("RelationshipTest: Terminate", function () {
...
+        const revokedReactivatedRelationshipPeer = syncedRelationshipsPeer[0];
+
+        expect(revokedReactivatedRelationshipPeer.id.toString()).toStrictEqual(relationshipId.toString());
+        expect(revokedReactivatedRelationshipPeer.status).toStrictEqual(RelationshipStatus.Terminated);
+        expect(revokedReactivatedRelationshipPeer.cache?.auditLog).toHaveLength(5);
+        expect(revokedReactivatedRelationshipPeer.cache!.auditLog[4].reason).toBe(RelationshipAuditLogEntryReason.RevocationOfReactivation);
+    });
+});
+
+describe("RelationshipTest: validations (on terminated relationship", function () {
```
Milena-Czierlinski, line 605: missing ")"

packages/transport/test/end2end/End2End.test.ts
```diff
@@ -455,6 +455,357 @@ describe("RelationshipTest: Terminate", function () {
...
+
+    afterAll(async function () {
+        await from.close();
+        await to.close();
+        await connection.close();
+    });
+
+    test("should request reactivating a relationship between two accounts and accept the reactivation", async function () {
+        await from.relationships.reactivate(relationshipId);
+        await TestUtil.syncUntilHasRelationships(to);
```
Milena-Czierlinski, line 488: Again, I'd expect a separate test for `reactivate` (matching auditLog is created).

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -370,4 +381,78 @@ describe("RelationshipTermination", () => {
         });
         expect(result).toBeAnError(/.*/, "error.transport.challenges.challengeTypeRequiresActiveRelationship");
     });
+
+    test("should revoke the relationship reactivation", async () => {
+        await services1.transport.relationships.reactivateRelationship({ relationshipId });
+        await expect(services1.eventBus).toHavePublished(
+            RelationshipChangedEvent,
+            (e) => e.data.id === relationshipId && e.data.auditLog[e.data.auditLog.length - 1].reason === RelationshipAuditLogEntryReason.ReactivationRequested
+        );
```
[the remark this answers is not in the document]
Magnus-Kuhn: Relationships have the issue that there is no really clean way of cleaning up after a test - moving this into a different test would mean that this and the accept reactivation are both tests ending in a different state.
Maybe name the test "should initate/request ... and revoke" to signal that the request is tested here?

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -370,4 +381,78 @@ describe("RelationshipTermination", () => {
...
+
+        const relationship2 = (await syncUntilHasRelationships(services2.transport))[0];
+        expect(relationship2.status).toBe(RelationshipStatus.Terminated);
+        await expect(services1.eventBus).toHavePublished(
+            RelationshipChangedEvent,
+            (e) => e.data.id === relationshipId && e.data.auditLog[e.data.auditLog.length - 1].reason === RelationshipAuditLogEntryReason.RevocationOfReactivation
+        );
+
+        const reactivationResult = await services2.transport.relationships.acceptRelationshipReactivation({ relationshipId });
+        expect(reactivationResult).toBeAnError(/.*/, "error.transport.relationships.reactivationNotRequested");
```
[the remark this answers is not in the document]
Magnus-Kuhn: I wasn't trying to test all negative paths, I felt that doing them in transport was enough, this was just a different way of checking that the state is "terminated and reactivation requested"

packages/runtime/test/transport/relationships.test.ts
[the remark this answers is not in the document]
Magnus-Kuhn: On a second thought, the decomposition won't work like this - now I'm considering an `ensureTerminatedRelationship` in `beforeEach` (which is definitely going to be an ugly method)

packages/runtime/src/extensibility/facades/transport/RelationshipsFacade.ts
Magnus-Kuhn: Identity deletion doesn't require your peer to approve, while requesting the reactivation is a request in everything but technical realization.

packages/runtime/test/transport/relationships.test.ts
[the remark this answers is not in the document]
Milena-Czierlinski: I agree, it isn't necessary to test all the transport errors here again. I would however test the errors you throw and unsuccessful result you return in the runtime use cases.

packages/runtime/src/extensibility/facades/transport/RelationshipsFacade.ts
Milena-Czierlinski: I see your point. `requestRelationshipReactivation` is fine for me. 👍

packages/transport/src/modules/relationships/RelationshipsController.ts
Magnus-Kuhn: You were not happy with me putting `relationship.cache!` either

packages/runtime/test/transport/relationships.test.ts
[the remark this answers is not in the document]
Milena-Czierlinski: I'd say the more tests the better. 😇

packages/transport/src/modules/relationships/RelationshipsController.ts
Magnus-Kuhn: This error is not supposed to happen in production, just in development

jkoenig134: @Milena-Czierlinski CoreErrors (also all errors with error codes) are errors that the user of the library cares about. This error indicates a bad state of the library and the user will have no possibility to do anything about it.

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -187,47 +188,123 @@ export class RelationshipsController extends TransportController {
...
+    private async getRelationshipWithCache(id: CoreId): Promise<Relationship & { cache: CachedRelationship }> {
+        const relationship = await this.getRelationship(id);
+        if (!relationship) throw CoreErrors.general.recordNotFound(Relationship, id.toString());
+        if (!relationship.cache) await this.updateCacheOfRelationship(relationship);
+        if (!relationship.cache) throw this.newCacheEmptyError(Relationship, id.toString());
+
+        return relationship as Relationship & { cache: CachedRelationship };
+    }
+
+    private assertRelationshipStatus(relationship: Relationship, status: RelationshipStatus) {
```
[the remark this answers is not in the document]
jkoenig134: static meh. please don't do that.

`this.bla` is way cleaner than `RelationshipsController.bla`

jkoenig134:
@slavistan private methods don't need a declaration. we use typescript's inline hint for that.

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -187,47 +188,123 @@ export class RelationshipsController extends TransportController {
...
-        }
-        if (relationship.status !== RelationshipStatus.Pending) {
-            throw CoreErrors.relationships.wrongRelationshipStatus(relationship.status);
+        const relationship = await this.getRelationshipWithCache(relationshipId);
+        this.assertRelationshipStatus(relationship, RelationshipStatus.Pending);
+
+        const lastAuditLogEntry = relationship.cache.auditLog[relationship.cache.auditLog.length - 1];
+        if (!lastAuditLogEntry.createdBy.equals(relationship.peer.address)) {
+            throw CoreErrors.relationships.operationOnlyAllowedForPeer(`Only your peer can accept the relationship ${relationshipId.toString()}`);
         }
```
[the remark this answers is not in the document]
Magnus-Kuhn: It would work with the template, but the reactivation decisions can't use the template - so might as well have everything crash if things break

packages/runtime/src/useCases/transport/relationships/RejectRelationship.ts
```diff
@@ -34,6 +34,10 @@ export class RejectRelationshipUseCase extends UseCase<RejectRelationshipRequest
             return Result.fail(RuntimeErrors.general.cacheEmpty(Relationship, relationship.id.toString()));
         }
 
+        if (relationship.status !== RelationshipStatus.Pending) {
+            return Result.fail(RuntimeErrors.relationships.wrongRelationshipStatus(relationship.status));
+        }
+
```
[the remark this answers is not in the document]
Magnus-Kuhn: Idk, the idea was to have the most important validations duplicated in the runtime :shrug:

packages/transport/src/modules/sync/externalEventProcessors/RelationshipReactivationRequestedExternalEventProcessor.ts
```diff
@@ -0,0 +1,21 @@
...
+class RelationshipReactivationRequestedExternalEventData extends Serializable {
+    @serialize()
+    @validate()
+    public relationshipId: string;
+}
+
+export class RelationshipReactivationRequestedExternalEventProcessor extends ExternalEventProcessor {
+    public override async execute(externalEvent: ExternalEvent): Promise<Relationship | undefined> {
+        const payload = RelationshipReactivationRequestedExternalEventData.fromAny(externalEvent.payload);
+        const relationship = await this.accountController.relationships.applyRelationshipChangedEvent(payload.relationshipId);
```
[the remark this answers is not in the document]
Magnus-Kuhn: it fetches backbone data two calls deep

packages/runtime/test/transport/relationships.test.ts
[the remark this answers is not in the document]
jkoenig134: You cannot test cacheEmpty as transport should make sure that caches are never empty. recordNotFound should be tested. And please don't abuse error messages for checking the status of a relationship, this test reads **really** difficult!

packages/transport/src/modules/relationships/RelationshipsController.ts
jkoenig134: I already updated that with a typeguard, removed much copy pasted code.

packages/runtime/test/transport/relationships.test.ts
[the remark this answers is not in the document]
jkoenig134: why?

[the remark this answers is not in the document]
Magnus-Kuhn: because the negative tests are run in transport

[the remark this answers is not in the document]
jkoenig134: but you should make sure that the transport functionality is correctly called and **more important** that every error codes are correctly passed to the result.

packages/runtime/src/useCases/transport/relationships/AcceptRelationship.ts
```diff
@@ -25,23 +25,8 @@ export class AcceptRelationshipUseCase extends UseCase<AcceptRelationshipRequest
...
-            return Result.fail(RuntimeErrors.general.cacheEmpty(Relationship, relationship.id.toString()));
-        }
-
-        if (relationship.status !== RelationshipStatus.Pending) {
-            return Result.fail(RuntimeErrors.relationships.wrongRelationshipStatus(relationship.id.toString(), relationship.status));
-        }
-
-        const updatedRelationship = await this.relationshipsController.accept(relationship.id);
-
+        const updatedRelationship = await this.relationshipsController.accept(CoreId.from(request.relationshipId));
```
jkoenig134, line 28: I talked about not checking for the status. Please re-add recordNotFound and checkCache

Magnus-Kuhn: your reasoning is valid for those checks as well

packages/runtime/test/transport/relationships.test.ts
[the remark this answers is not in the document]
Magnus-Kuhn: for the passing of error codes I would have expected a unit test for the Result type - but in general our tests are also not fully implementation agnostic (e. g. we're not checking any combination of errors to make sure the correct one is returned, not sure if this is a good example). The way it is written, the only way for things to go wrong is for transport to mess up - are you saying that we shouldn't trust the used libraries?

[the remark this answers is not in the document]
jkoenig134: yeah, anyway please test all things that can go wrong.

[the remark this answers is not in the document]
jkoenig134: there are no unit tests for the result type .. and if there were it would not help in any way.

packages/runtime/src/useCases/transport/relationships/AcceptRelationship.ts
jkoenig134: no, this is a refactoring and we don't do refactorings in such PRs. When we decide to remove such checks we have to remove them anywere. It's different with your new checks, as they were added in this PR.

packages/transport/src/modules/sync/externalEventProcessors/RelationshipReactivationCompletedExternalEventProcessor.ts
```diff
@@ -0,0 +1,21 @@
...
+    @validate()
+    public relationshipId: string;
+}
+
+export class RelationshipReactivationCompletedExternalEventProcessor extends ExternalEventProcessor {
+    public override async execute(externalEvent: ExternalEvent): Promise<Relationship | undefined> {
+        const payload = RelationshipReactivationCompletedExternalEventData.fromAny(externalEvent.payload);
+        const relationship = await this.accountController.relationships.applyRelationshipChangedEvent(payload.relationshipId);
+
+        this.eventBus.publish(new RelationshipChangedEvent(this.ownAddress, relationship));
```
jkoenig134, line 18: I think it would be great to have an additional event to be published on the eventbus (RelationshipReactivationCompleted)

packages/transport/src/modules/sync/externalEventProcessors/RelationshipReactivationRequestedExternalEventProcessor.ts
```diff
@@ -0,0 +1,21 @@
...
+
+        this.eventBus.publish(new RelationshipChangedEvent(this.ownAddress, relationship));
```
jkoenig134, line 18: I think it would be great to have an additional event to be published on the eventbus (RelationshipReactivationRequested)

packages/transport/src/modules/sync/externalEventProcessors/RelationshipReactivationCompletedExternalEventProcessor.ts
jkoenig134: please make sure to also add the new event in the runtime event proxy

jkoenig134 (changes requested):
some small details

packages/transport/src/modules/sync/externalEventProcessors/RelationshipReactivationCompletedExternalEventProcessor.ts
```diff
@@ -0,0 +1,21 @@
...
+    @validate()
+    public relationshipId: string;
+}
+
+export class RelationshipReactivationCompletedExternalEventProcessor extends ExternalEventProcessor {
+    public override async execute(externalEvent: ExternalEvent): Promise<Relationship | undefined> {
+        const payload = RelationshipReactivationCompletedExternalEventData.fromAny(externalEvent.payload);
+        const relationship = await this.accountController.relationships.applyRelationshipChangedEvent(payload.relationshipId);
+
+        this.eventBus.publish(new RelationshipReactivationCompletedEvent(this.ownAddress, relationship));
```
jkoenig134, line 18: as said, please publish both events here (RelationshipChanged and RelationshipReactivationCompletedEvent)

packages/transport/src/modules/sync/externalEventProcessors/RelationshipReactivationRequestedExternalEventProcessor.ts
```diff
@@ -0,0 +1,21 @@
...
+    @validate()
+    public relationshipId: string;
+}
+
+export class RelationshipReactivationRequestedExternalEventProcessor extends ExternalEventProcessor {
+    public override async execute(externalEvent: ExternalEvent): Promise<Relationship | undefined> {
+        const payload = RelationshipReactivationRequestedExternalEventData.fromAny(externalEvent.payload);
+        const relationship = await this.accountController.relationships.applyRelationshipChangedEvent(payload.relationshipId);
+
+        this.eventBus.publish(new RelationshipReactivationRequestedEvent(this.ownAddress, relationship));
```
jkoenig134, line 18: see above

jkoenig134 (dismissed):

jkoenig134 (approved):

PR #134, opened 2024-05-23
Fix/SharedToPeerDVO without sourceAttribute

Milena-Czierlinski:
If a RepositoryAttribute is deleted, the shareInfo.sourceAttribute property of its own shared IdentityAttributes is set to undefined. Thus, we cannot classify own shared IdentityAttributes by this property. Instead, we must check if the Attribute owner and shareInfo.peer match, in order to know whether the IdentityAttribute is own shared or peer shared. 

Currently, own shared IdentityAttributes are incorrectly expanded to PeerAttributeDVOs instead of SharedToPeerAttributeDVOs. This poses a problem for the implementation of Attribute deletion in the App.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12332,7 +12332,7 @@
         },
         "packages/runtime": {
             "name": "@nmshd/runtime",
-            "version": "4.8.0",
+            "version": "4.8.1",
             "license": "MIT",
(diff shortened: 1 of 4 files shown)


britsta (approved):
LGTM

PR #136, opened 2024-05-24
Refactor/get rid of typeof ... "undefined"

jkoenig134:


jkoenig134:
@coderabbitai review

jkoenig134:
https://github.com/coderabbitai review

jkoenig134:
@coderabbitai review

packages/transport/src/core/backbone/RESTClient.ts
britsta, line 1: In line 373, there is the IF statement `if (typeof formData.getHeaders !== "undefined")`. As far as I know, this could probably be rephrased as well.

Also, in line 356, there is the logical statement `typeof Buffer !== "undefined"`. But that doesn't seem to be possible to reformulate. I also noticed something similar in `packages/transport/src/core/CoreError.ts` in line 61 with the IF statement `if (typeof Error.captureStackTrace !== "undefined")`.

.github/workflows/test.yml
Milena-Czierlinski, line 1: What does this do?

packages/transport/src/core/TransportContext.ts
Milena-Czierlinski, line 1: Why was this deleted?

britsta (commented):
General question: This PR is about getting rid of formulations like `typeof ... === "undefined"` or `typeof ... !== "undefined"`. What about formulations of the kind `... === undefined` or `... !== undefined` (for example the IF statement `if (address === undefined)` in line 31 of `packages/transport/src/core/types/CoreAddress.ts`), is an adjustment also wanted and possible for these?

.github/workflows/test.yml
Milena-Czierlinski: @sebbi08 just explained it to me

packages/transport/src/core/TransportContext.ts
jkoenig134: it was bad and unused

jkoenig134:
@britsta casting to boolean (for example `if (!"")`) with a nullish value would be false. So every time you handle optional strings (or more important booleans) you have to make sure be more accurate for js to understand what you want to do.

packages/transport/src/core/backbone/RESTClient.ts
jkoenig134: I left out some things that make a difference in browsers. I will not touch that Buffer stuff. And for `Error.captureStackTrace` the compiler thinks that it is always defined.

.github/workflows/test.yml
jkoenig134: If you find an answer elsewhere - please write that down for everyone.

This stops pipeline runs that belong to the same PR so that there is always only one run per PR. Should help us with out GH Actions limits.

Milena-Czierlinski:
Does this require a version bump for the small fixes?

jkoenig134:
I don't made fixes, so no, no version bump required.

Milena-Czierlinski (approved):

PR #141, opened 2024-05-28
Fix/https agent is not applied correctly

jkoenig134:


packages/transport/src/core/backbone/RESTClient.ts
```diff
@@ -81,17 +82,24 @@ export class RESTClient {
             defaults.headers = _.defaultsDeep({}, defaults.headers, this.config.platformAdditionalHeaders);
         }
 
+        if (config.httpsAgent) {
+            defaults.httpsAgent = config.httpsAgent;
+        } else {
+            try {
+                // eslint-disable-next-line @typescript-eslint/no-require-imports
+                const httpsAgent = require("https")?.Agent;
```
sebbi08, line 90: Why is this required here when it is imported at the start of the file ?

packages/transport/src/core/backbone/RESTClient.ts
```diff
@@ -81,17 +82,24 @@ export class RESTClient {
...
+
+                if (httpsAgent) defaults.httpsAgent = new httpsAgent(this.config.httpsAgentOptions);
+            } catch (e) {
+                // ignore
+            }
+        }
+
         try {
             // eslint-disable-next-line @typescript-eslint/no-require-imports
             const agent = require("http")?.Agent;
```
sebbi08, line 100: Same why is this required and not imported ?

packages/transport/src/core/backbone/RESTClient.ts
jkoenig134: it's not available in browsers

PR #149, opened 2024-05-29
Feature/Add sourceAttribute to RelationshipAttributeDVOs

Milena-Czierlinski:


packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -1077,7 +1077,7 @@ export class DataViewExpander {
                     description = value.description;
                 }
 
-                // Peer Relationship Attribute
+                // Peer shared RelationshipAttribute
```
jkoenig134, line 1080: instead of changing these comments, please yeet them

Milena-Czierlinski: Maybe in case of IdentityAttributeDVOs they actually are a good translation (PeerAttribute = Peer shared IdentityAttribute, SharedToPeerAttribute = Own shared IdentityAttribute). Of course renaming them would even be better, but breaking.

packages/runtime/test/dataViews/PeerAttributeDVO.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    services2 = runtimeServices[1];
+
+    await ensureActiveRelationship(services1.transport, services2.transport);
+}, 30000);
+
+afterAll(() => serviceProvider.stop());
+
+describe("PeerAttributeDVO", () => {
+    test("check the BirthYear", async () => {
+        const sOwnSharedIdentityAttribute = await executeFullCreateAndShareRepositoryAttributeFlow(services2, services1, {
```
britsta, line 25: I know what you mean by "sOwnSharedIdentityAttribute" and "rPeerSharedIdentityAttribute". I think for people who are not so familiar with the naming conventions, it can be difficult to understand what "s" and "r" is referring to. Therefore, perhaps use "sender" and "recipient" here instead. 

If this is too long for you, maybe instead of "sOwnSharedIdentityAttribute" it could also be written "s2OwnSharedIdentityAttribute" and instead of "rPeerSharedIdentityAttribute" it could also be written "s1PeerSharedIdentityAttribute" to refer to "services2" or "services1", respectively.

Same for other tests.
(But if you do not like any of the suggested options better, it is okay for me to leave it as it is for now. 😊)

Milena-Czierlinski: I used "s" and "r" prefixes, since they are used in a lot of different tests in different libs throughout our monorepo. "s1" and "s2" aren't more speaking to me.

jkoenig134 (approved):
there are some open comments, make sure to resolve them before merging

PR #151, opened 2024-06-03
Fix/do not pass https agent in config

jkoenig134:
This undoes a part of #141 with a better approach doing exactly the same thing.


slavistan:
Bro, give me a chance. 

image

image

slavistan:
image

PR #163, opened 2024-06-18
Chore/ Only use productive dependencies for audit

sebbi08:


slavistan (approved):

.ci/runChecks.sh
```diff
@@ -5,4 +5,4 @@ npm run build:node
 npm run lint:eslint
 npm run lint:prettier
 npx -ws license-check
-npx better-npm-audit audit --exclude 1096302
+npx better-npm-audit audit -p --exclude 1096302
```
jkoenig134, line 8: I don't like that...

PR #165, opened 2024-06-18
Feature/Default expiration date for files

Milena-Czierlinski:


packages/runtime/test/transport/files.test.ts
RuthDiG, line 1: perhaps you can change it to "can upload a file with undefined as expiry date and the expiry date will be automatically set to '9999-12-31'" to make it more specific.

britsta (commented):
- After merging this PR, will you adapt the documentation of the corresponding use case on enmeshed.eu (https://enmeshed.eu/use-case-transport-upload-own-file) or should someone else do it?
- When checking the FileDTO.ts and FileDVO.ts, I noticed that the optional `description` property was probably forgotten in the `FileDVO`. Perhaps this is the right opportunity to add this one line.

packages/runtime/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@nmshd/runtime",
-    "version": "4.12.1",
+    "version": "4.13.0",
```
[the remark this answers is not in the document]
Milena-Czierlinski: I figured being allowed to omit the expiresAt property and setting it automatically to a maximum value is rather a new feature than a bug fix.

Milena-Czierlinski:
> After merging this PR, will you adapt the documentation of the corresponding use case on enmeshed.eu (https://enmeshed.eu/use-case-transport-upload-own-file) or should someone else do it?

Of course I will take care of this ;) 

> When checking the FileDTO.ts and FileDVO.ts, I noticed that the optional description property was probably forgotten in the FileDVO. Perhaps this is the right opportunity to add this one line.

I think the FileDVO is fine, since it extend DataViewObject, which possesses a description property.

Milena-Czierlinski:
> In the CreateTokenForFile.ts file and in the CreateTokenQRCodeForFile.ts file, I have noticed the line const defaultTokenExpiry = file.cache?.expiresAt ?? CoreDate.utc().add({ days: 12 });. Perhaps it is undesired if this constant is set to 9999 in case the default value for the expiresAt property of the File is used.

Well observed, that is a good question. It seems to me that `file.cache.expiresAt` should usually be set, since `expiresAt` is a required property. So, it should only be undefined, if we don't have any cached data. I'm not sure, when this would happen. So, currently I could set the expiration date of a file one or multiple years in the future. This would mean that the `defaultTokenExpiry` also covers this time range. Setting the expiry date of the file to a maximum value, why should there be a problem doing the same for the expiry date of the token?

packages/runtime/package.json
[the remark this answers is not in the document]
slavistan: https://semver.org/ 😊 Patches are used for bug fixes. Minor versions are bumped when functionality is added without breaking changes to the public API (which is what we're doing here).

packages/runtime/src/useCases/transport/files/UploadOwnFile.ts
```diff
@@ -53,7 +53,7 @@ class Validator extends SchemaValidator<UploadOwnFileValidatableRequest> {
             );
         }
 
-        if (DateTime.fromISO(input.expiresAt) <= DateTime.utc()) {
+        if (input.expiresAt && DateTime.fromISO(input.expiresAt) <= DateTime.utc()) {
```
slavistan, line 56: Could you use our `CoreDate` library to perform the datetime arithmetic? It's already a necessary dependency. Not really a big deal, but it's smells funny to me to use two different datetime libraries within a few lines of code :-)

```ts
input.expiresAt && CoreDate.from(input.expiresAt).isSameOrBefore(new CoreDate())
```

packages/runtime/test/transport/files.test.ts
```diff
@@ -95,6 +96,14 @@ describe("File upload", () => {
         expect(response).toBeSuccessful();
     });
 
+    test("can upload a file with undefined as expiry date and automatically set expiry date to maximum value", async () => {
```
slavistan, line 99: Cumbersome description. How about "uploading a file without expiry date will use the default"?

packages/runtime/test/transport/files.test.ts
```diff
@@ -95,6 +96,14 @@ describe("File upload", () => {
...
+        const response = await transportServices1.files.uploadOwnFile(await makeUploadRequest({ expiresAt: undefined as unknown as string }));
+        expect(response).toBeSuccessful();
+
+        const file = response.value;
```
slavistan, line 103: In general, use the comparison semantics of datetime objects, not their string representations. Strings representations come in various formats with different precisions, time zone offsets etc.

```ts
const defaultDate = CoreDate.from("9999-12-31T00:00:00.000Z")
expect(CoreDate.from(file.expiresAt).isSame(defaultDate)).toBe(true)

// Alternatively, you can write a custom assertion for CoreDate here:
// packages/runtime/test/customMatchers.ts
// Maybe something like isSameDate
```

britsta:
> Well observed, that is a good question. It seems to me that `file.cache.expiresAt` should usually be set, since `expiresAt` is a required property. So, it should only be undefined, if we don't have any cached data. I'm not sure, when this would happen. So, currently I could set the expiration date of a file one or multiple years in the future. This would mean that the `defaultTokenExpiry` also covers this time range. Setting the expiry date of the file to a maximum value, why should there be a problem doing the same for the expiry date of the token?

That's right, the value of 12 days as the `defaultTokenExpiry` should not currently be able to be used, which was already the case before. Nevertheless, it has made me think that such a short period of time is specified here as the default and such a distant year for the File with 9999. For security reasons, I could imagine that it would be better to have a shorter token validity by default, but I do not know much about tokens and whether it would also be conceivable to simply write 9999 too. Maybe someone else knows more about this.

packages/runtime/src/useCases/transport/files/UploadOwnFile.ts
```diff
@@ -53,7 +52,7 @@ class Validator extends SchemaValidator<UploadOwnFileValidatableRequest> {
             );
         }
 
-        if (input.expiresAt && DateTime.fromISO(input.expiresAt) <= DateTime.utc()) {
+        if (input.expiresAt && CoreDate.from(input.expiresAt).isSameOrBefore(new CoreDate())) {
```
jkoenig134, line 55: please use CoreDate.now() instead of new CoreDate, IMO this is way more descriptive.

Milena-Czierlinski: I guess you mean CoreDate.utc(), don't you?

jkoenig134: yep

jkoenig134 (approved):

Milena-Czierlinski:
> [BS] In the CreateTokenForFile.ts file and in the CreateTokenQRCodeForFile.ts file, I have noticed the line const defaultTokenExpiry = file.cache?.expiresAt ?? CoreDate.utc().add({ days: 12 });. Perhaps it is undesired if this constant is set to 9999 in case the default value for the expiresAt property of the File is used.

> [MC] Well observed, that is a good question. It seems to me that file.cache.expiresAt should usually be set, since expiresAt is a required property. So, it should only be undefined, if we don't have any cached data. I'm not sure, when this would happen. So, currently I could set the expiration date of a file one or multiple years in the future. This would mean that the defaultTokenExpiry also covers this time range. Setting the expiry date of the file to a maximum value, why should there be a problem doing the same for the expiry date of the token?

> [BS] That's right, the value of 12 days as the defaultTokenExpiry should not currently be able to be used, which was already the case before. Nevertheless, it has made me think that such a short period of time is specified here as the default and such a distant year for the File with 9999. For security reasons, I could imagine that it would be better to have a shorter token validity by default, but I do not know much about tokens and whether it would also be conceivable to simply write 9999 too. Maybe someone else knows more about this.

@jkoenig134 Do you have an opinion on this open discussion?

jkoenig134:
@Milena-Czierlinski should be ok

PR #166, opened 2024-06-24
Fix/bump vulnerable dev subdependency of iql

slavistan:
ws had a vulnerability whose fix was backported to 7.5.10.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2612,8 +2612,9 @@
         },
         "node_modules/accepts": {
             "version": "1.3.8",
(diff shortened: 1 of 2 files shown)


slavistan:
For some reason our package-lock.json was missing the fields `resolved` and `integrity` (the package's download source and the file hash, respectively) in the dependency specs. There may be several reasons for this
- legacy behavior from older npm versions
- bugs/quirks of npm (e.g. [issue 4263](https://github.com/npm/cli/issues/4263))

That's the reason for the extra additions to the package-lock.json. All that I was after was that `ws` is using `7.5.10` instead of `7.5.9` :-)

PR #169, opened 2024-06-25
chore/undo audit misconfiguration

jkoenig134:


jkoenig134:
it's way to restrictive to only check dependencies and just to ignore dev-deps. PLease ignore audits that we cannot change for the future.

slavistan (approved):

PR #170, opened 2024-06-25
Feature/restrict "any" content types

Magnus-Kuhn:


packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -14,7 +14,7 @@ export interface SendMessageRequest {
      * @minItems 1
      */
     recipients: AddressString[];
-    content: any;
+    content: MailJSON | IMail | ResponseWrapperJSON | IResponseWrapper | RequestJSON | IRequest | NotificationJSON | INotification;
```
jkoenig134, line 17: please only use jsons here

packages/runtime/src/types/transport/RelationshipDTO.ts
```diff
@@ -32,12 +33,14 @@ export interface RelationshipAuditLogEntryDTO {
 
 export interface RelationshipAuditLogDTO extends Array<RelationshipAuditLogEntryDTO> {}
 
+export type RelationshipCreationContentDTO = RelationshipCreationContentJSON | ArbitraryRelationshipCreationContentJSON;
```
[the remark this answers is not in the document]
Magnus-Kuhn: @jkoenig134 didn't like the previous version of RelationshipCreationContentContainingResponse, and I'm not seeing a way of adjusting RelationshipCreationContentDTO without it looking ugly and misleading in the direction of arbitraryContent right now. I would like to keep the type safety rather than revert to any

packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -285,7 +285,7 @@ export class DataViewExpander {
             wasReadAt: message.wasReadAt
         };
 
-        if (message.content["@type"] === "Mail" || message.content["@type"] === "RequestMail") {
```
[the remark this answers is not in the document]
Magnus-Kuhn: There was a comment previously somewhere stating `RequestMail` was for backwards compatibility - but I guess we can ditch it with v5 as well. Then I'll remove the other line as well

packages/content/src/messages/ArbitraryMessageContent.ts
```diff
@@ -0,0 +1,25 @@
+import { ISerializable, Serializable, serialize, type, validate } from "@js-soft/ts-serval";
+import { ContentJSON } from "../ContentJSON";
+
+export interface ArbitraryMessageContentJSON extends ContentJSON {
```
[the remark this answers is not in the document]
Magnus-Kuhn: yeah, this or push the actual content one level up by having something like `Map<string,...>` instead of `content`

packages/runtime/src/types/transport/MessageDTO.ts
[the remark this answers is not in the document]
Magnus-Kuhn: Transport wants to be more abstract and independent of Content, the typing happens when moving from Transport to Runtime

packages/runtime/src/dataViews/DataViewExpander.ts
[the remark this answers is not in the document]
Magnus-Kuhn: Sorry, misremembered that comment, but I still think removal is correct

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -58,23 +58,30 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+            const recipient = CoreAddress.from(recipients[0]);
+            if (!recipient.equals(localRequest.peer)) return RuntimeErrors.general.invalidPropertyValue("The recipient does not match the Request's peer.");
+            return;
+        } else if (
+            transformedContent instanceof Mail ||
+            transformedContent instanceof ResponseWrapper ||
+            transformedContent instanceof Notification ||
+            transformedContent instanceof ArbitraryMessageContent
+        ) {
+            return;
```
Milena-Czierlinski, line 82: I think an early return is preferred here, i.e. test for the valid types in the beginning and throw an error in case of an invalid type. Then, continue with the validation for Requests.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -58,23 +58,30 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
         }
-
-        if (recipients.length > 1) return RuntimeErrors.general.invalidPropertyValue("Only one recipient is allowed for sending Requests.");
-
-        const recipient = CoreAddress.from(recipients[0]);
-        if (!recipient.equals(localRequest.peer)) return RuntimeErrors.general.invalidPropertyValue("The recipient does not match the Request's peer.");
-
-        return;
+        return RuntimeErrors.general.invalidPropertyValue("The content type of a message must be Mail, Request, ResponseWrapper, Notification or ArbitraryMessageContent.");
```
Milena-Czierlinski, line 84: I'd capitalize Message here for consistency.

packages/runtime/src/useCases/transport/relationshipTemplates/CreateOwnRelationshipTemplate.ts
```diff
@@ -68,16 +68,19 @@ export class CreateOwnRelationshipTemplateUseCase extends UseCase<CreateOwnRelat
...
-        if (transformedContent.onExistingRelationship) {
-            const validationResult = await this.outgoingRequestsController.canCreate({ content: transformedContent.onExistingRelationship });
-            if (validationResult.isError()) return validationResult.error;
+            return;
+        } else if (transformedContent instanceof ArbitraryRelationshipTemplateContent) {
+            return;
         }
-
-        return;
+        return RuntimeErrors.general.invalidPropertyValue("A relationship template content must be of type RelationshipTemplateContent or ArbitraryRelationshipTemplateContent.");
```
Milena-Czierlinski, line 84: How about: "The content of a RelationshipTemplate must be of type RelationshipTemplateContent or ArbitraryRelationshipTemplateContent."?

packages/runtime/src/useCases/transport/relationshipTemplates/RelationshipTemplateMapper.ts
```diff
@@ -7,6 +8,11 @@ export class RelationshipTemplateMapper {
         if (!template.cache) {
             throw RuntimeErrors.general.cacheEmpty(RelationshipTemplate, template.id.toString());
         }
+        if (!(template.cache.content instanceof RelationshipTemplateContent || template.cache.content instanceof ArbitraryRelationshipTemplateContent)) {
+            throw RuntimeErrors.general.invalidPropertyValue(
+                `The content type of relationship template ${template.id} is neither RelationshipTemplateContent nor ArbitraryRelationshipTemplateContent.`
```
Milena-Czierlinski, line 13: RelationshipTemplate instead of relationship template

packages/runtime/src/useCases/transport/relationships/CreateRelationship.ts
```diff
@@ -27,18 +29,23 @@ export class CreateRelationshipUseCase extends UseCase<CreateRelationshipRequest
...
-        await this.accountController.syncDatawallet();
+            await this.accountController.syncDatawallet();
+
+            return Result.ok(RelationshipMapper.toRelationshipDTO(relationship));
+        }
 
-        return Result.ok(RelationshipMapper.toRelationshipDTO(relationship));
+        return Result.fail(
+            RuntimeErrors.general.invalidPropertyValue("A relationship creation content must be of type RelationshipCreationContent or ArbitraryRelationshipCreationContent.")
+        );
```
Milena-Czierlinski, line 49: Here again think about changing the order of returning an error result and the actual content of the use case.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
Magnus-Kuhn: putting the happy path first allows for cleaner ifs

packages/runtime/src/useCases/transport/relationships/CreateRelationship.ts
Magnus-Kuhn: I'd still go for the cleaner if-condition

Magnus-Kuhn:
on a second thought: do we actually need arbitrary contents for relationship template / relationship creation / messages?
For the relationship things the use I've thought of was to allow additional explanations for why you would want the relationship to help decide if you create resp. accept it - but I guess you'd have figured out who you are willing to enter a relationship with beforehand and mainly send authentication data with the relationship (template).
Even then, it would be possible to add e. g. a freeText property to the RelationshipTemplateContent etc.
As for messages, I don't see a use for ArbitraryMessageContent.

jkoenig134:
It's all about Connector to Connector communication. We don't want everyone have to strictly use our request Flows.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
jkoenig134: @Magnus-Kuhn I am 100% with Milena. we don't use `else if` in the code if avoidable and this is the perfect place to avoid it.

stnmtz (commented):
- Changes in tests look somehow strange to me, so many ! and toJSON() - but that seems to be the case because of the former unknown/any types.
- Arbitrary* naming is ok for me

packages/runtime/test/transport/relationshipTemplates.test.ts
```diff
@@ -92,34 +92,23 @@ describe("Template Tests", () => {
...
-            content: { a: "A", "@type": "Message" },
-            expiresAt: DateTime.utc().plus({ minutes: 1 }).toString()
+            content: {},
+            expiresAt: DateTime.utc().plus({ minutes: 1 }).toString(),
+            maxNumberOfAllocations: 1
         });
-        expect(response).toBeAnError("Message.secretKey :: Value is not defined", "error.runtime.requestDeserialization");
-    });
 
-    test("create a template with wrong content : not existent type", async () => {
```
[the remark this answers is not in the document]
Magnus-Kuhn: re-added it, don't know why I removed it :)

packages/runtime/src/types/transport/RelationshipTemplateDTO.ts
```diff
@@ -1,10 +1,14 @@
+import { ArbitraryRelationshipTemplateContentJSON, RelationshipTemplateContentJSON } from "@nmshd/content";
+
+export type RelationshipTemplateContentDTO = RelationshipTemplateContentJSON | ArbitraryRelationshipTemplateContentJSON;
```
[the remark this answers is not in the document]
britsta: When defining `RelationshipTemplateContentDTO = RelationshipTemplateContentJSON | ArbitraryRelationshipTemplateContentJSON`,  the term `RelationshipTemplateContent` has different meanings in different contexts:
- `RelationshipTemplateContentDTO` would refer to the `content` of the RelationshipTemplate, what sounds good :D
- However, `RelationshipTemplateContentJSON` refers only to the structured RelationshipTemplateContent containing Requests and not also to an arbitrary RelationshipTemplateContent.

A renaming of `RelationshipTemplateContentJSON` would therefore probably be more precise (for example, `RelationshipTemplateContentWithRequestsJSON` or `StructuredRelationshipTemplateContentJSON`). However, it is not easy to think of a more favorable name... 🐸 ...as @Magnus-Kuhn mentioned in another comment on a similar naming issue below.

packages/content/src/messages/ArbitraryMessageContent.ts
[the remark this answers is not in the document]
Magnus-Kuhn: > We should be careful however, because the Arbitrary data could contain @ type which is not an "enmeshed @ type"

Why are we deleting @ type anyway?

packages/runtime/src/types/transport/RelationshipTemplateDTO.ts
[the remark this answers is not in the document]
jkoenig134: `RelationshipTemplateContentDTO` is not our naming schema. When we call it `RelationshipTemplateContentDerivation` it gets way clearer. Or we just inline this, then we don't have to discuss naming.

packages/content/src/messages/ArbitraryMessageContent.ts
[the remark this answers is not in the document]
jkoenig134: > We should be careful however, because the Arbitrary data could contain @ type which is not an "enmeshed @ type"

Yes! Please don't use JSONWrapper!

packages/runtime/src/types/transport/RelationshipTemplateDTO.ts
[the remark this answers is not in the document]
Magnus-Kuhn: going with the Derivation, I want to be lazy with writing later on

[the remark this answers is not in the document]
jkoenig134: I think you misread my comment. I was talking about a `RelationshipTemplateContentDerivation` not a `RelationshipTemplateContentDerivationDTO`, both is too much.

packages/content/src/messages/ArbitraryMessageContent.ts
```diff
@@ -0,0 +1,25 @@
...
+    "@type": "ArbitraryMessageContent";
+    content: any;
+}
+export interface IArbitraryMessageContent extends ISerializable {
+    content: any;
+}
+
+@type("ArbitraryMessageContent")
+export class ArbitraryMessageContent extends Serializable implements IArbitraryMessageContent {
```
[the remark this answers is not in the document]
Magnus-Kuhn: I've created a ticket for discussing documenting this on the internal board.

packages/runtime/test/dataViews/RelationshipTemplateDVO.test.ts
```diff
@@ -173,7 +176,8 @@ describe("RelationshipTemplateDVO", () => {
     });
 
     test("TemplateDVO for requestor", async () => {
-        const requestorTemplate = (await requestor.transport.relationshipTemplates.loadPeerRelationshipTemplate({ reference: templatorTemplate.truncatedReference })).value;
+        const requestorTemplate = (await requestor.transport.relationshipTemplates.loadPeerRelationshipTemplate({ reference: templatorTemplate.truncatedReference }))
```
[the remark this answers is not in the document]
Magnus-Kuhn: This is a similar discussion to how to deal with the different template content types where in the end I settled on using RelationshipTemplateContentDerivation. However, here the name RelationshipTemplateDerivation(DTO) seems clunkier. Also we usually deal with situations like this by performing the cast later, could my early cast have been what's strange?

packages/content/src/relationships/ArbitraryRelationshipCreationContent.ts
```diff
@@ -0,0 +1,26 @@
...
+    value: any;
+}
+
+export interface IArbitraryRelationshipCreationContent extends ISerializable {
+    value: any;
+}
+
+@type("ArbitraryRelationshipCreationContent")
+export class ArbitraryRelationshipCreationContent extends Serializable implements IArbitraryRelationshipCreationContent {
+    @serialize()
```
jkoenig134, line 15: @stnmtz JSONWrapper uses `@serialize({ any: true })` for any contents.

1. what's the purpose of that?
2. should we use that for these `Arbitrary*` classes too?

packages/content/src/relationships/ArbitraryRelationshipCreationContent.ts
```diff
@@ -0,0 +1,26 @@
...
+    @validate()
+    public value: any;
```
jkoenig134, line 17: @stnmtz @tnotheis wouldn't `unknown` be the better option here (also in the interface and JSON)?

tnotheis: Isn't `unknown` **always** the better choice? At least I can't see any major downside in this case.

packages/runtime/test/dataViews/RelationshipTemplateDVO.test.ts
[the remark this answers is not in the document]
jkoenig134: @stnmtz what do you think about making it generic and write it as `RelationshipTemplateDTO<RelationshipTemplateContentJSON>`?

packages/runtime/test/transport/messages.test.ts
```diff
@@ -165,6 +165,14 @@ describe("Message errors", () => {
         expect(result).toBeAnError("Mail.to :: Value is not defined", "error.runtime.requestDeserialization");
     });
 
+    test("should throw correct error for false content type", async () => {
+        const result = await client1.transport.messages.sendMessage({
+            recipients: [client2.address],
+            content: {}
```
jkoenig134, line 171: why is this possible without tricking the compiler?

packages/runtime/test/transport/messages.test.ts
```diff
@@ -321,11 +329,23 @@ describe("Message query", () => {
...
 
         await client1.transport.messages.sendMessage({
-            content: {},
+            content: {
+                "@type": "Mail",
+                body: "A body",
+                cc: [],
+                subject: "A subject",
+                to: [addressRecipient1]
+            },
```
jkoenig134, line 338: why don't you use arbitrary / empty here?

packages/runtime/test/transport/messages.test.ts
```diff
@@ -321,11 +329,23 @@ describe("Message query", () => {
...
         });
         await client1.transport.messages.sendMessage({
-            content: {},
+            content: {
+                "@type": "Mail",
+                body: "A body",
+                cc: [],
+                subject: "A subject",
+                to: [addressRecipient2]
+            },
```
jkoenig134, line 348: why don't you use arbitrary / empty here?

jkoenig134 (changes requested):

packages/runtime/src/modules/RequestModule.ts
```diff
@@ -286,11 +281,14 @@ export class RequestModule extends RuntimeModule {
         const templateId = template.id;
         // do not trigger for templates without the correct content type
         if (template.content["@type"] !== "RelationshipTemplateContent") return;
+        if (createdRelationship.creationContent["@type"] !== "RelationshipCreationContent") {
+            this.logger.error(`The creation content of relationshipId ${createdRelationship.id} is not of type RelationshipCreationContent.`);
```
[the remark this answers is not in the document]
Magnus-Kuhn: whoops

packages/runtime/src/modules/RequestModule.ts
```diff
@@ -134,6 +128,7 @@ export class RequestModule extends RuntimeModule {
                     response: responseWrapper.response
                 });
                 break;
+            default:
```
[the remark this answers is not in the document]
Magnus-Kuhn: It's probably better to list the remaining cases, I'll change it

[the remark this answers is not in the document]
jkoenig134:
```
default:
  break;
```
would also be ok for me, I just dont like completely empty cases.

packages/runtime/src/useCases/transport/relationships/CreateRelationship.ts
```diff
@@ -32,10 +34,14 @@ export class CreateRelationshipUseCase extends UseCase<CreateRelationshipRequest
...
-            creationContent: request.creationContent
-        });
+        const transformedContent = Serializable.fromUnknown(request.creationContent);
+        if (!(transformedContent instanceof ArbitraryRelationshipCreationContent || transformedContent instanceof RelationshipCreationContent)) {
+            return Result.fail(
+                RuntimeErrors.general.invalidPropertyValue("A relationship creation content must be of type RelationshipCreationContent or ArbitraryRelationshipCreationContent.")
+            );
+        }
+
+        const relationship = await this.relationshipsController.sendRelationship({ template, creationContent: request.creationContent });
```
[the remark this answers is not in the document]
Magnus-Kuhn: I can change it but I don't see the difference

packages/runtime/test/consumption/requests.test.ts
```diff
@@ -349,8 +350,7 @@ describe("Requests", () => {
         let rTransportServices: TransportServices;
         let rEventBus: EventBus;
 
-        const templateContent = {
-            "@type": "RelationshipTemplateContent",
+        const templateContent = RelationshipTemplateContent.from({
```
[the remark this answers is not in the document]
Magnus-Kuhn: why should this be a safer way?

packages/runtime/test/transport/messages.test.ts
Magnus-Kuhn: We decided against fixing the type because the schema validator would fuck up

packages/runtime/src/useCases/transport/relationships/CreateRelationship.ts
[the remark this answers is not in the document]
jkoenig134: Serializable.fromJSON removes extra values, in request.creationContent they are still available

packages/runtime/test/consumption/requests.test.ts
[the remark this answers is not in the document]
jkoenig134: it's the way it's done when consuming the runtime

packages/runtime/test/transport/messages.test.ts
jkoenig134: IC, great.

packages/runtime/src/modules/RequestModule.ts
```diff
@@ -134,6 +128,9 @@ export class RequestModule extends RuntimeModule {
                     response: responseWrapper.response
                 });
                 break;
+            case "ArbitraryMessageContent":
+            case "Mail":
+            case "Notification":
```
jkoenig134, line 133: please add breaks for all of these cases or re-add default and add a break for this. This was my intention for the comment i had before.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -58,6 +58,18 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+        if (
+            !(
+                transformedContent instanceof Mail ||
+                transformedContent instanceof ResponseWrapper ||
+                transformedContent instanceof Notification ||
+                transformedContent instanceof ArbitraryMessageContent ||
+                transformedContent instanceof Request
+            )
+        ) {
+            return RuntimeErrors.general.invalidPropertyValue("The content type of a Message must be Mail, Request, ResponseWrapper, Notification or ArbitraryMessageContent.");
```
jkoenig134, line 70: I see some problems with the error message (this also will apply to other places)

- what is "the content type" and how does the user know what you mean by that
- Mail etc. should be in single ticks

I suggest
```suggestion
            return RuntimeErrors.general.invalidPropertyValue("The content's '@type‘ of a Message must be ‘Mail‘, ‘Request‘, ‘ResponseWrapper‘, ‘Notification‘ or ‘ArbitraryMessageContent‘.");
```
but I would be happy to hear other thoughts (cc @Milena-Czierlinski, @britsta, @slavistan)

packages/runtime/src/modules/RequestModule.ts
```diff
@@ -1,5 +1,6 @@
 import { LocalRequestStatus } from "@nmshd/consumption";
-import { RelationshipCreationContent, RelationshipTemplateContentJSON, RequestJSON, ResponseJSON, ResponseResult, ResponseWrapper, ResponseWrapperJSON } from "@nmshd/content";
+import { RelationshipCreationContent, RequestJSON, ResponseJSON, ResponseResult, ResponseWrapper } from "@nmshd/content";
+
```
jkoenig134, line 3: formatting

packages/runtime/src/useCases/transport/relationshipTemplates/CreateOwnRelationshipTemplate.ts
```diff
@@ -68,7 +68,16 @@ export class CreateOwnRelationshipTemplateUseCase extends UseCase<CreateOwnRelat
...
+
+        if (!(transformedContent instanceof RelationshipTemplateContent || transformedContent instanceof ArbitraryRelationshipTemplateContent)) {
+            return RuntimeErrors.general.invalidPropertyValue(
+                "The content of a RelationshipTemplate must be of type RelationshipTemplateContent or ArbitraryRelationshipTemplateContent."
+            );
+        }
+
+        if (transformedContent instanceof ArbitraryRelationshipTemplateContent) {
+            return;
+        }
```
jkoenig134, line 80:
```suggestion
        if (!(transformedContent instanceof RelationshipTemplateContent)) return;
```
this is future proof and will not explode if we add other types

packages/content/src/relationships/ArbitraryRelationshipCreationContent.ts
jkoenig134: @stnmtz wrote

> Ich glaub wir müssen any benutzen, ansonsten meckert er, wenn es z.B. undefined ist - und weil er keinen Datentyp bekommt, weiß er nicht genau wie er das zu validieren hat. Aber nur eine Vermutung

jkoenig134: also IMO it doesn't hurt to just use it.

jkoenig134: erm @Magnus-Kuhn.. why did you remove the unknown keyword?

Magnus-Kuhn: I thought we were doing things JSONWrapper-style...

jkoenig134: nope. we use `@serialize({ any: true })` to disable some ts-serval validation and serialization logic, but we still want the compiler to teat it as arbitrary content.

Additionally: JSONWrapper should also use unknown instead of any, back then we didn't knew about this.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
Milena-Czierlinski: For me "content type" is clear, but I don't know how an external user would think about it and I don't mind your suggestion. Only note, please, that you use varying single ticks.

jkoenig134: true, please don't use these strange ticks @Magnus-Kuhn :D

britsta: In principle, please do not use single ticks to enclose types from the data model overview. We don't do this otherwise either, but only for actual **values** (for example 'ID of a LocalAttribute'). Compare with the comment on this from @jkoenig134 in this PR https://github.com/nmshd/runtime/pull/179#issuecomment-2225292069.

Therefore, for reasons of consistency, `"The 'content' of a Message must contain either a Mail, a Request, a ResponseWrapper, a Notification or an ArbitraryMessageContent."` would be my suggestion.

Magnus-Kuhn: "The content must contain" was a chalk on chalkboard moment for me - but I like it with "The content must be"

britsta: 😂😂

jkoenig134: @britsta IMO this message is not clear enough.

I talk about actual strings, we use 's when we describe ids or other strings, so my formatting should be correct.

jkoenig134: So when I consider my comment from the other PR that you linked 

`"The content's @type of a Message must be 'Mail', 'Request', 'ResponseWrapper', 'Notification' or 'ArbitraryMessageContent'."`

jkoenig134: When not talking about properties I would write it as `"The content of a Message must either be a Mail, Request, ResponseWrapper, Notification or an ArbitraryMessageContent."`

britsta: Since in this variant proposed by you the types are not used directly as words in the sentence, but are referred to abstractly, it is okay for me to enclose them here exceptionally with single ticks (even though I would not do it to keep consistency as much as possible). However, types from the data model overview should generally not be enclosed with single ticks in error messages and the documentation as well.

jkoenig134: > Since in this variant proposed by you the types are not used directly as words in the sentence

again, in my first variant I **don't** talk about types, I talk about the value that can be stored in the `@type` field of the content. This is very convenient for the user as he don't has to understand our wording.

britsta: Yeah, so let's write `"The content's @type of a Message must be 'Mail', 'Request', 'ResponseWrapper', 'Notification' or 'ArbitraryMessageContent'."` or `"The content of a Message must either be a Mail, Request, ResponseWrapper, Notification or an ArbitraryMessageContent."`, right @jkoenig134? 😎 I'd prefer last, but first suggestion equally consistent to our formatting style and fine for me.

jkoenig134: I'm kind of ok with the second one, so let's take this.

jkoenig134 (approved):

jkoenig134:
you missed updating the tests :/

jkoenig134 (approved):

PR #172, opened 2024-06-28
Feature/Wrap Backbone error peerIsToBeDeleted

Milena-Czierlinski:


packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -154,12 +154,22 @@ export class RelationshipsController extends TransportController {
...
+                })
+            ).value;
+        } catch (error: any) {
+            if (error.code === "error.platform.validation.relationship.peerIsToBeDeleted") {
+                throw new Error(
+                    "The Identity who created the RelationshipTemplate is currently in the process of deleting itself. Thus, it is not possible to establish a Relationship to it."
+                );
+            }
+            throw error;
+        }
```
jkoenig134, line 172: you don't need try..catch - we have a result for this. If you don't call `.value` on the result of `createRelationship` you can just look if everything went through.

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -154,12 +154,22 @@ export class RelationshipsController extends TransportController {
...
+                await this.client.createRelationship({
+                    content: requestCipher.toBase64(),
+                    relationshipTemplateId: template.id.toString()
+                })
+            ).value;
+        } catch (error: any) {
+            if (error.code === "error.platform.validation.relationship.peerIsToBeDeleted") {
+                throw new Error(
+                    "The Identity who created the RelationshipTemplate is currently in the process of deleting itself. Thus, it is not possible to establish a Relationship to it."
+                );
```
jkoenig134, line 169: shouldn't this be a CoreError with a proper error code?

Milena-Czierlinski: I thought we use CoreErrors with ErrorCodes only in case the user did something wrong.

jkoenig134: The user tried to open a relationship to an identity that is deleting itself. The user cannot really know this, but it is the users fault.

If we don't give out a CoreError, why do we need this PR at all?

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -74,6 +74,28 @@ describe("Create Relationship", () => {
...
+        services2 = runtimeServices[1];
+
+        const templateId = (await exchangeTemplate(services1.transport, services2.transport)).id;
+        await services1.transport.identityDeletionProcesses.initiateIdentityDeletionProcess();
+
+        const createRelationshipResponse = await services2.transport.relationships.createRelationship({
+            templateId: templateId,
+            content: { a: "b" }
+        });
+        expect(createRelationshipResponse.isSuccess).toBe(false);
```
jkoenig134, line 91: this can be removed as you integrated the toBeAnError utility

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -74,6 +74,28 @@ describe("Create Relationship", () => {
         expect(relationship2Response).toBeSuccessful();
         expect(relationship2Response.value.status).toBe("Active");
     });
+
+    test("returns error if templator has active IdentityDeletionProcess", async () => {
+        // create new services that don't have a relationship yet
+        const runtimeServices = await serviceProvider.launch(2, { enableRequestModule: true, enableDeciderModule: true, enableNotificationModule: true });
+        services1 = runtimeServices[0];
+        services2 = runtimeServices[1];
```
jkoenig134, line 82: you never close the runtimeServices. Please wrap this test in a describe block and put the init in a beforeAll and the closing in an afterAll block

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -74,6 +74,28 @@ describe("Create Relationship", () => {
...
+            templateId: templateId,
+            content: { a: "b" }
+        });
+        expect(createRelationshipResponse.isSuccess).toBe(false);
+        expect(createRelationshipResponse).toBeAnError(
+            "The Identity who created the RelationshipTemplate is currently in the process of deleting itself. Thus, it is not possible to establish a Relationship to it.",
+            "error.transport.relationships.activeIdentityDeletionProcessOfOwnerOfRelationshipTemplate"
+        );
+
+        await services1.transport.identityDeletionProcesses.cancelIdentityDeletionProcess();
```
jkoenig134, line 97: this distracts. as you use throwAway identities for this test you can just remove that.

jkoenig134 (approved):
LGTM 🥳

PR #173, opened 2024-06-28
Feature/delete attributes when relationship is rejected

britsta:


jkoenig134 (dismissed):

packages/runtime/test/modules/RequestModule.test.ts
```diff
@@ -420,3 +426,143 @@ describe("RequestModule", () => {
...
+        await rRuntimeServices.eventBus.waitForRunningEventHandlers();
+        await syncUntilHasRelationships(sRuntimeServices.transport, 1);
+        await sRuntimeServices.eventBus.waitForRunningEventHandlers();
+
+        expect((await sRuntimeServices.consumption.attributes.getAttributes({})).value).toHaveLength(1);
+        expect((await rRuntimeServices.consumption.attributes.getAttributes({})).value).toHaveLength(1);
+    });
+});
+
+async function ensurePendingRelationshipWithTemplate(
```
jkoenig134, line 508: this function should be extracted to testUtils

packages/runtime/test/lib/testUtils.ts
```diff
@@ -336,6 +340,109 @@ export async function establishRelationshipWithContents(
     expect(relationships2).toHaveLength(1);
 }
 
+export async function establishPendingRelationshipWithRequestFlow(
+    sRuntimeServices: TestRuntimeServices,
+    rRuntimeServices: TestRuntimeServices,
+    existingRelationshipAttributeForFurtherSharing?: LocalAttributeDTO
+): Promise<RelationshipDTO> {
```
jkoenig134, line 347: @britsta this is not _exactly_ how we discussed it. IMO this would be way more useful if you can throw in the whole items of the request and the items for the accept:

```ts
export async function establishPendingRelationshipWithRequestFlow(
    sRuntimeServices: TestRuntimeServices,
    rRuntimeServices: TestRuntimeServices,
    requestItems: (RequestItemGroupJSON | RequestItemJSONDerivations)[],
    acceptParams: (DecideRequestItemParametersJSON | DecideRequestItemGroupParametersJSON)[]
): Promise<RelationshipDTO> {
    const templateContent: RelationshipTemplateContentJSON = {
        "@type": "RelationshipTemplateContent",
        onNewRelationship: {
            "@type": "Request",
            items: requestItems
        }
    };

    const template = await exchangeTemplate(sRuntimeServices.transport, rRuntimeServices.transport, templateContent);

    await rRuntimeServices.eventBus.waitForEvent(IncomingRequestStatusChangedEvent, (e) => e.data.request.source!.reference === template.id);

    const requestId = (await rRuntimeServices.consumption.incomingRequests.getRequests({ query: { "source.reference": template.id } })).value[0].id;
    await rRuntimeServices.consumption.incomingRequests.accept({ requestId, items: acceptParams });

    await rRuntimeServices.eventBus.waitForEvent(RelationshipChangedEvent);

    const sRelationship = (await syncUntilHasRelationships(sRuntimeServices.transport, 1))[0];
    expect(sRelationship.status).toStrictEqual(RelationshipStatus.Pending);

    await sRuntimeServices.eventBus.waitForRunningEventHandlers();

    return sRelationship;
}
```

what do you think about this. In your case you would call this with 
```ts
establishPendingRelationshipWithRequestFlow(s1, s2, [
{
                        "@type": "CreateAttributeRequestItem",
                        mustBeAccepted: true,
                        attribute: IdentityAttribute.from({
                            value: {
                                "@type": "GivenName",
                                value: "AGivenName"
                            },
                            owner: (await rRuntimeServices.transport.account.getIdentityInfo()).value.address
                        }).toJSON()
                    },
                    {
                        "@type": "CreateAttributeRequestItem",
                        mustBeAccepted: true,
                        attribute: RelationshipAttribute.from({
                            key: "AKey",
                            value: {
                                "@type": "ProprietaryString",
                                value: "AStringValue",
                                title: "ATitle"
                            },
                            owner: CoreAddress.from((await rRuntimeServices.transport.account.getIdentityInfo()).value.address),
                            confidentiality: RelationshipAttributeConfidentiality.Public
                        }).toJSON()
                    },
                    {
                        "@type": "ShareAttributeRequestItem",
                        mustBeAccepted: true,
                        sourceAttributeId: existingRelationshipAttributeForFurtherSharing.id,
                        attribute: existingRelationshipAttributeForFurtherSharing.content
                    }
],
[{ accept: true }, { accept: true }, { accept: true }]
)
```

jkoenig134: I don't see us reusing exactly the same content as we have here - but if you want to have it re-usable you can put it in your test where you use it.

packages/runtime/src/extensibility/facades/consumption/AttributesFacade.ts
```diff
@@ -14,6 +14,8 @@ import {
     DeletePeerSharedAttributeAndNotifyOwnerUseCase,
     DeleteRepositoryAttributeRequest,
     DeleteRepositoryAttributeUseCase,
+    DeleteSharedAttributesForRejectedOrRevokedRelationshipRequest,
+    DeleteSharedAttributesForRejectedOrRevokedRelationshipUseCase,
```
slavistan, line 18: Would you kindly explain your reasoning behind making this a use case instead of automatically deleting the data? The data left behind after a revoked or rejected relationship request feels of little use to me and I'd rather have it removed immediately. Automating this would mean less data trash and one less use case.

Sorry if you guys already though/spoke about this. I was gone most of last week.

jkoenig134: sure! this has to be done by a module, modules are pieces of code that automate things that runtime users **could** do themselves (when not enabling the module). Therefore modules only have access to things that are coming out of the extensibility -> we have to expose the functionality.

As soon as we don't expose this to App / Connector we should be very safe!

packages/runtime/test/modules/RequestModule.test.ts
```diff
@@ -510,3 +514,82 @@ describe("Handling the rejection and the revocation of a Relationship by the Req
         expect((await sRuntimeServices.consumption.attributes.getAttributes({})).value).toHaveLength(1);
     });
 });
+
+async function establishPendingRelationshipWithPredefinedRequestFlow(
+    sRuntimeServices: TestRuntimeServices,
+    rRuntimeServices: TestRuntimeServices,
+    existingRelationshipAttributeForFurtherSharing?: LocalAttributeDTO
```
jkoenig134, line 521: this is always defined, right?

britsta: In our case, we only use the function `establishPendingRelationshipWithPredefinedRequestFlow` with the optional parameter and since we cannot export the function from a test file, it would probably make sense to set the parameter to required, adapt the implementation and only make it optional again if necessary (i.e. if we want to call it in this file without specifying a value for this parameter). Is that what you mean? 🐸

packages/runtime/src/extensibility/facades/consumption/AttributesFacade.ts
slavistan: A different approach: how about we use a private, internal module that the user has no access to and into which we @Inject the attributes controller?

jkoenig134: There is no `the attribute controller` b/c we have to be prepared for multi accounting here. Therefore this is not possible.

jkoenig134: And also not required. I don't see any problems with the current approach. It's a widely used pattern to create usecases for things that are only used in our modules.

packages/runtime/test/modules/RequestModule.test.ts
jkoenig134: resolved in call!

packages/runtime/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@nmshd/runtime",
-    "version": "4.13.2",
+    "version": "4.14.1",
```
jkoenig134, line 3:
```suggestion
    "version": "4.14.0",
```

also `npm i` is missing ;)

jkoenig134 (approved):

jkoenig134:
@slavistan everything ready here?

PR #174, opened 2024-07-01
Feature/ShareAttributeRequestItem with deletionInfo

Milena-Czierlinski:


packages/app-runtime/package.json
jkoenig134, line 1: please also bump the peer-dependency (runtime) here

jkoenig134 (approved):

packages/consumption/src/modules/requests/itemProcessors/shareAttribute/ShareAttributeRequestItemProcessor.ts
```diff
@@ -47,7 +48,11 @@ export class ShareAttributeRequestItemProcessor extends GenericRequestItemProces
             }
 
             if (typeof recipient !== "undefined") {
-                const query: any = { "shareInfo.sourceAttribute": requestItem.sourceAttributeId.toString(), "shareInfo.peer": recipient.toString() };
+                const query: any = {
```
britsta, line 51: This adjustment is good, but we still need a similar adjustment for the query of the shared successors and predecessors directly below. (Like you did in the use case `ShareRepositoryAttribute.ts`.)

packages/consumption/src/modules/requests/itemProcessors/shareAttribute/ShareAttributeRequestItemProcessor.ts
britsta, line 1: - Open question: Should we allow further sharing of a RelationshipAttribute if it has `"ToBeDeleted"` as `deletionStatus`, for example?
- This is a more quibblingly comment and has lower priority than the others, however: Some error messages in this file were written without taking the Attribute deletion into account, because the Attribute deletion feature is relatively new. Therefore, it would be more accurate to change the tense of some error messages here from present perfect to present, for example, write `"The IdentityAttribute with the given sourceAttributeId '${requestItem.sourceAttributeId.toString()}' is already shared with the peer."` instead of `"The IdentityAttribute with the given sourceAttributeId '${requestItem.sourceAttributeId.toString()}' has already been shared with the peer."`. Adjust such sentences here or leave it as it is as you think.

britsta:
Also, change the code for the sharing of RelationshipAttributes to a third party in the ShareAttributeRequestItemProcessor. If a RelationshipAttribute is in a specific `deletionStatus`, it should be possible to share the source RelationshipAttribute again too.

PR #176, opened 2024-07-02
Refactor/error codes

RuthDiG:
There are some little changes and deletions regarding the error codes.


jkoenig134:
as this is a breaking change please
1. remove the version bumps
2. let it point to `release/v5`

packages/consumption/src/modules/common/ValidationResult.ts
```diff
@@ -21,7 +21,13 @@ export abstract class ValidationResult {
 
     public static fromItems(items: ValidationResult[]): ValidationResult {
         return items.some((r) => r.isError())
-            ? ValidationResult.error(new ApplicationError("inheritedFromItem", "Some child items have errors."), items)
+            ? ValidationResult.error(
+                  new ApplicationError(
+                      "error.consumption.validation.inheritedFromItem",
+                      "Some child items have errors. If this error occured during the creation of a request, we recommand to call 'validate' to get more information."
```
jkoenig134, line 27: i think there is no "validate". It's called canCreate

jkoenig134: also I would not talk about "us". I'd suggest `If this error occurred during the creation of a request, call 'canCreate' to get more information.`

RuthDiG: But you have to post /api/v2/Requests/Outgoing/Validate and then canCreate will be used or isn't so?
https://enmeshed.eu/use-case-consumption-check-if-outgoing-request-can-be-created

jkoenig134: true, the connector route is called like that - but we are in the runtime and should stick to runtime terminology

packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -134,7 +134,7 @@ class Notifications {
     }
 
     public cannotSaveSentNotificationFromPeerMessage(): ApplicationError {
-        return new ApplicationError("error.runtime.notifications.cannotSaveSendNotificationFromPeerMessage", "Cannot send Notification from peer message.");
+        return new ApplicationError("error.runtime.notifications.cannotSaveSentNotificationFromPeerMessage", "Cannot send Notification from peer message.");
```
Milena-Czierlinski, line 137: The error message seems misleading to me. If I understand it correctly, the use case SentNotification is meant to save a LocalNotification for a Notification that was sent inside a Message by oneself. If the owner of the Message is not the Identity itself, this error is thrown. In this case, I'd expect an error message like `The Message ${messageId} was received from a peer, but an own Message is expected here to save its Notification content.`.

Besides, also the naming of the SentNotificationUseCase seems not ideal to me, as it doesn't appropriately describe what happens inside of it. However, this is probably outside the scope of this PR.

packages/consumption/src/modules/common/ValidationResult.ts
slavistan: Just in case you keep a similar wording: It's _recomm**e**nd_, not _recomm**a**nd_.

PR #178, opened 2024-07-02
Fix/Adjust attribute deletion feature for rejected relationships

britsta:


jkoenig134:
@Magnus-Kuhn I fixed some things that were introduced by your new validations regarding canCreate of the Incoming- and OutgoingRequestController. Please review that properly and make sure I don't destroyed any ideas you had during implementing the initial validations.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -184,7 +184,12 @@ export class IncomingRequestsController extends ConsumptionBaseController {
 
         const relationship = await this.relationshipResolver.getRelationshipToIdentity(request.peer);
         // It is safe to decide an incoming Request when no Relationship is found as this is the case when the Request origins from onNewRelationship of the RelationshipTemplateContent
-        if (relationship && relationship.status !== RelationshipStatus.Active) {
+        const possibleStatuses =
+            request.source?.type === "RelationshipTemplate"
+                ? [RelationshipStatus.Active, RelationshipStatus.Rejected, RelationshipStatus.Revoked, RelationshipStatus.Terminated]
+                : [RelationshipStatus.Active];
+
```
Magnus-Kuhn, line 191: why should deciding relationship template requests still be allowed if the relationship is inactive?

jkoenig134: I think terminated could be wrong here. But rejected and revoked are like no relationship exists.

Magnus-Kuhn: Agreed that only Terminated is wrong. Is there a way (planned) to clean up rejected/revoked relationships?

jkoenig134: I mean, we could use your decompose logic for that. But please let US investigate that further before implementing sth like that.

jkoenig134: Alright, removed Terminated, should be fine now.

jkoenig134 (approved):
LGTM!

PR #180, opened 2024-07-03
chore/upgrade dependencies

sebbi08:


jkoenig134:
you have to replace no-if with no-conditionals

.vscode/settings.json
```diff
@@ -54,5 +54,6 @@
     "files.eol": "\n",
     "typescript.unstable": {
         "organizeImportsIgnoreCase": true
-    }
+    },
+    "typescript.tsdk": "node_modules/typescript/lib"
```
jkoenig134, line 58: Why is this required?

sebbi08: VS Code is still shipped with ts 5.4 and this will use the locale ts version.

jkoenig134: according vscode docs it should ship with latest ts, but maybe it will change to 5.5 in the next version..

jkoenig134 (approved):

.ci/runChecks.sh
```diff
@@ -2,7 +2,7 @@ set -e
 
 npm ci
 npm run build:node
-npm run lint:eslint
+npm run lint:eslint -- --rule 'jest/no-conditional-in-test: off'
```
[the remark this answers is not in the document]
jkoenig134: if not ppl will take much time fixing things that doesn't have to be fixed

jkoenig134 (approved):

PR #184, opened 2024-07-05
chore: change default encryption algorithm

stnmtz:


jkoenig134:
not possible in the foreseeable future

PR #185, opened 2024-07-05
Feature/Default RepositoryAttributes

Milena-Czierlinski:


packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -212,6 +213,33 @@ describe("get attribute(s)", () => {
             expect(attributeIds).toContain(relationshipAttributeId);
             expect(attributeIds).toStrictEqual(expect.arrayContaining(identityAttributeIds));
         });
+
+        test("should allow to get only defaul attributes", async function () {
```
RuthDiG, line 217: default ...

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -452,6 +480,22 @@ describe("get repository, own shared and peer shared attributes", () => {
             const repositoryAttributes = result.value;
             expect(repositoryAttributes).toStrictEqual([services1RepoSurnameV0, services1RepoSurnameV1, services1RepoGivenNameV0, services1RepoGivenNameV1]);
         });
+
+        test("should allow to get only defaul attributes", async function () {
```
RuthDiG, line 484: same: default

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -1258,6 +1331,40 @@ describe(SucceedRelationshipAttributeAndNotifyPeerUseCase.name, () => {
...
+        const result = await services1.consumption.attributes.changeDefaultRepositoryAttribute({ attributeId: desiredDefaultAttribute.id });
+        expect(result.isSuccess).toBe(true);
+        const newDefaultAttribute = result.value;
+        expect(newDefaultAttribute.default).toBe(true);
+
+        const updatedFormerDefaultAttribute = (await services1.consumption.attributes.getAttribute({ id: defaultAttribute.id })).value;
+        expect(updatedFormerDefaultAttribute.default).toBeUndefined();
+    });
+});
+
```
RuthDiG, line 1367: Perhaps you could combine these two test in  one and if you want add a test, that if someone deletes the default attribute one of the remaining attributes of the same type will be the default.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -276,6 +278,20 @@ export class AttributesController extends ConsumptionBaseController {
         }
     }
 
+    private async setAsDefaultAttributeIfNoOtherExists(attribute: LocalAttribute): Promise<LocalAttribute> {
```
slavistan, line 281: I find the name a little ambiguous: if no other _what_ exists? Default attribute for the value type or attribute of same value type?

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1042,6 +1074,46 @@ export class AttributesController extends ConsumptionBaseController {
         }
     }
 
+    private async handleDefault(attribute: LocalAttribute): Promise<void> {
```
slavistan, line 1077: If find the method's name a little vague ("handle" doesn't suggest to me what's really going on here). How about something like `setAlternativeDefault`?

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -177,6 +196,10 @@ export class LocalAttribute extends CoreSynchronizable implements ILocalAttribut
         return !!this.shareInfo;
     }
 
+    public isNotDefault(): this is LocalAttribute & { default: undefined } {
```
slavistan, line 199: Be positive :-) In general, give boolean things affirmative names such as `isDefault` or `isSet` as opposed to `isNotDefault` or `isNotSet`. That's not a hard rule by any means but more of an unwritten convention in programming.

https://softwareengineering.stackexchange.com/questions/196830

packages/runtime/src/useCases/consumption/attributes/GetAttributes.ts
```diff
@@ -19,6 +19,9 @@ export interface GetAttributesRequest {
 export interface GetAttributesRequestQuery {
     createdAt?: string;
     parentId?: string;
+    succeeds?: string | string[];
+    succeededBy?: string | string[];
+    default?: string;
```
slavistan, line 24: boolean, no?

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -212,6 +213,33 @@ describe("get attribute(s)", () => {
...
+            const attributeIds = attributes.map((attr) => attr.id);
+            expect(attributeIds).toContain(identityAttributeIds[0]);
+            expect(attributeIds).toContain(identityAttributeIds[1]);
+            expect(attributeIds).not.toContain(relationshipAttributeId);
+        });
+
+        test("should allow not to get defaul attributes", async function () {
+            const result = await services1.consumption.attributes.getAttributes({
+                query: { default: "!true" }
```
slavistan, line 234: Shouldn't this be `{ default: { not: true } }`? According to https://github.com/js-soft/ts-documentdb-access/tree/main/packages/querytranslator at least.

packages/runtime/test/consumption/attributes.test.ts
Milena-Czierlinski: Which two tests do you mean?

Regarding adding a test that checks if the default is transmitted correctly, I already do so in consumption. I think this should principally be sufficient.

packages/runtime/test/consumption/attributes.test.ts
jkoenig134: image
@Milena-Czierlinski your query is totally fine.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1049,6 +1081,46 @@ export class AttributesController extends ConsumptionBaseController {
...
+        const query: IIdentityAttributeQuery = {
+            valueType: valueType
+        };
+
+        const attributesWithSameValueType = await this.executeIdentityAttributeQuery(query);
+        if (attributesWithSameValueType.length <= 1) return;
+
+        const defaultCandidate = attributesWithSameValueType.find((attr) => !attr.succeededBy && attr.id.toString() !== attribute.id.toString());
+
+        if (!defaultCandidate) return;
```
Milena-Czierlinski, line 1097: `defaultCandidate` should never be undefined, if we have valid data. So I am wondering, if we should throw an error here instead. However, if under any circumstances this case should be reached, the user couldn't delete their Attribute and also couldn't do anything about it.

Somehow I thought I already started this discussion, but that doesn't appear to be the case...

packages/runtime/src/useCases/consumption/attributes/ChangeDefaultRepositoryAttribute.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    protected async executeInternal(request: ChangeDefaultRepositoryAttributeRequest): Promise<Result<LocalAttributeDTO>> {
+        const repositoryAttribute = await this.attributesController.getLocalAttribute(CoreId.from(request.attributeId));
+        if (!repositoryAttribute) return Result.fail(RuntimeErrors.general.recordNotFound(LocalAttribute));
+
+        if (!repositoryAttribute.isRepositoryAttribute(this.accountController.identity.address)) {
+            return Result.fail(RuntimeErrors.attributes.isNotRepositoryAttribute(CoreId.from(request.attributeId)));
+        }
+
+        const defaultRepositoryAttribute = await this.attributesController.changeDefaultRepositoryAttribute(repositoryAttribute);
```
Milena-Czierlinski, line 36: Should we allow the user to manually set a RepositoryAttribute as default that already has a successor? Maybe this wouldn't be great for UX.

packages/consumption/src/modules/attributes/AttributesController.ts
Magnus-Kuhn: From what I've seen here so far, I would expect an error without error code to let us know that our coding is faulty (would actually be like in the changeDefaultRepositoryAttribute error)

packages/consumption/src/modules/attributes/AttributesController.ts
Magnus-Kuhn: suggestion: `setAsDefaultAttributeIfTypeUnique`

packages/consumption/src/modules/attributes/AttributesController.ts
RuthDiG: What about `overwriteDefault`?

Milena-Czierlinski: This poses the same problem: The default is only overwritten, in case the Attribute I delete was the former default. So, if in the function `executeFullAttributeDeletionProcess` a function called `overwriteDefault` is called, this would cause confusion. In my opinion, `transmitDefault` has the advantage that you can figure, that something can only be transmitted, if it was there before.

packages/runtime/src/useCases/consumption/attributes/GetRepositoryAttributes.ts
```diff
@@ -15,6 +15,7 @@ export interface GetRepositoryAttributesRequest {
 
 export interface GetRepositoryAttributesRequestQuery {
     createdAt?: string;
+    default?: string;
```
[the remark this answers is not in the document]
Milena-Czierlinski: @slavistan Already commented on this above and I explained there why it doesn't work, if it is of type boolean.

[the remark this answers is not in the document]
Magnus-Kuhn: My idea was to then transform the boolean into the string as a first step - the way it works right now seems impossible to guess

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -135,6 +148,8 @@ export class LocalAttribute extends CoreSynchronizable implements ILocalAttribut
     public isOwnSharedAttribute(ownAddress: CoreAddress, peerAddress?: CoreAddress): this is OwnSharedIdentityAttribute | OwnSharedRelationshipAttribute {
         let isOwnSharedAttribute = this.isShared() && this.isOwnedBy(ownAddress);
 
+        isOwnSharedAttribute &&= !this.isDefault();
```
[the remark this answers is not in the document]
Magnus-Kuhn: sorry, I should have been clearer - I just considered this check to be covered by the isShared check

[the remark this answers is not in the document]
Milena-Czierlinski: But isShared doesn't check the default 🤔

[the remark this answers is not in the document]
Magnus-Kuhn: shared attributes aren't going to have the default set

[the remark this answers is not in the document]
Milena-Czierlinski: Exactly, but this isn't explicitly checked before. This is why I added this line. :)

packages/runtime/test/consumption/attributes.test.ts
RuthDiG: "should create a RepositoryAttribute that is the default if it is the first of its value type"
and 
"should create a RepositoryAttribute that is not the default if it is not the first of its value type" these two tests.

packages/consumption/src/modules/attributes/AttributesController.ts
Milena-Czierlinski: Personally, I prefer `setAsDefaultAttributeIfNoOtherExists` over `setAsDefaultAttributeIfTypeUnique` since it is easier language. @slavistan Any opinion?

packages/runtime/src/useCases/consumption/attributes/GetRepositoryAttributes.ts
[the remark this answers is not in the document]
Milena-Czierlinski: @jkoenig134  Would that make sense?

packages/runtime/test/consumption/attributes.test.ts
Milena-Czierlinski: It is good practice to check one thing per test. Thus, I'd like to keep them separated.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1049,6 +1081,46 @@ export class AttributesController extends ConsumptionBaseController {
...
+    private async transmitDefault(attribute: LocalAttribute): Promise<void> {
+        if (!attribute.default) return;
+
+        const valueType = attribute.content.value.constructor.name as AttributeValues.Identity.TypeName;
+        const query: IIdentityAttributeQuery = {
+            valueType: valueType
+        };
+
+        const attributesWithSameValueType = await this.executeIdentityAttributeQuery(query);
+        if (attributesWithSameValueType.length <= 1) return;
```
britsta, line 1093: I think it would be better to adjust the `query` used (for the case of Attribute succession). If you previously had a RepositoryAttribute with predecessor, `attributesWithSameValueType.length` is automatically greater than one, right? (Even though there is no RepositoryAttribute left as an appropriate default RepositoryAttribute.)

packages/consumption/src/modules/attributes/AttributesController.ts
britsta: If we throw an error, we would probably have to adjust the `query` above so that `defaultCandidate` really can't be `undefined`. See my other comment on this.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -611,6 +655,35 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
         await expect(services1.eventBus).toHavePublished(AttributeCreatedEvent, (e) => e.data.content.value["@type"] === "City");
         await expect(services1.eventBus).toHavePublished(AttributeCreatedEvent, (e) => e.data.content.value["@type"] === "Country");
     });
+
+    test("should create a RepositoryAttribute that is the default if it is the first of its value type", async () => {
+        const request: CreateRepositoryAttributeRequest = {
+            content: {
+                value: {
+                    "@type": "Pseudonym",
+                    value: "Pseudo Petra"
```
britsta, line 664: This makes me laugh a lot :D But please use a more descriptive value here, for example "A pseudonym".

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -611,6 +655,35 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
...
+        const attribute = result.value;
+        expect(attribute.default).toBe(true);
+    });
+
+    test("should create a RepositoryAttribute that is not the default if it is not the first of its value type", async () => {
+        const request: CreateRepositoryAttributeRequest = {
+            content: {
+                value: {
+                    "@type": "JobTitle",
+                    value: "Boss"
```
britsta, line 678: Please use a more descriptive value here, for example "A job title".

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -1258,6 +1331,40 @@ describe(SucceedRelationshipAttributeAndNotifyPeerUseCase.name, () => {
     });
 });
 
+describe(ChangeDefaultRepositoryAttributeUseCase.name, () => {
+    test("should change default repository attribute", async () => {
```
britsta, line 1335: Please write "RepositoryAttribute" instead of "repository attribute" here as you do in the tests above.

packages/runtime/src/useCases/consumption/attributes/GetRepositoryAttributes.ts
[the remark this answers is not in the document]
jkoenig134: nope, query values are strings by definition to be compatible with the connector. It's alright as it is.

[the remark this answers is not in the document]
Magnus-Kuhn: I was proposing a breaking change to the connector (booleans can be put in JSONs), but there's apparently no point in pursuing this any further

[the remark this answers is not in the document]
jkoenig134: eeehm, that would mean that all GET would become posts. No - we won't do that.

packages/consumption/src/modules/attributes/AttributesController.ts
Milena-Czierlinski: Since we delete the predecessors before transmitting the default, this shouldn't pose a problem. I added a test for it.

packages/runtime/src/useCases/consumption/attributes/ChangeDefaultRepositoryAttribute.ts
Milena-Czierlinski: Currently, the App doesn't show the predecessors at all. Thus, for now I'll prohibit to set outdated versions as default. If at a later point in time we'll expose the predecessors, we can think about adjusting the validation here, as well.

packages/consumption/src/modules/attributes/AttributesController.ts
britsta: Good point. However, would still adapt the `query` to increase the reusability of the function and reduce the error-proneness of the function, so that the functionality of the function is not so strongly influenced by its position within a function call.

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -33,36 +34,43 @@ export interface ILocalAttribute extends ICoreSynchronizable {
...
 export type ThirdPartyOwnedRelationshipAttribute = LocalAttribute & {
     content: RelationshipAttribute;
     shareInfo: LocalAttributeShareInfo;
+    default: undefined;
 };
 
 export type RepositoryAttribute = LocalAttribute & {
     content: IdentityAttribute;
     shareInfo: undefined;
+    deletionInfo: undefined;
```
slavistan, line 73: Just making sure we're on the same page: This is set because there's no multi-step process to delete a repository attribute since we just delete them, right?

Milena-Czierlinski: Exactly!

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1049,6 +1081,62 @@ export class AttributesController extends ConsumptionBaseController {
         }
     }
 
+    private async transmitDefault(attribute: LocalAttribute): Promise<void> {
```
slavistan, line 1084: 1. If find the name confusing as 'transmit' means to send something over the wire. Did you meant to call it 'transfer'? How about something simple as `setOtherAttributeAsDefault`?
2. Your other `...default` methods return the new default attribute, while this one returns void. I understand that you don't need the result anywhere but I find it easier to understand when functions which do similar things also have similar signatures.

packages/consumption/src/modules/attributes/AttributesController.ts
slavistan: >We always have a default attribute for each value type

Right. Keep the name then or use the other one, I don't mind either. But I do have a different suggestion: Have you thought about merging `changeDefaultRepositoryAttribute` and `setAsDefault..........` into a single method? They do similar things and could be implemented with a single `setAsDefaultRepositoryAttribute(attribute: LocalAttribute, override?: boolean)` where `override` would be a (false-if-undefined) boolean used to replace an existing default.

packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -189,6 +190,13 @@ class Attributes {
             `Attribute '${attributeId.toString()}' is not a third party owned relationship attribute.`
         );
     }
+
+    public hasSuccessor(predecessor: LocalAttribute): ApplicationError {
```
britsta, line 194: Why didn't we already have to use such an error for the `SucceedRepositoryAttributeUseCase`? How is the case of trying to succeed a predecessor RepositoryAttribute (which already has a successor) handled and what error is thrown?

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -212,6 +213,33 @@ describe("get attribute(s)", () => {
             expect(attributeIds).toContain(relationshipAttributeId);
             expect(attributeIds).toStrictEqual(expect.arrayContaining(identityAttributeIds));
         });
+
+        test("should allow to get only default attributes", async function () {
```
britsta, line 217: It would be cool if another RepositoryAttribute of type `Surname`, for example, would occur in that test, so that it is demonstrated that this RepositoryAttribute is not a default RepositoryAttribute as the other RepositoryAttribute of type `Surname` already is.

This RepositoryAttribute is then found in the test below as a RepositoryAttribute which is not a default Attribute.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -452,6 +480,22 @@ describe("get repository, own shared and peer shared attributes", () => {
             const repositoryAttributes = result.value;
             expect(repositoryAttributes).toStrictEqual([services1RepoSurnameV0, services1RepoSurnameV1, services1RepoGivenNameV0, services1RepoGivenNameV1]);
         });
+
+        test("should allow to get only default attributes", async function () {
```
britsta, line 484: Why is this test duplicated for the `GetAttributes` function, but the test `"should allow not to get default attributes"` is done for the `GetAttributes` function, but not for the `GetRepositoryAttributes` function?

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -1258,6 +1331,108 @@ describe(SucceedRelationshipAttributeAndNotifyPeerUseCase.name, () => {
...
+                content: {
+                    value: {
+                        "@type": "GivenName",
+                        value: "My new default name"
+                    }
+                }
+            })
+        ).value;
+
+        const result = await services1.consumption.attributes.changeDefaultRepositoryAttribute({ attributeId: desiredDefaultAttribute.id });
```
britsta, line 1358: Before the default RepositoryAttribute is changed, add expect statements that the `defaultAttribute` has currently `default: true` and the `desiredDefaultAttribute` has currently `default: undefined`.

packages/consumption/src/modules/attributes/AttributesController.ts
Milena-Czierlinski: In principle, that is a good idea, since as you pointed out they do similar things. However, I worry that having this as a public function then, somebody might mistakenly forget to set the overwrite flag and we'd end up with multiple default Attributes. Thus, I think it might be safer to keep them as two separate functions.

packages/runtime/src/useCases/common/RuntimeErrors.ts
Milena-Czierlinski: In `SucceedRepositoryAttributeUseCase` the validation from the AttributesController is called and in case of an error the consumption error is returned.

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -711,6 +740,304 @@ describe("AttributesController", function () {
...
+                    value: EMailAddress.from({
+                        value: "my2@email.address"
+                    }),
+                    owner: consumptionController.accountController.identity.address
+                })
+            };
+            const defaultRepositoryAttribute = await consumptionController.attributes.createRepositoryAttribute(attributeParams);
+            const otherRepositoryAttribute = await consumptionController.attributes.createRepositoryAttribute(attributeParams2);
+
+            await consumptionController.attributes.executeFullAttributeDeletionProcess(defaultRepositoryAttribute);
```
britsta, line 1010: Add a check that `defaultRepositoryAttribute` has `default: true` and that `otherRepositoryAttribute` has `default: undefined` before you delete `defaultRepositoryAttribute`.

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -1247,6 +1574,31 @@ describe("AttributesController", function () {
...
+                    content: IdentityAttribute.from({
+                        value: {
+                            "@type": "Nationality",
+                            value: "US"
+                        },
+                        owner: consumptionController.accountController.identity.address
+                    })
+                };
+
+                const { predecessor: updatedPredecessor, successor } = await consumptionController.attributes.succeedRepositoryAttribute(predecessor.id, successorParams);
```
britsta, line 1597: Before you succeed the `predecessor`, add an expect statement that it has `default: true`.

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -2035,18 +2432,18 @@ describe("AttributesController", function () {
             expect(result0).toStrictEqual([]);
 
             const result1 = await consumptionController.attributes.getPredecessorsOfAttribute(repositoryAttributeVersion1.id);
-            expect(result1).toStrictEqual([repositoryAttributeVersion0]);
+            expect(JSON.stringify(result1)).toStrictEqual(JSON.stringify([repositoryAttributeVersion0]));
```
britsta, line 2435: Why was it necessary to add these `JSON.stringify(...)` calls?

packages/runtime/test/consumption/attributes.test.ts
Milena-Czierlinski: Testing the negative is not really something that is connected to this PR, but to the query itself. I was wondering how it would work so I added the test above. But I don't think it is necessary to add such a test everywhere a query is used (as can be seen that there wasn't such a test before).

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1049,6 +1081,62 @@ export class AttributesController extends ConsumptionBaseController {
...
+        };
+
+        const defaultCandidates = await this.getLocalAttributes(query);
+        if (defaultCandidates.length === 0) return;
+
+        defaultCandidates[0].default = true;
+        await this.updateAttributeUnsafe(defaultCandidates[0]);
+    }
+
+    public async changeDefaultRepositoryAttribute(newDefaultAttribute: LocalAttribute): Promise<LocalAttribute> {
```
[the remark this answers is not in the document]
britsta: The `ChangeDefaultRepositoryAttributeUseCase` which calls this `changeDefaultRepositoryAttribute` function checks if `newDefaultAttribute` is a RepositoryAttribute. Do we want to replicate this check or do we want to transfer it from the use case to the function within the `AttributesController`?

And just out of interest: Public methods of the Controllers (within consumption, for example) can't be used by integrators but can be used by us? So your motivation of adding a check for the `changeDefaultRepositoryAttribute` function is, for example, that it causes _our_ coding to be less error-prone?

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -23,6 +23,7 @@ export interface LocalAttributeJSON {
     shareInfo?: LocalAttributeShareInfoJSON;
     deletionInfo?: LocalAttributeDeletionInfoJSON;
     parentId?: string;
+    default?: boolean;
```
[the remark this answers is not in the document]
britsta: Not sure either, question out of interest. 🐸

packages/runtime/src/useCases/consumption/attributes/GetAttributes.ts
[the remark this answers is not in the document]
britsta: Same here, with the `default` just the same as with the `isTechnical` property. Do you know more about this @jkoenig134?

packages/consumption/src/modules/attributes/AttributesController.ts
Milena-Czierlinski: 1. Yes, 'transfer' seems to describe it more appropriately. Regarding `setOtherAttributeAsDefault` I have still the same concerns as already discussed above. If the Attribute I delete wasn't the former default Attribute, we won't set another Attribute as default. This is a point I want to express with `transfer`.
2. I see where you are coming from, but I'm having trouble applying this here. In the other `...default` methods you have the same (adjusted) Attribute as input and output. However, here we an Attribute that could be of any kind. If it wasn't the former Attribute, we would simply return it unchanged. If it was and we transfer the default, would we then return the former default Attribute, which is not a default Attribute anymore? Or would we return the new default Attribute, which is contrary to the case, that the input Attribute was no a default Attribute. Thus, I think the solution with least complexity is to not return anything at all.

packages/runtime/src/useCases/consumption/attributes/GetAttributes.ts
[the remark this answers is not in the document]
jkoenig134: for bools it doesn't make sense to allow arrays as it does not make sense to ask for `["true", "false"]`, you can just skip that.

packages/consumption/src/modules/attributes/AttributesController.ts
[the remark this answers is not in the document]
Milena-Czierlinski: Yes, integrators can only use use cases. But we can internally use public controller functions, e.g. in RequestItemProcessors.

packages/runtime/src/useCases/consumption/attributes/GetRepositoryAttributes.ts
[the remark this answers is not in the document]
jkoenig134: isTechnical should also be string

packages/consumption/test/modules/attributes/AttributesController.test.ts
Milena-Czierlinski: Yeah, it's kind of weird behavior. So, if I don't add the `JSON.stringify(...)` calls, I'll get an error like the following: 

image

I think it is because for one side of the comparison `default` is explicitly set to `undefined`, since it was true before, but for the other it isn't listed explicitly, since it is undefined. I found this as a work around.

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -711,6 +740,304 @@ describe("AttributesController", function () {
...
+                    shareInfo: {
+                        peer: CoreAddress.from("peer"),
+                        requestReference: CoreId.from("reqRef2"),
+                        sourceAttribute: successorRepositoryAttribute.id
+                    }
+                };
+                await consumptionController.attributes.succeedOwnSharedIdentityAttribute(predecessorSharedCopy.id, sharedSuccessorParams);
+
+                const result = await consumptionController.attributes.validateFullAttributeDeletionProcess(successorRepositoryAttribute);
+                expect(result.isSuccess()).toBe(true);
```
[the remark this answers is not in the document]
Milena-Czierlinski: Note that this test only performs a validation and doesn't in fact delete or detach anything.

However, I adjusted this block to remove redundant code in the following tests.

packages/consumption/src/modules/attributes/AttributesController.ts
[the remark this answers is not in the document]
Milena-Czierlinski: I now duplicated the check, such that if the function is called from the runtime use case, a runtime error will be thrown instead of the consumption error.

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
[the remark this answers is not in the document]
jkoenig134: yep, would be great :)

packages/runtime/src/useCases/consumption/attributes/GetAttributes.ts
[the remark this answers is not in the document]
Milena-Czierlinski: See #203

packages/consumption/src/modules/attributes/AttributesController.ts
slavistan: If you want to override defaults then change the second parameter to `skipOverride?: boolean`. Does that solve the problem?

Milena-Czierlinski:
The problem with the property name `default` is that it is a key word. Even though Typescript allows us to use it as a property name, Dart does not. Hence, if we called it that way, we would need to rename it in the App-Runtime bringe. However, since we added this feature especially for the App, it is easier if we choose a consistent naming. Thus, I renamed it to `isDefault`.

slavistan (dismissed):

slavistan (approved):

PR #183, opened 2024-07-03
Refactor/Remove mustBeAccepted of RequestItemGroups

britsta:
I will also remove the `mustBeAccepted` property of the RequestItemGroup within the Connector, the App Runtime bridge 🐦‍⬛🌉 and the documentation 📝.

PR #186, opened 2024-07-08
chore/properly update lockfile

jkoenig134:
this was missed in #183

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12538,7 +12538,7 @@
         },
         "packages/app-runtime": {
             "name": "@nmshd/app-runtime",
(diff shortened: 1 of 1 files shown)


britsta:
> this was missed in #183

Do you know why the version bumping of transport and app-runtime was not updated correctly in the `package-lock.json` file, although version bumping was also carried out for transport and app-runtime in https://github.com/nmshd/runtime/pull/183, and how I can ensure that the changes from the `package.json` files of the different packages are correctly transferred to the `package-lock.json` file in the future?

britsta (approved):

PR #189, opened 2024-07-09
chore/update test tsconfigs

jkoenig134:
the ts-server just exploded in those test folders, was like we haven't any typescript and we could write gibberish code. This fixes the issues - but don't ask me why!


jkoenig134:
seems to be related with ts 5.5, but I really don't know why this explodes this hard.

sebbi08 (approved):
For me, it is still working with that setting had no problems earlier today.

From the documentation I can see that this option has something to do with the build process and as we do not really build the test I think we should be fine.

jkoenig134:
@sebbi08 it's a bit hard to find

PR #191, opened 2024-07-11
Refactor/remove sync callbacks

jkoenig134:
Sync Callbacks are a relic that were only made for the app. The Flutter App will no be able to consume this so we can remove it.


britsta (commented):
Was it intentional that no version bumping has been done in this PR so far?

jkoenig134:
yep, don't plan to publish this individually

britsta (approved):

PR #192, opened 2024-07-11
Prohibit sharing RelationshipAttributes of pending Relationships

britsta:
In this PR, the added validation prevents a RelationshipAttribute, which was created as part of a pending Relationship by accepting a Request, from being shared with a third party before the Relationship is activated. The following module and the following processors are affected by this validation:
- AttributeListenerModule
- ReadAttributeRequestItemProcessor
- ShareAttributeRequestItemProcessor

__Detail__: Please note that when accepting a Request contained in a RelationshipTemplate, the RelationshipAttributes  are created first, and only then is the pending Relationship created. For this reason, it is not enough within the coding to simply check whether there is currently a pending Relationship and only then to prevent the RelationshipAttribute from being shared further. However, it may be that certain checks can or must be adjusted again as soon as a pending Relationship is prevented from being created if both peers have not yet been decomposed. 

**Open question:** It must also be considered whether RelationshipAttributes of terminated but not yet mutual decomposed Relationships should still be able to be shared with third parties.


packages/consumption/test/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.test.ts
```diff
@@ -973,15 +1061,14 @@ describe("ReadAttributeRequestItemProcessor", function () {
 
             test("can be called with an arbitrary third party if the thirdParty string array of the ThirdPartyRelationshipAttributeQuery contains an empty string", async function () {
                 const sender = CoreAddress.from("Sender");
-                const aThirdParty = CoreAddress.from("AThirdParty");
-                const anUninvolvedThirdParty = CoreAddress.from("AnUninvolvedThirdParty");
+                const anInvolvedThirdParty = CoreAddress.from("AnInvolvedThirdParty");
```
Magnus-Kuhn, line 1064: Could be my fault fornot knowing what "involved" means in this context, but I was fairly confused by this naming where the contrast between `aThirdParty`and `anUninvolvedThirdParty` was lost

packages/runtime/src/modules/AttributeListenerModule.ts
```diff
@@ -19,20 +21,69 @@ export class AttributeListenerModule extends RuntimeModule {
...
+        const lastAuditLogEntry = changedRelationship.auditLog[changedRelationship.auditLog.length - 1];
+        if (lastAuditLogEntry.reason === RelationshipAuditLogEntryReason.AcceptanceOfCreation) {
+            const services = await this.runtime.getServices(event.eventTargetAddress);
+            const relationshipAttributesWithPeer = (
+                await services.consumptionServices.attributes.getAttributes({ query: { "content.@type": "RelationshipAttribute", "shareInfo.peer": changedRelationship.peer } })
+            ).value;
+
+            for (const relationshipAttribute of relationshipAttributesWithPeer) {
+                if (relationshipAttribute.content["@type"] !== "RelationshipAttribute") return;
+                if (relationshipAttribute.content.confidentiality === RelationshipAttributeConfidentiality.Private) return;
```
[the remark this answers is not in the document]
britsta: Removing the type check unfortunately leads to the `confidentiality` of the `relationshipAttribute.content` being underlined in red because it is allegedly not guaranteed that the `confidentiality` property is set for `relationshipAttribute.content`, although the query was previously specified so that `relationshipAttribute.content` is guaranteed to be a RelationshipAttribute... 😿😂 

I'd probably leave the confidentiality check where it is to comply with the above pattern.

packages/consumption/test/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.test.ts
britsta: I can understand that, that was also my concern. What do you think about `aQueriedThirdParty` instead of `anInvolvedThirdParty` as an alternative naming?

Magnus-Kuhn: I like that

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -138,6 +138,37 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+                        },
+                        {
+                            ["status"]: { $eq: RelationshipStatus.DeletionProposed }
+                        }
+                    ]
+                };
+
+                const activatedOrTerminatedRelationshipToPeer = await this.accountController.relationships.getRelationships(queryForActivatedOrTerminatedRelationship);
+
+                if (pendingRelationshipToPeer.length !== 0 || activatedOrTerminatedRelationshipToPeer.length === 0) {
```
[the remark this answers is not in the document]
britsta: Do you mean that in the case that both a pending Relationship and a deletionProposed Relationship exist, the sharing of RelationshipAttributes is prevented, that actually belong to the deletionProposed Relationship and not to the pending Relationship and should therefore be able to be shared? 

If you are referring to the name of the variable `activatedOrTerminatedRelationshipToPeer`, I have interpreted deletion proposed Relationships as terminated Relationships (see coding) so as not to extend the variable name any further. But I can also change this if you have a better suggestion.

I also thought about adding a comment here, but since it does not comply with our code guidelines, I decided against it. I will think again about whether it wouldn't be better to add a comment here temporary. As soon as the change you mentioned is made to release/v5, the check ‘IF (activatedOrTerminatedRelationshipToPeer.length === 0)’ (currently also includes deletionProposed Relationships despite the naming) should be sufficient in terms of implementation and also understandable enough without comment.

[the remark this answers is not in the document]
Magnus-Kuhn: > Do you mean that in the case that both a pending Relationship and a deletionProposed Relationship exist, the sharing of RelationshipAttributes is prevented, that actually belong to the deletionProposed Relationship and not to the pending Relationship and should therefore be able to be shared?

This or the case you mentioned that the pending relationship has not been created yet which would make the validation successful.

To the comment: My only point was that you have to know about the problem with the order of creating the request and the relationship to understand the second check

packages/consumption/src/consumption/CoreErrors.ts
```diff
@@ -283,6 +283,10 @@ class Requests {
         return new CoreError("error.consumption.requests.attributeQueryMismatch", message);
     }
 
+    public cannotShareRelationshipAttributeOfPendingRelationship(message: string) {
+        return new CoreError("error.consumption.requests.cannotShareRelationshipAttributeOfPendingRelationship", message);
```
Milena-Czierlinski, line 287: Why don't you use a static error message?

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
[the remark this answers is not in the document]
Milena-Czierlinski: I agree. I find the way it is right now confusing.

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -138,6 +138,37 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
                     );
                 }
+
+                const queryForPendingRelationship: any = {
+                    "peer.address": foundLocalAttribute.shareInfo.peer.address,
+                    status: RelationshipStatus.Pending
+                };
+                const pendingRelationshipToPeer = await this.accountController.relationships.getRelationships(queryForPendingRelationship);
+
+                const queryForActivatedOrTerminatedRelationship: any = {
```
Milena-Czierlinski, line 148: If you are to keep this, think about replacing "activated" by "active". I think the emphasize should be on the fact that the Relationship is active now and not on the fact that it was activated in the past.

Milena-Czierlinski (commented):
Why isn't it necessary to adjust the ProposeAttributeRequestItemProcessor, too?

britsta:
> Why isn't it necessary to adjust the ProposeAttributeRequestItemProcessor, too?
ThirdPartyRelationshipAttributes no longer play a role in the processing of ProposeAttributeRequestItems as of release/v5, see also the updated documentation https://github.com/nmshd/documentation/pull/229/files#diff-68fc5ca7e063208ef0fd50053e3262f501cce9c0e8ca8c555e38b01afd637c46.

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -138,6 +138,33 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+                        {
+                            ["status"]: { $eq: RelationshipStatus.Terminated }
+                        },
+                        {
+                            ["status"]: { $eq: RelationshipStatus.DeletionProposed }
+                        }
+                    ]
+                };
+
+                const activeOrTerminatedRelationshipToPeer = await this.accountController.relationships.getRelationships(queryForActiveOrTerminatedRelationship);
```
jkoenig134, line 163:
```suggestion
                const activeOrTerminatedRelationshipsToPeer = await this.accountController.relationships.getRelationships(queryForActiveOrTerminatedRelationship);
```

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -138,6 +138,33 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+                        {
+                            ["status"]: { $eq: RelationshipStatus.Active }
+                        },
+                        {
+                            ["status"]: { $eq: RelationshipStatus.Terminated }
+                        },
+                        {
+                            ["status"]: { $eq: RelationshipStatus.DeletionProposed }
+                        }
+                    ]
```
jkoenig134, line 160: $in operator

packages/consumption/src/modules/requests/itemProcessors/shareAttribute/ShareAttributeRequestItemProcessor.ts
```diff
@@ -115,6 +115,33 @@ export class ShareAttributeRequestItemProcessor extends GenericRequestItemProces
...
+                    {
+                        ["status"]: { $eq: RelationshipStatus.Active }
+                    },
+                    {
+                        ["status"]: { $eq: RelationshipStatus.Terminated }
+                    },
+                    {
+                        ["status"]: { $eq: RelationshipStatus.DeletionProposed }
+                    }
+                ]
```
jkoenig134, line 137: $in operator

packages/consumption/src/modules/requests/itemProcessors/shareAttribute/ShareAttributeRequestItemProcessor.ts
```diff
@@ -115,6 +115,33 @@ export class ShareAttributeRequestItemProcessor extends GenericRequestItemProces
...
+            };
+
+            const activeOrTerminatedRelationshipToPeer = await this.accountController.relationships.getRelationships(queryForActiveOrTerminatedRelationship);
```
jkoenig134, line 140:
```suggestion
            const activeOrTerminatedRelationshipsToPeer = await this.accountController.relationships.getRelationships(queryForActiveOrTerminatedRelationship);
```

packages/runtime/src/modules/AttributeListenerModule.ts
```diff
@@ -19,20 +21,68 @@ export class AttributeListenerModule extends RuntimeModule {
...
+        }
+
+        return false;
+    }
+
+    private async handleRelationshipChanged(event: RelationshipChangedEvent) {
+        const changedRelationship = event.data;
+
+        const lastAuditLogEntry = changedRelationship.auditLog.at(-1);
+        if (lastAuditLogEntry && lastAuditLogEntry.reason === RelationshipAuditLogEntryReason.AcceptanceOfCreation) {
```
jkoenig134, line 59: there is always at minimum one entry

packages/runtime/src/modules/AttributeListenerModule.ts
```diff
@@ -19,20 +21,68 @@ export class AttributeListenerModule extends RuntimeModule {
         const createdAttribute = event.data;
         if (createdAttribute.content["@type"] === "IdentityAttribute" && createdAttribute.shareInfo) return;
         if (createdAttribute.content["@type"] === "RelationshipAttribute" && createdAttribute.content.confidentiality === RelationshipAttributeConfidentiality.Private) return;
+        if (await AttributeListenerModule.relationshipAttributeOfInactiveRelationship(services, createdAttribute)) return;
 
         const getAttributeListenersResult = await services.consumptionServices.attributeListeners.getAttributeListeners();
         if (getAttributeListenersResult.isError) {
-            this.logger.error("Could not get attribute listeners", getAttributeListenersResult.error);
+            this.logger.error("Could not get LocalAttributeListeners", getAttributeListenersResult.error);
```
jkoenig134, line 28: it was better before, the modules consume the runtime and do not know anything about local stuff.

packages/runtime/src/modules/AttributeListenerModule.ts
```diff
@@ -19,20 +21,68 @@ export class AttributeListenerModule extends RuntimeModule {
...
 
         const attributeListeners = getAttributeListenersResult.value;
         const promises = attributeListeners.map((attributeListener) =>
-            this.createRequestIfAttributeMatchesQuery(services, attributeListener, event.data, event.eventTargetAddress)
+            this.createRequestIfAttributeMatchesQuery(services, attributeListener, createdAttribute, event.eventTargetAddress)
         );
         await Promise.all(promises);
     }
 
+    private static async relationshipAttributeOfInactiveRelationship(services: RuntimeServices, attribute: LocalAttributeDTO): Promise<boolean> {
```
jkoenig134, line 39: 1. naming. a function returning an attribute should be named isBlaBlaBla
2. why static?

packages/runtime/test/lib/testUtils.ts
```diff
@@ -428,6 +428,24 @@ export async function ensurePendingRelationship(sTransportServices: TransportSer
     return (await sTransportServices.relationships.getRelationships({})).value[0];
 }
 
+export async function mutualDecompositionIfActiveRelationshipExists(sTransportServices: TransportServices, rTransportServices: TransportServices): Promise<void> {
```
jkoenig134, line 431: test utils should be written as imperatives (for example exchangeAndAcceptRequestByMessage directly below => use sender and recipient to exchange and accept a request by message)

jkoenig134 (changes requested):

packages/runtime/src/modules/AttributeListenerModule.ts
britsta: Do you mean the check could be simplified to `if (lastAuditLogEntry.reason === RelationshipAuditLogEntryReason.AcceptanceOfCreation)`? When removing the check for whether `lastAuditLogEntry` is defined, I receive `'lastAuditLogEntry' is possible 'undefined'.`, even though it is always defined in principle, of course.

packages/runtime/src/modules/AttributeListenerModule.ts
britsta: I have no problem with undoing my changes. Nevertheless, I do not understand exactly why it is not more precise to speak of LocalAttributeListeners from the naming of the return types of the functions called here. An "attribute listener" would then not be a data object type, but just a word from normal usage?

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
britsta: It can be assumed that there is either no `Active` or `Terminated` or `DeletionProposed` Relationship, or at most one. I have therefore decided to use the singular instead of the plural, as the length of the array is at most one if the program runs without bugs. If you like, I can still use the plural form. Or do you have any better ideas for naming this variable that would be more suitable?

packages/runtime/src/modules/AttributeListenerModule.ts
jkoenig134: It could be misleading to talk about Local stuff as local is just an internal wording we don't really expose anywhere.

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
jkoenig134: true, but a variable called activeOrTerminatedRelationshipToPeer should not be an array.

packages/runtime/src/modules/AttributeListenerModule.ts
britsta: 2. I used `static` because I do not need to call a method or an attribute of an instance of the class `AttributeListenerModule` to call `relationshipAttributeOfInactiveRelationship`.
1. I will try to find a new wording that is coherent in combination with the prefix `AttributeListenerModule`, which is caused by the `static` function declaration currently used, as well.

jkoenig134: please don't use static here, It's way less readable on callee side.

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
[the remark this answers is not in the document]
britsta: After the PR https://github.com/nmshd/runtime/pull/227 is merged, a Relationship in status `"Pending"` and a Relationship in status `"DeletionProposed"` can no longer exist at the same time, which is why the implementation can be simplified here and therefore no additional comment is required to explain it anymore.

packages/runtime/src/modules/AttributeListenerModule.ts
jkoenig134: you can use ! in this case

jkoenig134: (sorry, I didn't saw this message)

jkoenig134 (approved):

PR #193, opened 2024-07-12
Chore/use did schema where id2 was used

jkoenig134:


packages/consumption/test/modules/requests/testHelpers/TestObjectFactory.ts
```diff
@@ -60,7 +60,7 @@ export class TestObjectFactory {
                     auditLog: [
                         {
                             createdAt: CoreDate.from("2020-01-01T00:00:00.000Z"),
-                            createdBy: CoreAddress.from("id2"),
+                            createdBy: CoreAddress.from("did:e:a-domain:dids:anidentity2"),
                             createdByDevice: CoreId.from("DVC1"),
```
britsta, line 64: Please adjust the device to `"DVC2"` instead of `"DVC1"` if we are already making adjustments here. 

The same applies to other places in this file.

jkoenig134: urgh

jkoenig134: i was already through with this change and then reverted it as it is not correct. DVC1 marks the device 1 of an identity. So DVC1 is correct everywhere.

britsta: It is probably a matter of interpretation, the most unambiguous would probably be `"DVC1ofidentitiy1"` and `"DVC1ofidentity2"`. But since it's a detail, we don't have to change it now, even if I would find it more understandable.

jkoenig134: I think it really doesn't matter in this place, we just need some string :D

britsta (approved):
🐸

PR #197, opened 2024-07-18
Fix/do not use string as event data

jkoenig134:


Milena-Czierlinski:
How does the string cause problems exactly?

slavistan:
`npm run build` fails at the `build:schemas` step. Does this work on your ends?

jkoenig134:
We have problems with node 22.5 wich came out today. All pipelines are failing.

jkoenig134:
> How does the string cause problems exactly?

It's away from how we use the data everywhere else and therefore simply not handled correctly in the bridge.

Milena-Czierlinski (approved):

britsta (commented):
The line `await expect(services1.eventBus).toHavePublished(RelationshipDecomposedBySelfEvent, (e) => e.data === relationshipId);` of the test `relationship should be decomposed` of the file `relationships.test.ts` needs to be adjusted.

Milena-Czierlinski (approved):

PR #198, opened 2024-07-18
Feature/Set the communication language

Magnus-Kuhn:


packages/runtime/test/transport/devices.test.ts
```diff
@@ -19,4 +19,14 @@ describe("Devices", () => {
...
     });
+
+    test("should set the communication language", async () => {
+        const result = await transportServices1.devices.setCommunicationLanguage({ communicationLanguage: "fr" });
+        expect(result).toBeSuccessful();
+    });
+
+    test("should not set the communication language with an invalid language", async () => {
+        const result = await transportServices1.devices.setCommunicationLanguage({ communicationLanguage: "fra" });
+        expect(result).toBeAnError(/.*/, "error.runtime.devices.communicationLanguageNotISO639");
```
Milena-Czierlinski, line 30: You could add the error message here as a first parameter (I know we are not consistent with this in our existing tests, but it probably won't hurt to pay more attention to it.).

packages/runtime/src/useCases/transport/devices/SetCommunicationLanguage.ts
```diff
@@ -0,0 +1,25 @@
...
+    communicationLanguage: string;
+}
+
+export class SetCommunicationLanguageUseCase extends UseCase<SetCommunicationLanguageRequest, void> {
+    public constructor(@Inject private readonly deviceController: DeviceController) {
+        super();
+    }
+
+    protected async executeInternal(request: SetCommunicationLanguageRequest): Promise<Result<void>> {
+        if (request.communicationLanguage in LanguageISO639) {
```
jkoenig134, line 17: early return pls

packages/transport/src/modules/devices/backbone/DeviceClient.ts
```diff
@@ -8,4 +9,8 @@ export class DeviceClient extends RESTClient {
     public async createDevice(value: BackbonePostDevicesRequest): Promise<ClientResult<BackbonePostDevicesResponse>> {
         return await this.post<BackbonePostDevicesResponse>("/api/v1/Devices", value, {});
     }
+
+    public async setCommunicationLanguage(value: BackboneSetCommunicationLanguageRequest): Promise<ClientResult<void>> {
+        return await this.put<void>("/api/v1/Devices/Self", value);
+    }
```
jkoenig134, line 15: we have two options

1. rename this function to updateCurrentDevice
2. update the value to just accept one string and make `{ "communicationLanguage": value }` the payload

The version we have here is destroying the REST client that tries to be complete.

packages/runtime/src/useCases/transport/devices/SetCommunicationLanguage.ts
```diff
@@ -0,0 +1,25 @@
+import { Result } from "@js-soft/ts-utils";
+import { LanguageISO639 } from "@nmshd/content";
+import { DeviceController } from "@nmshd/transport";
+import { Inject } from "typescript-ioc";
+import { RuntimeErrors, UseCase } from "../../common";
+
+export interface SetCommunicationLanguageRequest {
+    communicationLanguage: string;
```
britsta, line 8: Is there a way to further restrict the type of the property `communicationLanguage` here so that only values that meet the ISO 639-1 standard can be used? Perhaps this would have advantages over performing the check according to the ISO 639-1 standard within the function call when the use case is exposed. A `SchemaValidator` could be added then too.

packages/transport/src/modules/devices/backbone/DeviceClient.ts
```diff
@@ -8,4 +9,8 @@ export class DeviceClient extends RESTClient {
     public async createDevice(value: BackbonePostDevicesRequest): Promise<ClientResult<BackbonePostDevicesResponse>> {
         return await this.post<BackbonePostDevicesResponse>("/api/v1/Devices", value, {});
     }
+
+    public async setCommunicationLanguage(value: BackboneSetCommunicationLanguageRequest): Promise<ClientResult<void>> {
+        return await this.put<void>("/api/v1/Devices/Self", value);
```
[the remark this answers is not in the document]
jkoenig134: https://github.com/nmshd/backbone/tree/main/ConsumerApi/http

packages/runtime/src/useCases/transport/devices/SetCommunicationLanguage.ts
jkoenig134: that's actually a great Idea, we could yeet the manual check then!

packages/runtime/src/useCases/transport/devices/SetCommunicationLanguage.ts
Magnus-Kuhn: AFAIK this is a matter of taste

packages/runtime/src/useCases/transport/devices/SetCommunicationLanguage.ts
Magnus-Kuhn: I'm generally not sure how the connector would handle enums properly, although this doesn't apply here - also is the policy for the schema validator to try adding it and removing it if things crash?

packages/transport/src/modules/devices/backbone/DeviceClient.ts
Magnus-Kuhn: I'll take option 2 - or are there other updating operations planned?

jkoenig134: I'm not sure what this route currently accepts. But I thought is's more than the communicationLanguage

packages/runtime/src/useCases/transport/devices/SetCommunicationLanguage.ts
jkoenig134: the connector does not handle an enum at all. for the connector this is just a string.

packages/runtime/src/useCases/transport/devices/SetCommunicationLanguage.ts
jkoenig134: no, it isn't.

jkoenig134: please see all the other UseCases, all of them first do the validation and then the action. Why should we do it different here because some matter of taste?

packages/transport/src/modules/devices/backbone/DeviceClient.ts
Magnus-Kuhn: According to swagger, it's only the language

packages/runtime/src/useCases/transport/devices/SetCommunicationLanguage.ts
Magnus-Kuhn: I didn't look at the other use cases, I look at this and think it's fine :) . It will be gone anyway using the enum

packages/transport/src/modules/devices/backbone/DeviceClient.ts
jkoenig134: Ok, I'd still rename the function to be ready for future changes.

packages/transport/test/modules/devices/SetCommunicationLanguage.test.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    afterAll(async function () {
+        await deviceTest.close();
+
+        await connection.close();
+    });
+
+    test("should set the communication language", async function () {
+        await expect(deviceAccount.activeDevice.setCommunicationLanguage("fr")).resolves.not.toThrow();
+    });
```
jkoenig134, line 33: missing a negative test here

packages/runtime/test/transport/devices.test.ts
jkoenig134, line 1: missing a negative test here

jkoenig134 (changes requested):
only the negative tests, rest lgtm

packages/runtime/test/transport/devices.test.ts
Magnus-Kuhn: the compiler won't let me do this anymore

jkoenig134: it does :) just do `"wrong-code" as any as LanguageISO639`

packages/transport/test/modules/devices/SetCommunicationLanguage.test.ts
Magnus-Kuhn: added the test, but it's failing @tnotheis

tnotheis: You passed a 3-digit language code. That's not allowed. Only 2-digit language codes are accepted.

tnotheis: Ah, sorry, I misread the code (or rather mixed up different things).

jkoenig134:
@tnotheis does all of this already works at version 5.9.1 of the backbone?

jkoenig134:
@Magnus-Kuhn the communication language feature dropped in the backbone at [6.0.0-alpha.13](https://github.com/nmshd/backbone/releases/tag/6.0.0-alpha.13) - we are way behind. Point this to release/v5 and it should work.

packages/transport/test/modules/devices/SetCommunicationLanguage.test.ts
tnotheis: So... I just tested it locally, and I get a 400 as expected:

image

jkoenig134 (approved):

jkoenig134:
@Magnus-Kuhn you've added an authenticated request to an unauthenticated client -> 401 :)

jkoenig134 (approved):

PR #199, opened 2024-07-22
Fix/Better error message for response to deleteAttributeRequest

Magnus-Kuhn:


packages/consumption/src/modules/requests/itemProcessors/deleteAttribute/DeleteAttributeRequestItemProcessor.ts
```diff
@@ -42,10 +42,11 @@ export class DeleteAttributeRequestItemProcessor extends GenericRequestItemProce
         params: AcceptDeleteAttributeRequestItemParametersJSON,
         _requestInfo: LocalRequestInfo
     ): Promise<ValidationResult> {
+        const parsedParams = AcceptDeleteAttributeRequestItemParameters.from(params);
         const attribute = await this.consumptionController.attributes.getLocalAttribute(requestItem.attributeId);
         if (!attribute) return ValidationResult.success();
 
-        const deletionDate = CoreDate.from(params.deletionDate);
+        const deletionDate = parsedParams.deletionDate;
```
britsta, line 49: Nice catch! 🐸 

I understand that we do not want to add a test that checks whether a better error message is now displayed. Hence the question: Have you checked that the right error message is now displayed instead of the incorrect previous one?

Magnus-Kuhn: Of course!!!

In the attributes runtime test "should send a sharing request containing a repository attribute that was already shared but is to be deleted by the peer", deleting the deletionDate gives "error.consumption.requests.servalErrorDuringRequestItemProcessing: 'AcceptDeleteAttributeRequestItemParameters.deletionDate :: Value is not defined' "

britsta: Niceee!

Tried the same, but my local Backbone don't want to start right now xD

britsta (approved):

jkoenig134 (approved):

PR #200, opened 2024-07-22
Fix/local development with different ways to connect to the backbone

jkoenig134:
When developing against a local backbone the backbone is always configured with `Application.didDomainName` set to `localhost`. There are different ways to connect to this connector (e.g. through the hostname in the docker network e.g. `"consumer-api"` or `"host.docker.internal"`. All these versions result into an address mismatch of the local generated address and the backbones.

This PR resolves this by providing a so called `addressGenerationHostnameOverride` that overrides that behavior of the runtime extracting the hostname for the address generation from the baseUrl.

this means you can configure the transport lib as follows:
```
{
  "baseUrl": "http://host.docker.internal:8080",
  "addressGenerationHostnameOverride": "localhost",
  ...
}
```

I am not really sure if the naming of the new config flag is the best, I'm open for suggestions here!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12546,7 +12546,7 @@
         },
         "packages/app-runtime": {
             "name": "@nmshd/app-runtime",
-            "version": "5.0.0-alpha.8",
+            "version": "5.0.0-alpha.9",
             "license": "MIT",
             "dependencies": {
                 "@js-soft/docdb-access-loki": "^1.1.0",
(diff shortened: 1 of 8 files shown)


jkoenig134:
It seems like the own address is not a sensitive information that may not be logged (as it is already logged in some places), I simplified the error message again.

sebbi08 (approved):

PR #204, opened 2024-07-24
Feature/Expand IdentityDVO from incoming Request if available

Magnus-Kuhn:


jkoenig134:
The idea was to not expand from a RelationshipTemplate and instead to expand from the open incoming Request.

packages/runtime/test/dataViews/IdentityDVO.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    templator.eventBus.reset();
+});
+
+describe("IdentityDVO after loading a relationship template sharing a DisplayName", () => {
+    beforeAll(async () => {
+        const senderAttribute = await templator.consumption.attributes.createRepositoryAttribute({
+            content: {
+                value: {
+                    "@type": "DisplayName",
+                    value: "Dr. Theodor Munchkin von Reichenhardt"
```
jkoenig134, line 29: "A Display Name"

packages/runtime/test/dataViews/IdentityDVO.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        templatorTemplate = await createTemplate(templator.transport, templateContent);
+    });
+    test("IdentityDVO should use the DisplayName", async () => {
+        const requestorTemplate = (await requestor.transport.relationshipTemplates.loadPeerRelationshipTemplate({ reference: templatorTemplate.truncatedReference })).value;
+
+        const dto = requestorTemplate;
+        const dvo = (await requestor.expander.expandRelationshipTemplateDTO(dto)) as PeerRelationshipTemplateDVO;
+        expect(dvo).toBeDefined();
+        expect(dvo.createdBy.name).toBe("Dr. Theodor Munchkin von Reichenhardt");
+        expect(dvo.createdBy.initials).toBe("DTMvR");
```
jkoenig134, line 57: also check peer

Magnus-Kuhn: what do you mean? Letting the peer expand his own address? The templator expanding the requestor's address could be a thing once personalized templates are there

jkoenig134: wait what? `dvo.peer` ist the same identity as `dvo.createdBy` the peer should be expanded in the exactly same way

All I want you to do is putting 

```
expect(dvo.peer.name).toBe("Dr. Theodor Munchkin von Reichenhardt");
expect(dvo.peer.initials).toBe("DTMvR");
```

as checks to this test.

If this doesn't work your expander doesn't work.

Magnus-Kuhn: dvo.peer doesn't exist on PeerRelationshipTemplateDVO

jkoenig134: Oooh, you are expanding a relationshipTemplate here? I didn't got that context. Please add another test where you expand a LocalRequest to a LocalRequestDVO where you can check `dvo.peer` AND `dvo.createdBy`. That's the usecase in the app.

packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -1588,11 +1588,24 @@ export class DataViewExpander {
...
+                query: {
+                    peer: address,
+                    status: [LocalRequestStatus.ManualDecisionRequired, LocalRequestStatus.DecisionRequired]
+                }
+            })
+        ).value;
+        requestResult.sort((request1, request2) => (request2.createdAt > request1.createdAt ? 1 : -1));
+        if (requestResult.length > 0) {
+            return this.expandAddressFromRequest(requestResult.at(-1)!);
+        }
```
jkoenig134, line 1606: we know there is no relationship - there will only be 0 or 1 request available.

Magnus-Kuhn: it is for the case that multiple different templates have been loaded previously, each generating their own request

jkoenig134: nope, we block these other loads.

Magnus-Kuhn:
```
await exchangeTemplate(sRuntimeServices.transport, rRuntimeServices.transport, templateContent);
await exchangeTemplate(sRuntimeServices.transport, rRuntimeServices.transport, templateContent);
const requestsResult = await rRuntimeServices.consumption.incomingRequests.getRequests({});
```
(tried it in the RequestModule.test around line 100) results in a length 2 array

jkoenig134: this is **really** bad. I'll look into it.

Magnus-Kuhn: To not have this conversation open forever: I assume now that there is max 1 request.

jkoenig134 (approved):

PR #208, opened 2024-07-26
Validate Backbone Compatibility

Magnus-Kuhn:
# Readiness checklist

-   [x] I labeled the PR.


packages/transport/src/core/Transport.ts
```diff
@@ -59,6 +63,8 @@ export class Transport {
         allowIdentityCreation: true,
         supportedDatawalletVersion: 1,
         supportedIdentityVersion: -1,
+        supportedMinBackboneVersion: 5,
+        supportedMaxBackboneVersion: 5,
```
Magnus-Kuhn, line 67: don't know what to put here

jkoenig134: 5 seems fine, we don't know if we will be compatible with v6

packages/transport/test/modules/account/AccountController.test.ts
jkoenig134, line 1: all these new tests will fail if the backbone reaches a specific major version. Wtb querying the backbone version before the test and adding / subtracting something from the actual version? Then the test is future proof.

packages/runtime/src/Runtime.ts
```diff
@@ -191,6 +191,14 @@ export abstract class Runtime<TConfig extends RuntimeConfig = RuntimeConfig> {
         this._anonymousServices = Container.get<AnonymousServices>(AnonymousServices);
     }
 
+    public async validateUsedBackboneVersion(): Promise<Result<void>> {
```
jkoenig134, line 194: "used" backbone version sounds quite off. WTB `checkBackboneCompatiblity`? This naming has to be aligned everywhere then.

packages/runtime/src/Runtime.ts
```diff
@@ -191,6 +191,14 @@ export abstract class Runtime<TConfig extends RuntimeConfig = RuntimeConfig> {
...
+        if (!this._accountController) {
```
jkoenig134, line 195: I don't see a reason why this is only possible with an active account.

packages/transport/src/core/Transport.ts
```diff
@@ -36,6 +38,8 @@ export interface IConfigOverwrite {
     platformClientId: string;
     platformClientSecret: string;
     supportedIdentityVersion: number;
+    supportedMinBackboneVersion?: number;
+    supportedMaxBackboneVersion?: number;
```
jkoenig134, line 42: this should not be overridable

packages/transport/src/core/Transport.ts
jkoenig134, line 67:
```suggestion
        supportedMinBackboneVersion: 6,
        supportedMaxBackboneVersion: 6,
```

packages/transport/test/modules/account/AccountController.test.ts
Magnus-Kuhn: Just a number is simple, they only need to be large enough

packages/runtime/src/Runtime.ts
Magnus-Kuhn: Both are fine to me

packages/transport/src/core/Transport.ts
Magnus-Kuhn: I put the overrides in for testing - changing the limits is necessary for the test if there is no mock

packages/runtime/src/Runtime.ts
jkoenig134: will you change this? YOu thumbed up.

Magnus-Kuhn: yeah, I will

packages/transport/test/modules/account/AccountController.test.ts
jkoenig134: Why not make it really future proof bei just taking the actual version and add to it? Why just increase numbers..?

packages/transport/src/modules/accounts/AccountController.ts
jkoenig134, line 1: I think this is not needed anymore

packages/runtime/src/extensibility/facades/anonymous/VersionFacade.ts
```diff
@@ -0,0 +1,11 @@
+import { Result } from "@js-soft/ts-utils";
+import { Inject } from "typescript-ioc";
+import { CheckBackboneCompatibilityUseCase } from "../../../useCases/anonymous/CheckBackboneCompatibilityUseCase";
+
+export class VersionFacade {
```
jkoenig134, line 5: Version -> BackboneVersion or BackboneCompatibility

same for everywhere in this PR. I prefer BackboneCompatibility. The only place where the wording is valid is the VersionClient in transport.

packages/transport/test/modules/account/AccountController.test.ts
Magnus-Kuhn: I'd have to think about how to get to the backbone version number and I don't expect enmeshed to still be around in the next millennium 💀

Milena-Czierlinski: @jkoenig134 If we want to get the Backbone version, we should probably expose it via a method of the `BackboneCompatibilityController`. This might be a valid method exceeding just testing. Then, the question arises if `BackboneCompatibilityController` is still an appropriate name or if we should switch back to something more general.

(I also noticed that the wording `version` is still used in some places. Depending on the outcome of this conversion, I'll adjust it accordingly.)

jkoenig134: > the next millennium

why do it hacky when you can do it clean?!

jkoenig134: > If we want to get the Backbone version, we should probably expose it via a method of the BackboneCompatibilityController

we should not expose things only for testing.

Milena-Czierlinski: > This might be a valid method exceeding just testing.

🤔

packages/runtime/src/useCases/anonymous/CheckBackboneCompatibilityUseCase.ts
```diff
@@ -0,0 +1,15 @@
+import { Result } from "@js-soft/ts-utils";
+import { BackboneCompatibilityController } from "@nmshd/transport";
+import { Inject } from "typescript-ioc";
+import { UseCase } from "../common";
+
+export class CheckBackboneCompatibilityUseCase extends UseCase<void, void> {
```
jkoenig134, line 6: void as successful result and throwing an error if it's not successful is very hard to process. Why not return a proper response that has 

1. a boolean to know if it's successful
2. an explanation on why it's not successful if it's not.

packages/transport/test/modules/backboneCompatibility/BackboneCompatibilityController.test.ts
```diff
@@ -0,0 +1,41 @@
+import { IDatabaseConnection } from "@js-soft/docdb-access-abstractions";
+import { BackboneCompatibilityController, Transport } from "../../../src";
+import { TestUtil } from "../../testHelpers/TestUtil";
+
+describe("AccountController", function () {
```
jkoenig134, line 5: naming

packages/transport/test/modules/account/AccountController.test.ts
jkoenig134: missed this.. but what do we need this for?

Milena-Czierlinski: Not sure at the current moment. But if we don't add a function for it, we'd have to make `client` of `BackboneCompatibilityController` public and I suppose we don't want that. Or do you see another way to get the Backbone version for testing?

jkoenig134: it's js / ts ..  we don't have to make anything public to access it :D

jkoenig134: I resolved it.

packages/runtime/src/extensibility/AnonymousServices.ts
```diff
@@ -1,6 +1,9 @@
 import { Inject } from "typescript-ioc";
-import { AnonymousTokensFacade } from "./facades/anonymous";
+import { AnonymousTokensFacade, BackboneCompatibilityFacade } from "./facades/anonymous";
 
 export class AnonymousServices {
-    public constructor(@Inject public readonly tokens: AnonymousTokensFacade) {}
+    public constructor(
+        @Inject public readonly tokens: AnonymousTokensFacade,
+        @Inject public readonly version: BackboneCompatibilityFacade
```
Milena-Czierlinski, line 7: `backboneCompatibility` instead of `version`.

packages/runtime/test/anonymous/version.test.ts
Milena-Czierlinski, line 1: File naming

packages/runtime/test/anonymous/version.test.ts
```diff
@@ -0,0 +1,56 @@
...
+afterEach(async () => {
+    await serviceProvider.stop();
+    await noLoginRuntime.stop();
+});
+
+describe("Backbone Compatibility Check", () => {
+    test("should successfully check the compatibility", async () => {
+        const transportConfigOverride = {
+            supportedMinBackboneVersion: 1,
+            supportedMaxBackboneVersion: 5000
```
Milena-Czierlinski, line 15: Should we do some hacking here as well?

packages/transport/src/core/CoreErrors.ts
```diff
@@ -197,6 +197,13 @@ class General {
     public invalidTruncatedReference() {
         return new CoreError("error.transport.files.invalidTruncatedReference", "invalid truncated reference");
     }
+
+    public runtimeVersionIncompatibleWithBackboneVersion(backboneVersion: number) {
+        return new CoreError(
+            "error.transport.files.runtimeVersionIncompatibleWithBackboneVersion",
```
Milena-Czierlinski, line 203: Should be `.general.` instead of `.files.`. Some other error codes here are inconsistent as well, but this is unrelated to the PR. Can we change it in a separate PR or would this be a breaking change such that we leave it for now?

packages/transport/test/modules/backboneCompatibility/BackboneCompatibilityController.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+        const versionCheckResult = await versionController.checkBackboneCompatibility();
+        expect(versionCheckResult.isSuccess).toBe(true);
+    });
+
+    test("should catch a too low backbone version", async function () {
+        const versionController = new BackboneCompatibilityController({
+            ...transport.config,
+            supportedMinBackboneVersion: actualBackboneVersion + 1,
+            supportedMaxBackboneVersion: actualBackboneVersion + 2
```
Milena-Czierlinski, line 50: I guess it's not important, but is there a reason for `+2` instead of `1` which is used in the other tests?

jkoenig134: I think it doesn't matter here, if it's confusing I can use +1 instead.

packages/runtime/test/anonymous/version.test.ts
jkoenig134: yep :)

jkoenig134 (approved):

packages/runtime/src/useCases/anonymous/backboneCompatibility/CheckBackboneCompatibilityUseCase.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface CheckBackboneCompatibilityResponse {
+    isCompatible: boolean;
+    backboneVersion: number;
+    supportedMinBackboneVersion: number;
+    supportedMaxBackboneVersion: number;
+}
+
+export class CheckBackboneCompatibilityUseCase extends UseCase<void, CheckBackboneCompatibilityResponse> {
+    public constructor(@Inject private readonly anonymousVersionController: BackboneCompatibilityController) {
```
Milena-Czierlinski, line 14: `backboneCompatibilityController` instead of `anonymousVersionController`

Milena-Czierlinski (approved):

jkoenig134 (approved):

PR #211, opened 2024-07-30
fix/correct test after merge

Magnus-Kuhn:


Milena-Czierlinski:
I think if you run `npm audit fix`, you'll get rid of the failing test.

Also, probably a version bump should be performed for this fix.

Magnus-Kuhn:
Yeah, it would fix it - I also checked that dependabot did the fix on main, but it's not on v5 due to merge conflicts. I'll ask Julian to merge that, then the error should be gone.

jkoenig134:
already done when merging.

PR #213, opened 2024-07-31
Fix/ShareAttribute with deletionInfo

Milena-Czierlinski:
This tackles the open comments from @britsta on https://github.com/nmshd/runtime/pull/174


jkoenig134 (approved):
LGTM

britsta (approved):

britsta:
Do we also need a version bump in this PR?

Milena-Czierlinski:
> Do we also need a version bump in this PR?

Strictly speaking yes. But I'll merge another PR that contains a version bump directly, too, so we should be fine. :D

PR #218, opened 2024-08-05
Feature/Personalized templates and tokens

Magnus-Kuhn:


jkoenig134:
Please split templates and tokens and then come back with another PR.

jkoenig134:
@Magnus-Kuhn you can use backbone version `6.6.0-alpha.1` which implements personalized tokens.

PR #219, opened 2024-08-05
ci/Add code coverage

Magnus-Kuhn:


.github/workflows/test.yml
```diff
@@ -31,6 +31,11 @@ jobs:
           NMSHD_TEST_BASEURL: http://localhost:8090
           NMSHD_TEST_CLIENTID: test
           NMSHD_TEST_CLIENTSECRET: test
+      - name: Upload coverage reports to Codecov with GitHub Action
+        uses: codecov/codecov-action@v4
+        env:
+          fail_ci_if_error: true
```
jkoenig134, line 37: I'm not a fan of this (in fact we should also disable this in the connector again). coverage should only be an information for us and not a hard requirement.

.github/workflows/test.yml
jkoenig134, line 1: the triggers for this workflow should be moved from 

```
on:
  push:
    branches-ignore: [main]
```

to 

```
on:
  pull_request:
  push:
    branches: [main]
```

packages/app-runtime/package.json
```diff
@@ -23,11 +23,21 @@
         "cdep": "tsc && madge --circular dist",
         "lint:tsc": "tsc --noEmit && tsc -p test/tsconfig.json --noEmit",
         "test": "jest --forceExit",
-        "test:ci": "jest -i --forceExit"
+        "test:coverage": "npm run test -- --coverage=true",
+        "test:ci": "jest -i --forceExit --coverage=true"
```
jkoenig134, line 27:
```suggestion
        "test:coverage": "npm run test -- --coverage",
        "test:ci": "jest -i --forceExit --coverage"
```

same for all other occurences.

packages/consumption/package.json
```diff
@@ -24,17 +24,27 @@
         "lint:tsc": "tsc --noEmit && tsc -p test/tsconfig.json --noEmit",
         "test:ci": "npm run test",
         "test:ci:lokijs": "USE_LOKIJS=true jest -i",
-        "test:ci:mongodb": "jest -i",
+        "test:ci:mongodb": "jest -i  --coverage=true",
         "test:local:ferretdb": "npm run test:local:start:ferretdb && CONNECTION_STRING='mongodb://localhost:27022' jest",
         "test:local:lokijs": "USE_LOKIJS=true jest",
         "test:local:mongodb": "npm run test:local:start:mongodb && CONNECTION_STRING='mongodb://root:example@localhost:27021' jest",
+        "test:local:coverage": "npm run test:local:lokijs -- --coverage=true",
```
jkoenig134, line 31: IMO we don't need the local scripts. What if you want to test the coverage against mongodb? just run `npm run test:local:mongodb -- --coverage` same for **all** other places.

packages/runtime/package.json
```diff
@@ -24,17 +24,27 @@
         "cdep": "tsc && madge --circular dist",
         "lint:tsc": "tsc --noEmit && tsc -p test/tsconfig.json --noEmit",
         "test:ci:lokijs": "USE_LOKIJS=true jest -i",
-        "test:ci:mongodb": "jest -i",
+        "test:ci:mongodb": "jest -i  --coverage=true",
```
jkoenig134, line 27:
```suggestion
        "test:ci:mongodb": "jest -i --coverage=true",
```

same for all other places.

jkoenig134:
some details. Overall looks good.

@sebbi08 please also give your sense as you made the changes on the connector.

.github/workflows/test.yml
sebbi08: That option has nothing to do with the coverage check, it is only for the upload of the coverage files.
We can disable this, but I think the CI check needs to be disabled in the codecov config.

jkoenig134: Ah, IC. then nvm. Upload errors SHOULD fail this ppl.

packages/consumption/package.json
sebbi08: I don't like it either that there are so many npm commands.

I also think we should remove it and maybe document it in the dev README

Magnus-Kuhn: > What if you want to test the coverage against mongodb?

Does this matter?
I can change it, I simply copied it from the connector ;)

jkoenig134: yeah, it does.

jkoenig134 (approved):

PR #221, opened 2024-08-06
Fix/display unknown identities as unknown instead of a random string

jkoenig134:
Unknown identities get assigned a "random" string which cannot be caught by the app. For example with `localhost` as hostname it's `:e:loc`.

This now assigns a safe, reproducable and processable string to these identities.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12548,7 +12548,7 @@
         },
         "packages/app-runtime": {
             "name": "@nmshd/app-runtime",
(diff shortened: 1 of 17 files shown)


sebbi08:
Might this string also come up in the connector ? And should we handle it there ?

sebbi08 (approved):

jkoenig134:
The data views are not used in the connector at all, so no - ignore that.

PR #227, opened 2024-08-09
Block new Relationship creation if old Relationship is not yet fully decomposed

britsta:
Adding validation for blocking the creation of a new Relationship to a peer locally for which a Relationship still exists that is in the "Terminated" or "DeletionProposed" state. 

**Completed TODO of the Backbone:** If the old Relationship has already been decomposed by an Identity itself, but the peer has not yet been decomposed the old Relationship, the Backbone must provide this information to the Identity as it is not available to it locally. As soon as this feature is provided by the Backbone (https://github.com/nmshd/backbone/pull/799), it can be utilized accordingly in the Runtime.


jkoenig134:
> the Backbone must provide this information to the Identity as it is not available to it locally. As soon as this feature is provided by the Backbone (https://github.com/nmshd/backbone/pull/799)

https://github.com/nmshd/backbone/pull/799 does not provide the /canCreate feature for relationships, it only blocks the creation of relationships. Thus we have to catch errors occuring from createRelationship which is called in `RequestModule#respondToRequestViaRelationship` and do a proper rollback there. As soon as we have the functionality from the backbone we can enhance `canDecide` to say no if we cannot answer the Request via a Relationship.

=> that feature will be implemented in the next sprint by the bkb team.

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -168,6 +168,35 @@ export class RelationshipsController extends TransportController {
...
+                            ["status"]: { $eq: RelationshipStatus.DeletionProposed }
+                        }
+                    ]
+                };
+
+                const existingRelationshipToPeer = await this.getRelationships(queryForExistingRelationship);
+
+                if (existingRelationshipToPeer.length !== 0) {
+                    throw CoreErrors.relationships.alreadyExists(existingRelationshipToPeer[0].status);
+                }
```
jkoenig134, line 195: you should do that **before** calling the backbone with this.client.createRelationship.

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -168,6 +168,35 @@ export class RelationshipsController extends TransportController {
...
+                        {
+                            ["status"]: { $eq: RelationshipStatus.Active }
+                        },
+                        {
+                            ["status"]: { $eq: RelationshipStatus.Terminated }
+                        },
+                        {
+                            ["status"]: { $eq: RelationshipStatus.DeletionProposed }
+                        }
+                    ]
```
jkoenig134, line 188: There is an operator for this:
```
                    "status": { "$in": [RelationshipStatus.Pending, RelationshipStatus.Active, ...] }
```

packages/transport/src/modules/relationships/RelationshipsController.ts
britsta: Ahh I see, because information from the Backbone is only required if I throw the `CoreErrors.relationships.notYetDecomposedByPeer`?

packages/transport/src/modules/relationships/RelationshipsController.ts
britsta: Nice, that's more compact 😎 I will refactor this here and in other open PRs of mine.

packages/runtime/test/lib/testUtils.ts
Milena-Czierlinski, line 1: I would treat both services symmetrically. In tests, where `ensureActiveRelationship` is called, it might be the case the there is not a clear sender and recipient, but that Requests or RelationshipTemplates are sent in both directions.

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -82,6 +82,20 @@ describe("Create Relationship", () => {
     });
 
     describe("tests on pending relationship", () => {
+        test("should not be able to create a new relationship if a pending relationship already exists", async () => {
```
Milena-Czierlinski, line 85: This test isn't successful on its own, as it is dependent on "create pending relationship". However, I noticed that this is the case for also other tests, so we should take care of this in a separate PR.

packages/transport/src/core/CoreErrors.ts
```diff
@@ -14,6 +14,20 @@ class Relationships {
         );
     }
 
+    public alreadyExists(status: RelationshipStatus) {
```
Milena-Czierlinski, line 17: I'd rename the error to `relationshipAlreadyExists` for clarity.

packages/transport/src/core/CoreErrors.ts
```diff
@@ -14,6 +14,20 @@ class Relationships {
...
+        return new CoreError(
+            "error.transport.relationships.alreadyExists",
+            `No new Relationship to the peer can be created as a Relationship in status '${status}' currently exists.`
+        );
+    }
+
+    public notYetDecomposedByPeer() {
```
Milena-Czierlinski, line 24: I'd rename the error to `relationshipNotYetDecomposedByPeer` or `relationshipNotDecomposedByPeer` for clarity.

packages/transport/src/core/CoreErrors.ts
```diff
@@ -14,6 +14,20 @@ class Relationships {
...
+        return new CoreError(
+            "error.transport.relationships.notYetDecomposedByPeer",
+            "No new Relationship can be created as the former Relationship is not yet fully decomposed by the peer."
```
Milena-Czierlinski, line 27: I'd omit the "fully", since the peer may either have called the decompose route or not but nothing in between.

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -155,6 +155,17 @@ export class RelationshipsController extends TransportController {
             throw this.newCacheEmptyError(RelationshipTemplate, template.id.toString());
         }
 
+        const queryForExistingRelationship = {
+            "peer.address": template.cache.createdBy.toString(),
+            status: { $in: [RelationshipStatus.Pending, RelationshipStatus.Active, RelationshipStatus.Terminated, RelationshipStatus.DeletionProposed] }
```
Milena-Czierlinski, line 160: Do I see it correctly, that this line won't be needed, if the idea of @jkoenig134 is implemented to directly delete rejected and revoked relationships?

packages/transport/src/core/CoreErrors.ts
britsta: This error will be thrown as `error.transport.relationships.notYetDecomposedByPeer`, as `notYetDecomposedByPeer` is just the name of the error's suffix. Do you still want to rename it, so that `error.transport.relationships.relationshipNotYetDecomposedByPeer` will be thrown?

packages/transport/src/core/CoreErrors.ts
britsta: This error will be thrown as `error.transport.relationships.alreadyExists`, as `alreadyExists` is just the name of the error's suffix. Do you still want to rename it, so that `error.transport.relationships.relationshipAlreadyExists` will be thrown?

packages/transport/src/core/CoreErrors.ts
Milena-Czierlinski: I'd prefer it that way. 👍

packages/runtime/test/transport/relationships.test.ts
britsta: You are absolutely right that this is not good. This should be adjusted in a separate PR.

packages/runtime/test/lib/testUtils.ts
```diff
@@ -417,13 +417,38 @@ export async function establishPendingRelationshipWithRequestFlow(
...
+
+    const sTransportServicesAddress = (await sTransportServices.account.getIdentityInfo()).value.address;
+    const rRelationships = (await rTransportServices.relationships.getRelationships({ query: { peer: sTransportServicesAddress } })).value;
+
+    if (sRelationships.length === 0 && rRelationships.length === 0) {
         await establishRelationship(sTransportServices, rTransportServices);
-    } else if (relationships[0].status === RelationshipStatus.Pending) {
-        const relationship = relationships[0];
-        await sTransportServices.relationships.acceptRelationship({ relationshipId: relationship.id });
+    } else if (sRelationships.length === 0 && rRelationships.length !== 0) {
```
Milena-Czierlinski, line 427: Could we (currently) run into a problem here, since the recipient might have a rejected or revoked Relationship and not like you implicitly assume a deletion proposed Relationship? Maybe add a check for safety (and even if revoked and rejected Relationships are directly deleted in the future) and for better understanding.

packages/runtime/test/lib/testUtils.ts
```diff
@@ -417,13 +417,38 @@ export async function establishPendingRelationshipWithRequestFlow(
...
+        const relationship = rRelationships[0];
+        await rTransportServices.relationships.decomposeRelationship({ relationshipId: relationship.id });
+        await establishRelationship(sTransportServices, rTransportServices);
+    } else if (sRelationships[0].status === RelationshipStatus.Pending) {
+        if (sRelationships[0].template.isOwn) {
+            const relationship = sRelationships[0];
+            await sTransportServices.relationships.acceptRelationship({ relationshipId: relationship.id });
+            await syncUntilHasRelationships(rTransportServices, 1);
+        } else if (rRelationships[0].template.isOwn) {
```
Milena-Czierlinski, line 436: Since you queried for the peer above, I think a simple `else` is sufficient here.

packages/runtime/test/lib/testUtils.ts
Milena-Czierlinski: Same thing if the sender has a rejected or revoked Relationship.

Milena-Czierlinski (approved):

jkoenig134:
@britsta please wait with the merge for #236 to be merged!

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -155,6 +155,17 @@ export class RelationshipsController extends TransportController {
...
+        };
+
+        const existingRelationshipToPeer = await this.getRelationships(queryForExistingRelationship);
+
+        if (existingRelationshipToPeer.length !== 0) {
+            throw CoreErrors.relationships.relationshipAlreadyExists(existingRelationshipToPeer[0].status);
```
jkoenig134, line 166: is **already**Exists the correct naming for a deletion proposed relationship? I think the better wording would be **still**Exists in that case. But that's a detail, if you don't see it like that you can ignore this :)

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -155,6 +155,17 @@ export class RelationshipsController extends TransportController {
...
+        }
```
jkoenig134, line 167: you could put this code to a canCreateRelationship method for re-use! But you can also do that in another PR.

packages/transport/src/modules/relationships/RelationshipsController.ts
jkoenig134: yeeeah :)

jkoenig134 (approved):

PR #229, opened 2024-08-12
test/add identityDVO with a revoked relationship test

Magnus-Kuhn:


packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -1595,7 +1595,7 @@ export class DataViewExpander {
         }
 
         const result = await this.transport.relationships.getRelationshipByAddress({ address });
-        // revoked relationships should be handled like not existant as the will never have attributes attached
+        // revoked relationships should be handled like not existing as they will never have attributes attached
```
jkoenig134, line 1598:
```suggestion
        // revoked relationships should be treated as non-existent as they will never have attributes attached
```

jkoenig134 (approved):

PR #234, opened 2024-08-15
Fix/Detach copies of child Attributes

Milena-Czierlinski:
I previously thought we didn't need this, since when sharing a complex Attribute no child copies are created. However, it is possible to also share child Attributes on their own and then when deleting the complex parent Attribute, these copies need to be detached, too.

### DIFF
--- a/packages/consumption/test/modules/attributes/AttributesController.test.ts
+++ b/packages/consumption/test/modules/attributes/AttributesController.test.ts
@@ -958,6 +958,8 @@ describe("AttributesController", function () {
             let predecessorChildAttribute: LocalAttribute;
             let successorChildAttribute: LocalAttribute;
 
+            let predecessorChildOwnSharedIdentityAttribute: LocalAttribute;
+
             beforeEach(async () => {
                 predecessorComplexAttribute = await consumptionController.attributes.createRepositoryAttribute({
                     content: IdentityAttribute.from({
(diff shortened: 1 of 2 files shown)


jkoenig134 (approved):

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -1040,6 +1048,15 @@ describe("AttributesController", function () {
                 expect(updatedSuccessorChildAttribute!.succeeds).toBeUndefined();
             });
 
+            test("should detach shared attribute copies of deleted child attribute", async function () {
+                const sharedChildAttributeBeforeDeletion = await consumptionController.attributes.getLocalAttribute(predecessorChildOwnSharedIdentityAttribute.id);
+                expect(sharedChildAttributeBeforeDeletion!.shareInfo!.sourceAttribute).toStrictEqual(predecessorChildAttribute.id);
```
britsta, line 1053: The `parentId` field of the `predecessorChildOwnSharedIdentityAttribute` is already undefined before the deletion of the `predecessorComplexAttribute`, right?

britsta (approved):

britsta:
Was there no version bump on purpose?

jkoenig134:
@Milena-Czierlinski stop it! please don't do version bumps for now, we will now start to pack releases.

packages/consumption/test/modules/attributes/AttributesController.test.ts
Milena-Czierlinski: Yes

PR #236, opened 2024-08-19
Ci/versioning with GitHub releases

jkoenig134:
We will now publish our runtime using [GitHub Releases](https://github.com/nmshd/runtime/releases), this has multiple effects on the work with this repository:

1. as discussed IQL was moved to another repo again as it doesn't need to get published along with the runtime: https://github.com/nmshd/iql
2. we do not upgrade versions of libs anymore in this Repository, instead we do our work and then release if we want using tags (created by gh releases)
3. We have to label our pull requests now. Additionally we do not use Scopes for PR titles anymore
    
    `Feature/A cool new Feature` becomes `A cool new Feature` with the Label `Feature`


.ci/publishNpm.sh
```diff
@@ -0,0 +1,19 @@
...
+if [ -z "$VERSION" ]; then
+    echo "The environment variable 'VERSION' must be set."
+    exit 1
+fi
+
+# set the version of all packages in the workspace to $VERSION
+npm version -ws $VERSION
+
+# replace all dependencies from the current workspace that are marked with "*" with the version that gets published
+find . -regex "./packages/[A-Za-z-]*/package\.json$" -exec sed -i -e "s/\"\*\"/\"$VERSION\"/g" {} \;
```
slavistan, line 13: `-i` differs between GNU/Linux and MacOS implementations of `sed` (see [here](https://www.grymoire.com/Unix/Sed.html#uh-62h)). I'm not sure if that's still an issue, but I have this in the back of my head, since I've had a couple of scripts break when I ran them on a mac a couple of years ago.

If this script will only ever run in (linux) pipelines forget about it.

slavistan, line 13: This works by replacing all `*`s with $VERSION which is fine. If ever you should need a little more robust solution you can use the json parser `jq`, which can overwrite fields, too.

jkoenig134: I only tried it on mac so far, if it runs on linux like this it should work.

jkoenig134: It should replace `"*"` with `"<the-version>"` which should be quite robust. As we have more than one field to write (for example in runtime it's three) it would be many commands.

.github/workflows/check-pr.yml
```diff
@@ -0,0 +1,17 @@
+name: Check Pull Request
+
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize]
```
Milena-Czierlinski, line 5: For what is `synchronize` (without ending d?) used?

jkoenig134: pushing on the PR

jkoenig134:
@Magnus-Kuhn & @britsta make sure to read the description and update the titles and labels of your open PRs!

package-lock.json
Milena-Czierlinski, line 1: Did we only need the package-lock for the manual versioning?

jkoenig134: the file was not removed, it's just way smaller.

jkoenig134:
@stnmtz FYI!

Milena-Czierlinski (approved):
Looks reasonable to me, but I wouldn't consider myself to be an expert in the `.ci` and `.github` stuff. @jkoenig134 Just in case you want to wait for a more in depth review on that. Otherwise, if anything should arise we can fix it agily. :D

jkoenig134:
Thanks @Milena-Czierlinski, lets be agile!

PR #240, opened 2024-08-20
Move natives to app runtime

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

# Description

When migration from cordova to flutter we tried to make the natives less complex by splitting them down into "runtimeNatives". This was a way smaller effort then this - but we still couldn't get rid of the real complexity of the `@js-soft/native-abstraction` library. I've copied everything we really need from there to the app-runtime. This will allow us to archive the natives and have one repo less to maintain.


jkoenig134:
Draft b/c it should be possible to remove many unused things.

packages/app-runtime/src/modules/AppRuntimeModule.ts
```diff
@@ -12,14 +12,9 @@ export interface AppRuntimeModuleConfiguration extends ModuleConfiguration {}
...
-    protected subscribeToNativeEvent<TEvent>(event: Event, handler: EventHandler<TEvent>): void {
+    protected subscribeToNativeEvent<TEvent>(event: SubscriptionTarget<TEvent>, handler: EventHandler<TEvent>): void {
         const subscriptionResult = this.runtime.nativeEnvironment.eventBus.subscribe(event, handler);
-        if (subscriptionResult.isError) {
-            this.logger.error(subscriptionResult.error);
-            throw subscriptionResult.error;
-        }
-
-        this.nativeEventSubscriptionIds.push(subscriptionResult.value);
+        this.nativeEventSubscriptionIds.push(subscriptionResult);
```
Milena-Czierlinski, line 17: What is the reason for these changes?

packages/app-runtime/src/natives/INativeDeviceInfoAccess.ts
```diff
@@ -0,0 +1,23 @@
+import { Result } from "@js-soft/ts-utils";
+
+export enum PushServices {
+    "apns" = "apns", // eslint-disable-line @typescript-eslint/naming-convention
+    "fcm" = "fcm", // eslint-disable-line @typescript-eslint/naming-convention
+    "none" = "none" // eslint-disable-line @typescript-eslint/naming-convention
```
Milena-Czierlinski, line 6: Can we add `"dummy"` here? (I remember we did some black magic a while ago to set that... https://github.com/nmshd/runtime/blob/ac65997fdaab9df917f46012e75ee6a57744a583/packages/app-runtime/test/mocks/NativeDeviceInfoAccessMock.ts#L14)

Milena-Czierlinski: Wow, such a pretty formatting from git. 🤩

packages/app-runtime/src/natives/NativeEvents.ts
Milena-Czierlinski, line 1: I know the individual code snippets are small, but wouldn't is align better with our exiting code if we had a separate file for every event?

packages/app-runtime/src/natives/INativeDeviceInfoAccess.ts
jkoenig134: hmmmm, IMO we should do that later, but good catch!

packages/app-runtime/src/natives/NativeEvents.ts
jkoenig134: true!

packages/app-runtime/src/natives/INativeDeviceInfoAccess.ts
jkoenig134: no, NOW!

packages/app-runtime/src/modules/AppRuntimeModule.ts
Milena-Czierlinski: And this not only affects logging and throwing errors, but also lets you push a result and not its value?

jkoenig134: I don't understand :D

jkoenig134: it's only simplified

Milena-Czierlinski (approved):

PR #241, opened 2024-08-21
Add configuration for setting isDefault of LocalAttributes

Milena-Czierlinski:
# Readiness checklist

-   [x] I labeled the PR.


jkoenig134:
Awesome!!! We have to bring that to the Connector ASAP!

packages/consumption/src/consumption/ConsumptionControllerName.ts
Milena-Czierlinski, line 1: @jkoenig134 Can I delete the Controllers that are not used?

packages/consumption/src/consumption/ConsumptionController.ts
```diff
@@ -87,7 +89,12 @@ export class ConsumptionController {
         requestItemProcessorOverrides = new Map<RequestItemConstructor, RequestItemProcessorConstructor>(),
         notificationItemProcessorOverrides = new Map<NotificationItemConstructor, NotificationItemProcessorConstructor>()
     ): Promise<ConsumptionController> {
-        this._attributes = await new AttributesController(this, this.transport.eventBus, this.accountController.identity).init();
+        this._attributes = await new AttributesController(
+            this,
+            this.transport.eventBus,
+            this.accountController.identity,
+            this.consumption.config.setDefaultRepositoryAttributes
```
Milena-Czierlinski, line 96: Not sure if I should do it like this or hand down `config` as a whole.

packages/runtime/src/Runtime.ts
```diff
@@ -148,6 +151,7 @@ export abstract class Runtime<TConfig extends RuntimeConfig = RuntimeConfig> {
         this.eventBus.publish(new RuntimeInitializedEvent());
     }
 
+    // TODO: well this seems redundant
     protected initInfrastructure(): void | Promise<void> {
         return;
     }
```
Milena-Czierlinski, line 157: @jkoenig134 Can these no-op functions be removed?

jkoenig134: nope. these are protected and used in the connector.

jkoenig134: *overwritten

packages/consumption/src/consumption/ConsumptionControllerName.ts
jkoenig134: yes

packages/consumption/src/index.ts
jkoenig134, line 1: as discussed - undo pls

jkoenig134:
only the imports - rest lgtm

packages/runtime/src/useCases/consumption/attributes/ChangeDefaultRepositoryAttribute.ts
```diff
@@ -26,6 +26,10 @@ export class ChangeDefaultRepositoryAttributeUseCase extends UseCase<ChangeDefau
     }
 
     protected async executeInternal(request: ChangeDefaultRepositoryAttributeRequest): Promise<Result<LocalAttributeDTO>> {
+        if (!this.attributesController.parent.consumptionConfig.setDefaultRepositoryAttributes) {
+            return Result.fail(RuntimeErrors.attributes.setDefaultRepositoryAttributesIsDisabled());
+        }
+
```
jkoenig134, line 32: do we really need this?

1. I am not sure if I like the access to the ConsumptionController to the config here
2. this will fail in line 40 with a good enough error
3. Point 2 should be enough as it's very unlikely that someone who has configured `setDefaultRepositoryAttributes` to false accesses that usecase.

jkoenig134: if you agree, don't forget to also remove the error from `packages/runtime/src/useCases/common/RuntimeErrors.ts`

Milena-Czierlinski: Not sure if I would agree with your 3. point, also given that false is the default setting. ;) In this case I think it is more appropriate to return this error before one of the other errors can be returned. Otherwise it might be frustrating figuring out what you did wrong, fixing it, and then getting the error that you cannot use this use case in the first place. And I don't see how it hurts here.

jkoenig134: wtb not setting a default?

Milena-Czierlinski: What do you mean by that?

jkoenig134: let's have a ☎ about this

packages/consumption/src/consumption/ConsumptionController.ts
jkoenig134: In general I am not a big fan of boolean parameters, but with enabled inlay hints it's ok here.

jkoenig134 (dismissed):

jkoenig134 (approved):

PR #244, opened 2024-08-21
simplify

jkoenig134:
I just created an approach to massively simplify your PR. I am not a fan of the Transport class, we should definitely not copy the mistakes we did when implementing this.


jkoenig134:
I fd up, but the idea should be clear :) seams like I misconfigured the consumption tests

PR #249, opened 2024-08-23
Personalized Tokens

Magnus-Kuhn:
# Readiness checklist

-   [x] I labeled the PR.


packages/transport/test/modules/tokens/TokenController.test.ts
```diff
@@ -151,6 +151,63 @@ describe("TokenController", function () {
...
+        const sentToken = await sender.tokens.sendToken({
+            content,
+            expiresAt,
+            ephemeral: false,
+            forIdentity: sender.identity.address
+        });
+        const reference = sentToken.toTokenReference().truncate();
+        await TestUtil.expectThrowsAsync(async () => {
+            await recipient.tokens.loadPeerTokenByTruncated(reference, false);
+        }, /transport.general.notIntendedForYou/);
```
Milena-Czierlinski, line 190: Judging from other tests, I think we prefer strings. (Also below.)

packages/transport/test/modules/tokens/TokenController.test.ts
```diff
@@ -151,6 +151,63 @@ describe("TokenController", function () {
...
+    });
+
+    test("should throw if a personalized token is not loaded by the right identity and it's uncaught before reaching the BB", async function () {
```
Milena-Czierlinski, line 193: What do you mean by "it's uncaught before reaching the BB"? Also, I'd write "Backbone" instead of "BB".

Milena-Czierlinski:
We should also keep the ongoing discussion about adjusting the truncated Reference for a specific Backbone or passwords in mind.

jkoenig134:
> We should also keep the ongoing discussion about adjusting the truncated Reference for a specific Backbone or passwords in mind.

I'm also confused why this PR changes the truncated reference. Theres another ticket that should do that, I'd say we should remove this code for now.

packages/transport/test/modules/tokens/TokenController.test.ts
jkoenig134: yeah, no abbreviations..

Magnus-Kuhn: When loading a personalized token only with ID and key we can't catch it here, so I want to test that the BB catches it. I'll clarify it in the test name

packages/transport/src/modules/tokens/local/Token.ts
```diff
@@ -50,7 +50,8 @@ export class Token extends CoreSynchronizable implements IToken {
     }
 
     public toTokenReference(): TokenReference {
-        return TokenReference.from({ id: this.id, key: this.secretKey, forIdentity: this.cache?.forIdentity });
+        // TODO: add the token reference forIdentity once available
```
jkoenig134, line 53: no todo comments in code we want to merge please.

packages/runtime/test/lib/testUtils.ts
```diff
@@ -171,6 +171,21 @@ export async function uploadOwnToken(transportServices: TransportServices): Prom
     return response.value;
 }
 
+export async function uploadPersonalizedOwnToken(transportServices: TransportServices, forIdentity: string): Promise<TokenDTO> {
```
jkoenig134, line 174: why not simplfy extend the function above with an optional forIdentity parameter?

packages/transport/src/modules/tokens/AnonymousTokenController.ts
```diff
@@ -19,10 +19,14 @@ export class AnonymousTokenController {
     }
 
     public async loadPeerTokenByReference(tokenReference: TokenReference): Promise<Token> {
+        // TODO: add the token reference forIdentity once available
```
jkoenig134, line 22: TODO in code - remove please. ALso we don't have to change sth here when the updated truncated refereces are landing as the only thing that changes will be the length of the string.

packages/transport/src/modules/tokens/AnonymousTokenController.ts
```diff
@@ -19,10 +19,14 @@ export class AnonymousTokenController {
...
         return await this.loadPeerToken(tokenReference.id, tokenReference.key);
     }
 
-    public async loadPeerToken(id: CoreId, secretKey: CryptoSecretKey): Promise<Token> {
+    public async loadPeerToken(id: CoreId, secretKey: CryptoSecretKey, forIdentity?: CoreAddress): Promise<Token> {
+        if (forIdentity) {
+            throw TransportCoreErrors.general.notIntendedForYou(id.toString());
+        }
```
jkoenig134, line 29:
```suggestion
        }

```

packages/transport/src/modules/tokens/TokenController.ts
```diff
@@ -187,11 +190,16 @@ export class TokenController extends TransportController {
...
         return await this.loadPeerToken(tokenReference.id, tokenReference.key, ephemeral);
     }
 
-    public async loadPeerToken(id: CoreId, secretKey: CryptoSecretKey, ephemeral: boolean): Promise<Token> {
+    public async loadPeerToken(id: CoreId, secretKey: CryptoSecretKey, ephemeral: boolean, forIdentity?: CoreAddress): Promise<Token> {
         const tokenDoc = await this.tokens.read(id.toString());
+        if (!tokenDoc && forIdentity && !forIdentity.equals(this.parent.identity.address)) {
+            // if you created the token, it exists already
+            throw TransportCoreErrors.general.notIntendedForYou(id.toString());
+        }
```
jkoenig134, line 202: you can not throw here. you have to do it after the call to `await this.updateCacheOfToken(token);` below. PLease also check why the test didn't shows this.

packages/transport/test/modules/tokens/AnonymousTokenController.test.ts
```diff
@@ -0,0 +1,61 @@
+import { IDatabaseConnection } from "@js-soft/docdb-access-abstractions";
+import { Serializable } from "@js-soft/ts-serval";
+import { CoreDate } from "@nmshd/core-types";
+import { AccountController, AnonymousTokenController, Transport } from "../../../src";
+import { TestUtil } from "../../testHelpers/TestUtil";
+
+describe("TokenController", function () {
```
jkoenig134, line 7: test name.

packages/transport/src/modules/tokens/local/Token.ts
sebbi08: I converted the PR to a draft as I would highly recommend adding this feature back so we can catch wrong personalized tokens before sending them to the backend and getting that error.

Backend error will be and 404 Not found. And we could send "Not intended for you"

packages/transport/src/modules/tokens/TokenController.ts
Magnus-Kuhn: I don't understand the issue, I'm not using data from the existing token

sebbi08: Why should we not throw here? If it is not intended for you, there should be nothing in the DB and the update of cache would throw an 404.

jkoenig134: How will it come to the db?

packages/transport/src/modules/tokens/local/Token.ts
jkoenig134: We should not handle this task like this.

The extension of the truncated reference should only be an optimization to intercept early when you don't have access to the token. As the trunacted reference is not the only way to load tokens we should find other ways to make this secure - for example going back to the backbone team and asking for more interfaces / routes to make it usable.

Magnus-Kuhn: My original idea was that the backbone would send a different error in this case, I think a 403 would be appropriate

jkoenig134: yes, we should request that.

jkoenig134: Also for example a route like /canAccess could be good.

Magnus-Kuhn: @tnotheis also had the idea to not differentiate between no token and not correctly personalized token for security purposes. Also why do we offer multiple ways to load tokens?

tnotheis: This would also be in line with what we currently have for other objects. For example, if you have the ID of an existing message of which you are not a participant (-> neither sender nor recipient), you will receive a 404 without a specific reason.

jkoenig134: yeah, but for us it's sadly unusable.

tnotheis: Why?

jkoenig134: b/c we can not determine why we cannot access it. Why not just return the token but without content when forIdentity is incorrect? Then we can figure out in the app if we have the right identity available to do the operation.

Magnus-Kuhn: The app will be able to use the adapted truncated reference once it's implemented, so we don't have to change things for the app if we can wait

jkoenig134: I have a bad feeling about relying completely on truncated references with the current impl. Maybe we should remove the UseCases that allow to load by id and key?

tnotheis: > b/c we can not determine why we cannot access it. 

So if I understand you correctly, the use case is: 
"The user scans a QR code of a RealtionshipTemplate with `forIdentity=A` and tries to open it with identity `B`."
And you would like to show an error message like "The Relationship Template you are trying to open belongs to another identity". 
If that's the case, you could also show something like "The Relationship Template you are trying to open either does not exist or belongs to another identity". 

> Why not just return the token but without content when forIdentity is incorrect? Then we can figure out in the app if we have the right identity available to do the operation.

Because of the two reasons that are already mentioned above:
1. It would be inconsistent to how we do it for other entities like Messages or Relationships.
2. It would leak some information even though it's not 100% necessary, which might lead to security "problems".


Don't get me wrong, I'm not 100% against it. I just don't want to change it now and have to undo it a few days later because you realize there's a better way.

jkoenig134: > Since I don't know a use for loading by id and key, I'm in favour of it :)

@Magnus-Kuhn let's discuss this elsewhere and get back to Timo after that. I think this could be the solution, but I could be wrong.

jkoenig134: @tnotheis thanks for clarifying again :)

packages/transport/src/modules/tokens/TokenController.ts
sebbi08: Should it go to the DB if it is not intended for you ?

.dev/compose.backbone.env
```diff
@@ -1 +1 @@
-BACKBONE_VERSION=6.6.1
+BACKBONE_VERSION=6.7.2
```
jkoenig134, line 1: if we bump we should bump properly: [`6.10.0`](https://github.com/nmshd/backbone/releases/tag/6.10.0) is out

packages/runtime/test/anonymous/tokens.test.ts
```diff
@@ -1,13 +1,14 @@
...
 const serviceProvider = new RuntimeServiceProvider();
 let transportServices: TransportServices;
 let noLoginRuntime: TestRuntime;
+let runtimeService: TestRuntimeServices;
 
 beforeAll(async () => {
-    const runtimeServices = await serviceProvider.launch(1);
-    transportServices = runtimeServices[0].transport;
+    runtimeService = (await serviceProvider.launch(1))[0];
+    transportServices = runtimeService.transport;
```
jkoenig134, line 11: you should remove the transportServices if you expose the whole RuntimeServices.

packages/runtime/test/anonymous/tokens.test.ts
```diff
@@ -33,4 +34,12 @@ describe("Anonymous tokens", () => {
         const token = result.value;
         expect(token.content).toStrictEqual(uploadedToken.content);
     });
+
+    test("should catch a personalized token", async () => {
+        const uploadedPersonalizedToken = await uploadOwnToken(transportServices, runtimeService.address);
```
jkoenig134, line 39: wait - this is possible? You can personalize the token to yourself??

packages/runtime/test/anonymous/tokens.test.ts
```diff
@@ -33,4 +34,12 @@ describe("Anonymous tokens", () => {
...
+        const resultNotIntended = await noLoginRuntime.anonymousServices.tokens.loadPeerToken({
```
jkoenig134, line 40: `resultNotIntended` is a really strange name, what do you want to say by that?

packages/runtime/test/transport/tokens.test.ts
```diff
@@ -1,13 +1,17 @@
...
+let runtimeServices2: TestRuntimeServices;
 let transportServices1: TransportServices;
 let transportServices2: TransportServices;
 
 beforeAll(async () => {
     const runtimeServices = await serviceProvider.launch(2);
+    runtimeServices1 = runtimeServices[0];
+    runtimeServices2 = runtimeServices[1];
     transportServices1 = runtimeServices[0].transport;
     transportServices2 = runtimeServices[1].transport;
```
jkoenig134, line 16: there should not be runtimeServices **and** runtimeServices. See above.

packages/transport/src/modules/tokens/AnonymousTokenController.ts
```diff
@@ -18,11 +18,15 @@ export class AnonymousTokenController {
         return await this.loadPeerTokenByReference(reference);
     }
 
-    public async loadPeerTokenByReference(tokenReference: TokenReference): Promise<Token> {
-        return await this.loadPeerToken(tokenReference.id, tokenReference.key);
+    private async loadPeerTokenByReference(tokenReference: TokenReference): Promise<Token> {
```
jkoenig134, line 21: why is this private? This should be the default now...

jkoenig134: you can also just move the content of this method to loadPeerTokenByTruncated

packages/transport/src/modules/tokens/TokenController.ts
```diff
@@ -186,12 +189,16 @@ export class TokenController extends TransportController {
...
+        return await this.loadPeerToken(tokenReference.id, tokenReference.key, ephemeral, tokenReference.forIdentityTruncated);
     }
 
-    public async loadPeerToken(id: CoreId, secretKey: CryptoSecretKey, ephemeral: boolean): Promise<Token> {
+    private async loadPeerToken(id: CoreId, secretKey: CryptoSecretKey, ephemeral: boolean, forIdentityTruncated?: string): Promise<Token> {
         const tokenDoc = await this.tokens.read(id.toString());
+        if (!tokenDoc && forIdentityTruncated && !this.parent.identity.address.toString().endsWith(forIdentityTruncated)) {
+            // if you created the token, it exists already
+            throw TransportCoreErrors.general.notIntendedForYou(id.toString());
+        }
```
jkoenig134, line 201: newline

packages/transport/src/modules/tokens/TokenController.ts
```diff
@@ -186,12 +189,16 @@ export class TokenController extends TransportController {
...
-    public async loadPeerTokenByReference(tokenReference: TokenReference, ephemeral: boolean): Promise<Token> {
-        return await this.loadPeerToken(tokenReference.id, tokenReference.key, ephemeral);
+    private async loadPeerTokenByReference(tokenReference: TokenReference, ephemeral: boolean): Promise<Token> {
+        return await this.loadPeerToken(tokenReference.id, tokenReference.key, ephemeral, tokenReference.forIdentityTruncated);
     }
 
-    public async loadPeerToken(id: CoreId, secretKey: CryptoSecretKey, ephemeral: boolean): Promise<Token> {
+    private async loadPeerToken(id: CoreId, secretKey: CryptoSecretKey, ephemeral: boolean, forIdentityTruncated?: string): Promise<Token> {
         const tokenDoc = await this.tokens.read(id.toString());
+        if (!tokenDoc && forIdentityTruncated && !this.parent.identity.address.toString().endsWith(forIdentityTruncated)) {
```
jkoenig134, line 198: why do you even query the db?

packages/transport/src/modules/tokens/local/Token.ts
```diff
@@ -50,7 +50,7 @@ export class Token extends CoreSynchronizable implements IToken {
     }
 
     public toTokenReference(): TokenReference {
-        return TokenReference.from({ id: this.id, key: this.secretKey });
+        return TokenReference.from({ id: this.id, key: this.secretKey, forIdentityTruncated: this.cache?.forIdentity?.toString().slice(-4) });
```
jkoenig134, line 53:
```suggestion
        return TokenReference.from({ id: this.id, key: this.secretKey, forIdentityTruncated: this.cache!.forIdentity?.toString().slice(-4) });
```

packages/transport/test/modules/tokens/AnonymousTokenController.test.ts
```diff
@@ -0,0 +1,65 @@
...
+            forIdentity: sender.identity.address
+        });
+
+        const reference = sentToken.toTokenReference();
+        reference.forIdentityTruncated = undefined;
+        const truncatedReference = reference.truncate();
+
+        await expect(async () => {
+            await anonymousTokenController.loadPeerTokenByTruncated(truncatedReference);
+        }).rejects.toThrow("error.platform.recordNotFound");
```
jkoenig134, line 63:
```suggestion
        await  expect(anonymousTokenController.loadPeerTokenByTruncated(truncatedReference)).rejects.toThrow("error.platform.recordNotFound");
```

packages/transport/test/modules/tokens/AnonymousTokenController.test.ts
```diff
@@ -0,0 +1,65 @@
...
+        const sentToken = await sender.tokens.sendToken({
+            content,
+            expiresAt,
+            ephemeral: false,
+            forIdentity: sender.identity.address
+        });
+
+        await expect(async () => {
+            await anonymousTokenController.loadPeerTokenByTruncated(sentToken.toTokenReference().truncate());
+        }).rejects.toThrow("transport.general.notIntendedForYou");
```
jkoenig134, line 44: inline, see below

packages/transport/test/modules/tokens/AnonymousTokenController.test.ts
```diff
@@ -0,0 +1,65 @@
+import { IDatabaseConnection } from "@js-soft/docdb-access-abstractions";
+import { Serializable } from "@js-soft/ts-serval";
+import { CoreDate } from "@nmshd/core-types";
+import { AccountController, AnonymousTokenController, Transport } from "../../../src";
+import { TestUtil } from "../../testHelpers/TestUtil";
+
+describe("AnonymousTokenController", function () {
```
jkoenig134, line 7: you only introduced negative tests here, but no negative ones - not for personalized neither for non personalized tokens.

jkoenig134 (changes requested):

packages/runtime/test/anonymous/tokens.test.ts
Magnus-Kuhn: yes, you can

packages/runtime/test/anonymous/tokens.test.ts
Magnus-Kuhn: there was the id-and-key test before and now that it's gone, this name doesn't have any purpose anymore

packages/transport/src/modules/tokens/TokenController.ts
Magnus-Kuhn: the idea was that the personalization check will only be executed when loading the token for the first time (if it's not ephemeral)

packages/transport/test/modules/tokens/AnonymousTokenController.test.ts
Magnus-Kuhn: will add a positive test for a non-personalized token, but it's not possible to anonymously load a personalized token

packages/runtime/test/anonymous/tokens.test.ts
jkoenig134: Does not really make sense, but nice for testing :D

packages/runtime/test/anonymous/tokens.test.ts
jkoenig134: please rename it.

packages/transport/test/modules/tokens/AnonymousTokenController.test.ts
jkoenig134: right

packages/transport/src/modules/tokens/TokenController.ts
jkoenig134: I would love to see this check at the first place and 

1. check if you created the token or
2. you can access the token

It's so hard to process why you have to query the DB before doing that check.

Magnus-Kuhn: How do you want to check if you created the token without querying the DB? Not allowing you to load your own token would not be that bad but I'd like to do without it. Coding it as `const isOwnTokenOrAlreadyLoadedToken = !!tokenDoc` would also be a way to clean up

jkoenig134: I think I understand the issue now. 

We do not have access to the creator of the token if we didn't fetch the token. But we can skip the "is the token mine" step when the token is already in the DB. In this case I'm ok with this check.

packages/transport/src/modules/tokens/TokenController.ts
```diff
@@ -176,22 +178,24 @@ export class TokenController extends TransportController {
...
 
-    public async loadPeerTokenByReference(tokenReference: TokenReference, ephemeral: boolean): Promise<Token> {
-        return await this.loadPeerToken(tokenReference.id, tokenReference.key, ephemeral);
-    }
-
-    public async loadPeerToken(id: CoreId, secretKey: CryptoSecretKey, ephemeral: boolean): Promise<Token> {
+    private async loadPeerToken(id: CoreId, secretKey: CryptoSecretKey, ephemeral: boolean, forIdentityTruncated?: string): Promise<Token> {
         const tokenDoc = await this.tokens.read(id.toString());
+        if (!tokenDoc && forIdentityTruncated && !this.parent.identity.address.toString().endsWith(forIdentityTruncated)) {
+            // if you created the token, it exists already
```
jkoenig134, line 195: the comment in this line is misleading, I would 

a) completely remove it
b) put it before the if

I'd vote for a, but if you choose b please add a blank line between the comment and `const tokenDoc`

jkoenig134 (approved):

PR #250, opened 2024-08-23
Empty owner instead of explicit address for ShareAttributeRequestItem

britsta:
# Readiness checklist

-   [x] I labeled the PR.

# Description

Please note that an error message was already thrown if an empty string was specified instead of the explicit address of the owner when sharing an `attribute` with the ShareAttributeRequestItem, only that so far a misleading error message was displayed. There has been some code removed that caused this misleading error message to be thrown. In addition, a few old tests were removed that had not fulfilled their intended function anyway, as they were not sufficiently well thought out.


.ci/runChecks.sh
```diff
@@ -5,4 +5,4 @@ npm run build:node
 npm run lint:eslint
 npm run lint:prettier
 npx -ws license-check
-npx better-npm-audit audit --exclude 1096302,1098615
+npx better-npm-audit audit --exclude 1096302,1098615,1098618
```
jkoenig134, line 8: I see a misuse of audit excludes here: next time please check

1. `npm audit fix`, then
2. `npx better-npm-audit audit` (without the excludes)

this audit error *is* fixable, so we should not just exclude it.

jkoenig134: you can also wait for #208 to be merged, I already did it there.

jkoenig134 (approved):

PR #252, opened 2024-08-27
RelationshipChangedEvent is triggered even if the Relationship has not changed

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

# Description

Fixes the [AutoDecomposeDeletionProposedRelationships module](https://github.com/nmshd/connector/pull/241) trying to decompose the relationship twice.


packages/transport/test/modules/relationships/TerminateDecompose.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    });
+
+    test("should only get one RelationshipChangedEvent for Terminate and Decompose at the same time", async function () {
+        await sender.relationships.terminate(relationshipId);
+        await sender.relationships.decompose(relationshipId);
+
+        const events: RelationshipChangedEvent[] = [];
+        transport.eventBus.subscribe(RelationshipChangedEvent, function (event) {
+            events.push(event);
+        });
```
Milena-Czierlinski, line 42: Is the order of initiating events and subscribing to then irrelevant?

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
```diff
@@ -0,0 +1,49 @@
...
+
+        const relationships = await TestUtil.syncUntilHasRelationships(recipient);
+        expect(relationships).toHaveLength(1);
+
```
Milena-Czierlinski, line 46: Why is this here?

packages/transport/src/modules/sync/externalEventProcessors/RelationshipExternalEventProcessor.ts
```diff
@@ -0,0 +1,17 @@
...
+import { ExternalEventProcessor } from "./ExternalEventProcessor";
+
+export abstract class RelationshipExternalEventProcessor extends ExternalEventProcessor {
+    protected triggerRelationshipChangedEvent(changedRelationship: Relationship, oldRelationship?: Relationship): void {
+        if (!this.didChangeRelationship(changedRelationship, oldRelationship)) return;
+
+        this.eventBus.publish(new RelationshipChangedEvent(this.ownAddress, changedRelationship));
+    }
+
+    private didChangeRelationship(changedRelationship: Relationship, oldRelationship?: Relationship): boolean {
```
Milena-Czierlinski, line 12: The function name makes me think of dids. 😆 Maybe rename it to `changedRelationship` or `relationshipWasChanges` (however this doesn't sound as good with `this.` before) or something else. Or just leave it if you don't consider it to be important.

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -487,18 +487,23 @@ export class RelationshipsController extends TransportController {
...
-            if (newRelationship.status === RelationshipStatus.Pending) {
-                return newRelationship;
-            }
+            const changedRelationship = await this.createNewRelationshipByIncomingCreation(relationshipId);
+            if (changedRelationship.status === RelationshipStatus.Pending) return { changedRelationship };
+
             // this path is for a revocation that is processed before its corresponding creation
             relationshipDoc = await this.relationships.read(relationshipId);
+
+            const updatedRelationship = await this.updateRelationshipWithPeerResponse(relationshipDoc);
```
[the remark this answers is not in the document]
jkoenig134: not part of this PR.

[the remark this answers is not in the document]
britsta: I thought it belonged here, because it would influence whether you might have to return an oldRelationship in addition to the changedRelationship.

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
```diff
@@ -0,0 +1,49 @@
...
+        relationshipId = relationships.acceptedRelationshipFromSelf.id;
+    });
+
+    afterAll(async function () {
+        await sender.close();
+        await recipient.close();
+        await connection.close();
+    });
+
+    test("should only get one RelationshipChangedEvent for Terminate and Decompose at the same time", async function () {
```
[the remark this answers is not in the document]
jkoenig134: this test only tests one behaviour and I will **not** add more expects that will hide the actual test case. Termination and decomposition is already tested extensively elsewhere.

> how an Identity get the change of the status of the Relationship from 'Active' to 'DeletionProposed'

the only thing we do in the processors is fetching the current state from the backbone. If we have 5 RelationshipChanged external events we will do the same thing 5 times.

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
jkoenig134: right, this should not be there :)

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
jkoenig134: what?

jkoenig134: I have to subscribe to events BEFORE they will be triggered, so yeah - the order is very relevant :D

packages/transport/src/modules/sync/externalEventProcessors/RelationshipExternalEventProcessor.ts
jkoenig134: changed it

Milena-Czierlinski: Sure, I just stumbled across this. Also, because one wouldn't use "did" here and just use past tense, unless to put an emphasize on this. But if that is what you're aiming at, it's all good.

packages/transport/src/modules/relationships/RelationshipsController.ts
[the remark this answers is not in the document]
jkoenig134: this is why I HATE comments ..

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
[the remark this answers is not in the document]
britsta: All right, I now understand that if I have a Relationship in the `'Active'` status and the Relationship in the Backbone is already in the `'DeletionProposed'` status, it only makes sense to receive the RelationshipChangedEvent once and not twice, as it does not provide any additional information.

Nevertheless, I don't understand how we deal with the Relationship going directly to the `'DeletionProposed'` status from the `'Active'` status in the Runtime.

packages/transport/src/modules/relationships/RelationshipsController.ts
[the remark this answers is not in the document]
Milena-Czierlinski: @britsta As I understand this, this is for the case that the peer sends you a Relationship Request (Pending Relationship) and revokes it before you received the pending Relationship. In this case, when you fetch you'll only get the revoked Relationship and never had the "old" pending Relationship.

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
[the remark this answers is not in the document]
jkoenig134: we have no alternative than dealing with a direct status change as we don't get status changes that happened in between.

packages/transport/src/modules/relationships/RelationshipsController.ts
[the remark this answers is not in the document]
britsta: > @britsta As I understand this, this is for the case that the peer sends you a Relationship Request (Pending Relationship) and revokes it before you received the pending Relationship. In this case, when you fetch you'll only get the revoked Relationship and never had the "old" pending Relationship.

Ohh, I see, that could be...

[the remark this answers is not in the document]
jkoenig134: exactly.

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
Milena-Czierlinski: Ah and the relevant call is the sync and not the terminate/decompose, right?

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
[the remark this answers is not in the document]
britsta: Makes sense, so both `auditLog` entries (from `'Active'` to `'Terminated'` and from `'Terminated'` to `'DeletionProposed'`) are added together directly?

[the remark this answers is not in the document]
jkoenig134: yep

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
Milena-Czierlinski: However, why did the test fail here before and returned 2 relationships? 👀

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
jkoenig134: all (good) tests have the same setup `prepare - act - validate`, this is done exactly like this.

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
jkoenig134: it did? I don't care.

jkoenig134: ah, 2 events so 2 "changed" relationships

jkoenig134: also a "bug" but not part of this PR.

packages/transport/test/modules/relationships/TerminateDecompose.test.ts
Milena-Czierlinski: So that's a yes? The events that are subscribed to arrive when the sync is done and not when the sender terminates the Relationship, don't they?

jkoenig134: I think so, yes?

Milena-Czierlinski: According to your first answer, this should be the case. Otherwise we wouldn't receive the events. I was just wondering why you subscribed to the event bus **after** the relationship is terminated and decomposed.

jkoenig134: I am so confused of this conversation :D

Milena-Czierlinski: Ok, never mind. The tests are green so everything seems to be working as it should. :D

Milena-Czierlinski (approved):

PR #254, opened 2024-08-28
Move core types to extra package

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

# Description / Motivation

As the Core* types were placed in transport before we could not re-use them in e.g. the connector sdk. With an extra package that transport and content can depend on we now have the ability to re-use content types.


packages/transport/src/core/CoreIdHelper.ts
Milena-Czierlinski, line 1: Why is this not in the core-types package?

packages/app-runtime/src/multiAccount/data/LocalAccount.ts
```diff
@@ -1,7 +1,7 @@
-import { serialize, type, validate } from "@js-soft/ts-serval";
-import { CoreAddress, CoreDate, CoreId, CoreSerializable, ICoreDate, ICoreSerializable } from "@nmshd/transport";
+import { ISerializable, Serializable, serialize, type, validate } from "@js-soft/ts-serval";
+import { CoreAddress, CoreDate, CoreId, ICoreDate } from "@nmshd/core-types";
 
-export interface ILocalAccount extends ICoreSerializable {
+export interface ILocalAccount extends ISerializable {
```
Milena-Czierlinski, line 4: Why did you change this in some places?

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
```diff
@@ -8,10 +8,11 @@ import {
     RelationshipAttribute,
     RelationshipAttributeJSON
 } from "@nmshd/content";
-import { CoreAddress, CoreDate, CoreId, CoreSynchronizable, ICoreDate, ICoreId, ICoreSynchronizable } from "@nmshd/transport";
+import { CoreAddress, CoreDate, CoreId, ICoreDate, ICoreId } from "@nmshd/core-types";
+import { CoreSynchronizable, ICoreSynchronizable } from "@nmshd/transport";
```
Milena-Czierlinski, line 12: Can CoreSynchronizable (and CoreSerializable) also move to core-types?

Milena-Czierlinski (commented):
Do the (app-)runtime tsconfig files need `"references": [{ "path": "../core-types" }]` as well?

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
jkoenig134: CoreSerializable is a utility type that would bring crypto to the core-types library which we dont want. The long-running plan from my side is to remove this type completely. This will introduce more changes so I will do this in a next iteration.

CoreSynchronizable is a data type strictly coupeled to how the transport sync works so I have not moved this.

packages/app-runtime/src/multiAccount/data/LocalAccount.ts
jkoenig134: explained in other comment

packages/transport/src/core/CoreIdHelper.ts
jkoenig134: because it would **kill** the whole idea of the core-types package. CoreIdHelper is no type and would bring crypto to core-types.

jkoenig134:
No, things are already referenced in library below.

packages/core-types/src/CoreDate.ts
```diff
@@ -138,17 +136,17 @@ export class CoreDate extends CoreSerializable {
     }
 
     private get asValidDateTime(): DateTime<true> {
-        if (!this.dateTime.isValid) throw new TransportError("The date is invalid.");
+        if (!this.dateTime.isValid) throw new Error("The date is invalid.");
```
Milena-Czierlinski, line 139: Would it make sense to have a CoreError here?

jkoenig134: no, I had this, but then removed it.

packages/consumption/src/modules/attributes/local/LocalAttribute.ts
jkoenig134: To not have more confusion I remove CoreSerializable completely.

packages/core-types/src/CoreDate.ts
jkoenig134: To give more context. We already have a CoreError, but this has code, message, ... - this would be too much for this class.

Milena-Czierlinski (approved):

PR #257, opened 2024-08-29
Add core-types readme

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ This monorepo consolidates the following packages:
 
 ## Documentation
 
-You can find the documentation for Enmeshed on [enmeshed.eu](https://enmeshed.eu).
+You can find the documentation for enmeshed on [enmeshed.eu](https://enmeshed.eu).
 
 To get started developing in this repository, see the [developer's guide](README_dev.md).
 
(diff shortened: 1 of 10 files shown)


packages/core-types/README.md
```diff
@@ -0,0 +1,25 @@
+# Core Types
+
+[![npm version](https://badge.fury.io/js/@nmshd%2fcore-types.svg)](https://www.npmjs.com/package/@nmshd/core-types)
+
+The core-types library provides common types for the transport and content library.
+
+## Documentation
+
+You can find the documentation for Enmeshed on [enmeshed.eu](https://enmeshed.eu).
```
Milena-Czierlinski, line 9: enmeshed not capitalized 🤔 (also below)

README.md
```diff
@@ -6,6 +6,7 @@ This monorepo consolidates the following packages:
 
 | Package                              | Version                                                                                                           |
 | :----------------------------------- | :---------------------------------------------------------------------------------------------------------------- |
+| [core-types](packages/core-types/)   | [![version](https://badge.fury.io/js/@nmshd%2fcore-types.svg)](https://www.npmjs.com/package/@nmshd/core-types)   |
```
Milena-Czierlinski, line 9: Should we really keep the version? It should be the same for all libraries, so there is no actual information that this column adds.

packages/core-types/README.md
jkoenig134: copy paste from other places.. but ok

README.md
jkoenig134: it looks shitty without. so yes.

Milena-Czierlinski (approved):

PR #266, opened 2024-09-10
Simplify Decider module

Milena-Czierlinski:
# Readiness checklist

-   [x] I labeled the PR.

# Description

This PR contains some refactorings I came across working on automating Request that should not clutter the actual PR.

### DIFF
--- a/packages/runtime/src/modules/DeciderModule.ts
+++ b/packages/runtime/src/modules/DeciderModule.ts
@@ -38,14 +38,13 @@ export class DeciderModule extends RuntimeModule {
             return;
         }
 
-        await this.publishEvent(event, services, "ManualRequestDecisionRequired", request.id);
+        await this.publishEvent(event, services, "ManualRequestDecisionRequired");
     }
 
     private async publishEvent(
         event: IncomingRequestStatusChangedEvent,
         services: RuntimeServices,
-        result: keyof typeof RelationshipTemplateProcessedResult & keyof typeof MessageProcessedResult,
-        requestId?: string
+        result: keyof typeof RelationshipTemplateProcessedResult & keyof typeof MessageProcessedResult
     ) {
         const request = event.data.request;
         switch (request.source!.type) {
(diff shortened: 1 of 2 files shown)


packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -63,13 +62,11 @@ export class DeciderModule extends RuntimeModule {
...
 
                 if (result === "ManualRequestDecisionRequired") {
-                    if (!requestId) throw new Error("Request ID is required for manual decision required result.");
-
                     this.runtime.eventBus.publish(
                         new RelationshipTemplateProcessedEvent(event.eventTargetAddress, {
                             template,
                             result: result as RelationshipTemplateProcessedResult.ManualRequestDecisionRequired,
-                            requestId
+                            requestId: request.id
```
jkoenig134, line 69: undo, and re-add the above

jkoenig134 (changes requested):

packages/runtime/src/modules/DeciderModule.ts
Milena-Czierlinski: But why?

jkoenig134: and you can check that the request is valid

Milena-Czierlinski: Can you please elaborate on how the Request's validity was checked before in contrast to now?

Milena-Czierlinski: From my point of view this only checks if the function is used correctly. With my change, there no longer is the risk to use it incorrectly. Also, I think readability actually improves, since an unnecessary parameter is removed.

jkoenig134 (dismissed):

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -38,16 +38,16 @@ export class DeciderModule extends RuntimeModule {
...
     private async publishEvent(
         event: IncomingRequestStatusChangedEvent,
         services: RuntimeServices,
-        result: keyof typeof RelationshipTemplateProcessedResult & keyof typeof MessageProcessedResult,
-        requestId?: string
+        result: keyof typeof RelationshipTemplateProcessedResult & keyof typeof MessageProcessedResult
     ) {
         const request = event.data.request;
+        const requestId = request.id;
         switch (request.source!.type) {
```
jkoenig134:
```suggestion
        const requestId = request.id;

        switch (request.source!.type) {
```

packages/runtime/src/modules/DeciderModule.ts
jkoenig134: yeah 👯

PR #268, opened 2024-09-11
Add correlation ID to all use cases and send it to backbone

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/packages/transport/test/core/backbone/CorrelationId.test.ts
+++ b/packages/transport/test/core/backbone/CorrelationId.test.ts
@@ -23,7 +23,7 @@ describe("CorrelationId", function () {
         await connection.close();
     });
 
-    test("should send correlation id to the backend when given", async function () {
+    test("should send correlation id to the backbone when given", async function () {
         interceptor.start();
         await correlationIdLib.withId("test-correlation-id", async () => {
             await testAccount.syncEverything();
--- a/packages/transport/test/core/backbone/CorrelationId.test.ts
+++ b/packages/transport/test/core/backbone/CorrelationId.test.ts
@@ -1,29 +1,33 @@
+import { IDatabaseConnection } from "@js-soft/docdb-access-abstractions";
 import correlationIdLib from "correlation-id";
+import { AccountController } from "../../../src";
 import { RequestInterceptor } from "../../testHelpers/RequestInterceptor";
 import { TestUtil } from "../../testHelpers/TestUtil";
 
 describe("CorrelationId", function () {
-    test("should send correlation id to the backend when given", async function () {
-        // Arrange
-        const connection = await TestUtil.createDatabaseConnection();
+    let connection: IDatabaseConnection;
+    let testAccount: AccountController;
+    let interceptor: RequestInterceptor;
+    beforeAll(async function () {
+        connection = await TestUtil.createDatabaseConnection();
         const transport = TestUtil.createTransport(connection);
         await transport.init();
         const accounts = await TestUtil.provideAccounts(transport, 1);
-        const testAccount = accounts[0];
-        const interceptor = new RequestInterceptor((testAccount as any).synchronization.client);
+        testAccount = accounts[0];
+        interceptor = new RequestInterceptor((testAccount as any).synchronization.client);
+    });
 
-        // Act
+    afterAll(async function () {
+        await testAccount.close();
+        await connection.close();
+    });
+    test("should send correlation id to the backend when given", async function () {
         interceptor.start();
         await correlationIdLib.withId("test-correlation-id", async () => {
             await testAccount.syncEverything();
         });
 
-        // Assert
         const requests = interceptor.stop().requests;
         expect(requests.at(-1)!.headers!["x-correlation-id"]).toBe("test-correlation-id");
-
-        // Cleanup
-        await testAccount.close();
-        await connection.close();
     });
 });
--- a/packages/transport/test/core/backbone/CorrelationId.test.ts
+++ b/packages/transport/test/core/backbone/CorrelationId.test.ts
@@ -1,0 +1,29 @@
+import correlationIdLib from "correlation-id";
+import { RequestInterceptor } from "../../testHelpers/RequestInterceptor";
+import { TestUtil } from "../../testHelpers/TestUtil";
+
+describe("CorrelationId", function () {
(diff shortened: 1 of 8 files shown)


Milena-Czierlinski:
I'm not sure if adding a newline actually counts as updating a test. :D Is there a way we can truly test this feature?

jkoenig134 (approved):

jkoenig134:
> Is there a way we can truly test this feature?

there should be a way. But I think it would require some mocking of the rest client.

sebbi08:
I added some test but forgot to add the file :/

packages/transport/test/core/backbone/CorrelationId.test.ts
```diff
@@ -0,0 +1,29 @@
+import correlationIdLib from "correlation-id";
+import { RequestInterceptor } from "../../testHelpers/RequestInterceptor";
+import { TestUtil } from "../../testHelpers/TestUtil";
+
+describe("CorrelationId", function () {
+    test("should send correlation id to the backend when given", async function () {
+        // Arrange
```
jkoenig134, line 7: comments?

packages/transport/test/core/backbone/CorrelationId.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        const connection = await TestUtil.createDatabaseConnection();
+        const transport = TestUtil.createTransport(connection);
+        await transport.init();
+        const accounts = await TestUtil.provideAccounts(transport, 1);
+        const testAccount = accounts[0];
+        const interceptor = new RequestInterceptor((testAccount as any).synchronization.client);
```
jkoenig134, line 13: what happened to beforeAll?

packages/transport/test/core/backbone/CorrelationId.test.ts
```diff
@@ -0,0 +1,29 @@
...
+            await testAccount.syncEverything();
+        });
+
+        // Assert
+        const requests = interceptor.stop().requests;
+        expect(requests.at(-1)!.headers!["x-correlation-id"]).toBe("test-correlation-id");
+
+        // Cleanup
+        await testAccount.close();
+        await connection.close();
```
jkoenig134, line 27: for cleanup we have afterEach / afterAll

packages/transport/test/core/backbone/CorrelationId.test.ts
```diff
@@ -1,29 +1,33 @@
...
 import { TestUtil } from "../../testHelpers/TestUtil";
 
 describe("CorrelationId", function () {
-    test("should send correlation id to the backend when given", async function () {
-        // Arrange
-        const connection = await TestUtil.createDatabaseConnection();
+    let connection: IDatabaseConnection;
+    let testAccount: AccountController;
+    let interceptor: RequestInterceptor;
+    beforeAll(async function () {
```
jkoenig134, line 11:
```suggestion
    let interceptor: RequestInterceptor;

    beforeAll(async function () {
```

packages/transport/test/core/backbone/CorrelationId.test.ts
```diff
@@ -0,0 +1,33 @@
...
+        const accounts = await TestUtil.provideAccounts(transport, 1);
+        testAccount = accounts[0];
+        interceptor = new RequestInterceptor((testAccount as any).synchronization.client);
+    });
+
+    afterAll(async function () {
+        await testAccount.close();
+        await connection.close();
+    });
+    test("should send correlation id to the backend when given", async function () {
```
britsta, line 24: Writing "backbone" instead of "backend" as you do in the PR title should be more precise.

packages/transport/test/core/backbone/CorrelationId.test.ts
```diff
@@ -1,29 +1,33 @@
...
+        testAccount = accounts[0];
+        interceptor = new RequestInterceptor((testAccount as any).synchronization.client);
+    });
 
-        // Act
+    afterAll(async function () {
+        await testAccount.close();
+        await connection.close();
+    });
+    test("should send correlation id to the backend when given", async function () {
```
jkoenig134, line 24:
```suggestion
    });

    test("should send correlation id to the backend when given", async function () {
```

jkoenig134 (changes requested):

packages/transport/test/core/backbone/CorrelationId.test.ts
```diff
@@ -0,0 +1,33 @@
...
+import correlationIdLib from "correlation-id";
+import { AccountController } from "../../../src";
+import { RequestInterceptor } from "../../testHelpers/RequestInterceptor";
+import { TestUtil } from "../../testHelpers/TestUtil";
+
+describe("CorrelationId", function () {
+    let connection: IDatabaseConnection;
+    let testAccount: AccountController;
+    let interceptor: RequestInterceptor;
+    beforeAll(async function () {
```
britsta, line 11: Please add an empty line before `beforeAll`.

packages/transport/test/core/backbone/CorrelationId.test.ts
britsta, line 24: Please add an empty line after `afterAll`.

packages/transport/test/core/backbone/CorrelationId.test.ts
britsta: Julian was faster :D

jkoenig134: :)

jkoenig134:
@sebbi08 we don't need correlation IDs for every request? My hope would be to have one in the whole runtime scope to also be able to use it in the app.

sebbi08:
> @sebbi08 we don't need correlation IDs for every UseCase? My hope would be to have one in the whole runtime scope to also be able to use it in the app.

Ah, i see. The problem I see with this is that we cannot distinguish between the different actions of a user in one session, as all will have the same. Or are you creating a new runtime for each action?

jkoenig134:
I don't, but why do we have to create a new runtime when we create correlation id's for each usecase?

jkoenig134 (approved):

jkoenig134 (changes requested):
I have to check something within the correlation-id lib

PR #269, opened 2024-09-12
Extend DeciderModule to automatically decide Requests

Milena-Czierlinski:
# Readiness checklist

-   [x] I labeled the PR.


packages/app-runtime/src/modules/appEvents/RelationshipTemplateProcessedModule.ts
```diff
@@ -21,6 +21,7 @@ export class RelationshipTemplateProcessedModule extends AppRuntimeModule<Relati
 
         const data = event.data;
         switch (data.result) {
+            case RelationshipTemplateProcessedResult.RequestAutomaticallyDecided:
```
Milena-Czierlinski, line 24: Is this what we want here?

jkoenig134: no, in this case we want to hide the request from the user. So just noop here.

packages/runtime/test/lib/testUtils.ts
```diff
@@ -61,6 +61,45 @@ import {
...
+export async function expectThrowsAsync(method: Function | Promise<any>, customExceptionMatcher?: (e: Error) => void): Promise<void>;
+export async function expectThrowsAsync(method: Function | Promise<any>, errorMessagePatternOrRegexp: RegExp): Promise<void>;
+/**
+ *
+ * @param method The function which should throw the exception
+ * @param errorMessagePattern the pattern the error message should match (asterisks ('\*') are wildcards that correspond to '.\*' in regex)
+ */
+export async function expectThrowsAsync(method: Function | Promise<any>, errorMessagePattern: string): Promise<void>;
+
+export async function expectThrowsAsync(method: Function | Promise<any>, errorMessageRegexp: RegExp | string | ((e: Error) => void) | undefined): Promise<void> {
```
jkoenig134, line 73: we don't need this as it's already in the standard set of jest. see https://jestjs.io/docs/expect#rejects

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -22,18 +75,261 @@ export class DeciderModule extends RuntimeModule {
     private async handleIncomingRequestStatusChanged(event: IncomingRequestStatusChangedEvent) {
         if (event.data.newStatus !== LocalRequestStatus.DecisionRequired) return;
 
-        if (event.data.request.content.items.some(flaggedAsManualDecisionRequired)) return await this.requireManualDecision(event);
+        const requestContent = event.data.request.content;
+        if (this.containsItem(requestContent, (item) => item["requireManualDecision"] === true)) {
+            return await this.requireManualDecision(event);
```
Magnus-Kuhn, line 80: That the setting for whether a request can be decided automatically is not made on the entire request seems kinda strange

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -22,18 +75,261 @@ export class DeciderModule extends RuntimeModule {
...
+        }
+
+        const automationResult = await this.automaticallyDecideRequest(event);
+        if (automationResult.isSuccess) {
+            const services = await this.runtime.getServices(event.eventTargetAddress);
+            await this.publishEvent(event, services, "RequestAutomaticallyDecided");
```
Magnus-Kuhn, line 86: You can put the automationResult in here

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -22,18 +75,261 @@ export class DeciderModule extends RuntimeModule {
...
+        const nestedProperty = path.split(".").reduce((currentObject, key) => currentObject?.[key], object);
+        return nestedProperty;
+    }
+
+    private checkTagCompatibility(requestConfigTags: string[], requestTags: string[]): boolean {
+        const atLeastOneMatchingTag = requestConfigTags.some((tag) => requestTags.includes(tag));
+        return atLeastOneMatchingTag;
+    }
+
+    private createDecideRequestItemParametersForGeneralResponseConfig(event: IncomingRequestStatusChangedEvent, responseConfigElement: ResponseConfig): Result<{ items: any[] }> {
```
Magnus-Kuhn, line 206: How could this fail in a way that you use Result.fail?

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -22,18 +75,261 @@ export class DeciderModule extends RuntimeModule {
...
+        const itemsOfRequest = request.content.items;
+
+        let decideRequestItemParameters = this.createResponseItemsWithSameDimension(itemsOfRequest, undefined);
+
+        for (const automationConfigElement of this.configuration.automationConfig) {
+            const requestConfigElement = automationConfigElement.requestConfig;
+            const responseConfigElement = automationConfigElement.responseConfig;
+
+            if (isGeneralRequestConfig(requestConfigElement)) {
+                const generalRequestIsCompatible = this.checkCompatibility(requestConfigElement, request);
```
Magnus-Kuhn, line 106: you can move the general compatibility check outside of the if-block by using the checkGeneralCompatibility

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -22,18 +75,261 @@ export class DeciderModule extends RuntimeModule {
...
+                if (decideRequestItemParameterResult.isError) {
+                    continue;
+                }
+
+                const decideRequestResult = await this.decideRequest(event, decideRequestItemParameterResult.value);
+                return decideRequestResult;
+            }
+
+            if (isRequestItemDerivationConfig(requestConfigElement)) {
+                const checkCompatibilityResult = this.checkRequestItemCompatibilityAndApplyReponseConfig(
```
Magnus-Kuhn, line 121: The name feels misleading since you also check the general request compatibility

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -22,18 +75,261 @@ export class DeciderModule extends RuntimeModule {
...
+        });
+    }
+
+    private checkRequestItemCompatibilityAndApplyReponseConfig(
+        itemsOfRequest: (RequestItemJSONDerivations | RequestItemGroupJSON)[],
+        parametersToDecideRequest: any,
+        request: LocalRequestDTO,
+        requestConfigElement: RequestItemDerivationConfig,
+        responseConfigElement: ResponseConfig
+    ): Result<{ items: any[] }> {
```
Magnus-Kuhn, line 272: Is there a way that this returns Result.fail?

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -22,18 +75,261 @@ export class DeciderModule extends RuntimeModule {
...
+                    request,
+                    requestConfigElement,
+                    responseConfigElement
+                );
+                if (checkCompatibilityResult.isError) {
+                    continue;
+                }
+
+                decideRequestItemParameters = checkCompatibilityResult.value;
+                if (!this.containsItem(decideRequestItemParameters, (element) => element === undefined)) {
```
Magnus-Kuhn, line 133: Could just be me, but it feels cleaner to separate the filling of the decideRequestItemParameters and the actual deciding

packages/runtime/test/modules/DeciderModule.test.ts
```diff
@@ -1,126 +1,2900 @@
...
+                });
+                await recipient.consumption.incomingRequests.checkPrerequisites({ requestId: receivedRequestResult.value.id });
+
+                await expect(recipient.eventBus).toHavePublished(
+                    MessageProcessedEvent,
+                    (e) => e.data.result === MessageProcessedResult.ManualRequestDecisionRequired && e.data.message.id === message.id
+                );
+            });
+
+            test("cannot decide a Request given a config with not fitting general and fitting RequestItem-specific elements", async () => {
```
Magnus-Kuhn, line 2208: This test and the one below have their names swapped

packages/runtime/test/modules/DeciderModule.test.ts
```diff
@@ -1,126 +1,2900 @@
...
+                });
+                await recipient.consumption.incomingRequests.checkPrerequisites({ requestId: receivedRequestResult.value.id });
+
+                await expect(recipient.eventBus).toHavePublished(
+                    MessageProcessedEvent,
+                    (e) => e.data.result === MessageProcessedResult.ManualRequestDecisionRequired && e.data.message.id === message.id
+                );
+            });
+
+            test("cannot decide a Request with RequestItemGroup if there is no fitting RequestItemConfig for every RequestItem", async () => {
```
Magnus-Kuhn, line 2783: this test belongs to the RequestItemGroup section

packages/runtime/test/modules/DeciderModule.test.ts
```diff
@@ -1,126 +1,2900 @@
...
+
+                const requestAfterAction = (await recipient.consumption.incomingRequests.getRequest({ id: receivedRequestResult.value.id })).value;
+                expect(requestAfterAction.status).toStrictEqual(LocalRequestStatus.Decided);
+                expect(requestAfterAction.response).toBeDefined();
+
+                const responseContent = requestAfterAction.response!.content;
+                expect(responseContent.result).toBe(ResponseResult.Rejected);
+            });
+
+            test("cannot decide a Request if there is no fitting RequestItemConfig for every RequestItem", async () => {
```
Magnus-Kuhn, line 2741: this test belongs to the RequestItemGroup section

packages/runtime/test/modules/DeciderModule.test.ts
```diff
@@ -1,126 +1,2900 @@
...
+                await recipient.consumption.incomingRequests.checkPrerequisites({ requestId: receivedRequestResult.value.id });
+
+                await expect(recipient.eventBus).toHavePublished(
+                    MessageProcessedEvent,
+                    (e) => e.data.result === MessageProcessedResult.ManualRequestDecisionRequired && e.data.message.id === message.id
+                );
+            });
+        });
+
+        test("should throw an error if the automationConfig is invalid", async () => {
```
Magnus-Kuhn, line 2881: this one also doesn't have multiple automationConfig elements

packages/runtime/test/modules/DeciderModule.test.ts
Magnus-Kuhn, line 1: You win the longest test file award 😵‍💫

packages/runtime/test/modules/DeciderModule.test.ts
Milena-Czierlinski: It's actually outside of the description block. :)

packages/runtime/src/modules/DeciderModule.ts
Magnus-Kuhn: this was my original idea (at the cost of performance) since it's then possible to break after processing a general request config. But I'd now rather move decideRequest out of the if-block but still keep it inside the for-block - and at the same time change the general request config logic to continue processing more configs after them (as stated in a different comment).

packages/app-runtime/src/modules/appEvents/RelationshipTemplateProcessedModule.ts
```diff
@@ -21,6 +21,10 @@ export class RelationshipTemplateProcessedModule extends AppRuntimeModule<Relati
 
         const data = event.data;
         switch (data.result) {
+            case RelationshipTemplateProcessedResult.RequestAutomaticallyDecided: {
+                break;
+            }
```
jkoenig134, line 26: I would put this to the last place (it is the most irrelevant case for the reader)

packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -234,6 +234,39 @@ class IdentityDeletionProcess {
     }
 }
 
+class DeciderModule {
+    public doesNotHaveAutomationConfig() {
+        return new ApplicationError("error.runtime.decide.doesNotHaveAutomationConfig", "The Request can't be decided automatically, since no automationConfig was provided.");
+    }
```
jkoenig134, line 240: I don't understand why this is an error. Why can't we then simply let the user decide?

packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -234,6 +234,39 @@ class IdentityDeletionProcess {
...
+
+    public someItemsOfRequestCouldNotBeDecidedAutomatically() {
+        return new ApplicationError(
+            "error.runtime.decide.someItemsOfRequestCouldNotBeDecidedAutomatically",
+            "The Request couldn't be decided automatically, since it contains RequestItems for which no suitable automationConfig was provided."
+        );
+    }
```
jkoenig134, line 247: see above

jkoenig134 (commented):
some details. if @Magnus-Kuhn approves I'm ok with the PR.

packages/runtime/src/useCases/common/RuntimeErrors.ts
Milena-Czierlinski: @jkoenig134  Actually, I use all "DeciderModule RuntimeErrors" only within the DeciderModule to return error Results I can handle there and they are never returned to a user. Probably, I shouldn't list them here in the RuntimeError file and just define them within the DeciderModule, should I?

jkoenig134: if it never get's shown to the user please don't use error codes at all.

Milena-Czierlinski: I use Result.fail() which expects an ApplicationError and there a code is required.

jkoenig134: that's not good. you should just throw then

Milena-Czierlinski: But we shouldn't throw inside the DeciderModule. It must continue running and e.g. set the Request to ManualDecisionRequired. 🤔

jkoenig134: try..catch?

jkoenig134: wait for changing anything, I'll look deeper into it and come back to you.

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -22,18 +75,261 @@ export class DeciderModule extends RuntimeModule {
...
+            if (property.endsWith("tags")) {
+                compatible &&= this.checkTagCompatibility(requestConfigProperty, requestProperty);
+                if (!compatible) break;
+                continue;
+            }
+
+            if (Array.isArray(requestConfigProperty)) {
+                compatible &&= requestConfigProperty.includes(requestProperty);
+            } else {
+                compatible &&= requestConfigProperty === requestProperty;
```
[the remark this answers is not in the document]
Milena-Czierlinski: So I am still not completely sure what is the expected behavior regarding the dates. If `expiresAt`, `validFrom` or `validTo` is configured, would the integrator expect that the dates exactly match or e.g. in the case of `expiresAt` that the configured date is the lower or upper boundary? Also, a voice in my head says that nobody will use this in the near future, so as stated above maybe we don't need to implement it right now. @Magnus-Kuhn @jkoenig134 Opinions?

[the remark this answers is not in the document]
Magnus-Kuhn: I think a lot of the offered options will not be used :D - the connector didn't support this for queries even though it was advertised, I don't know if this has changed. IMO we don't have to implement it, it won't be missed but just look silly.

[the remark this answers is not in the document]
Magnus-Kuhn: @jkoenig134 change it or leave it as is?

packages/runtime/src/useCases/common/RuntimeErrors.ts
jkoenig134: I think I would solve this buy just returning an object in this form: `{ wasDecided: true; requestId: string } | { wasDecided: false }` we should not use Result here as it forces us to introduce errors.

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -1,18 +1,72 @@
...
+
+        for (const automationConfigElement of this.configuration.automationConfig) {
+            const isCompatible = this.validateAutomationConfig(automationConfigElement.requestConfig, automationConfigElement.responseConfig);
+            if (!isCompatible) {
+                throw RuntimeErrors.deciderModule.requestConfigDoesNotMatchResponseConfig();
+            }
+        }
+    }
+
+    public validateAutomationConfig(requestConfig: RequestConfig, responseConfig: ResponseConfig): boolean {
```
jkoenig134, line 53: why is this public?

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -22,18 +76,269 @@ export class DeciderModule extends RuntimeModule {
...
+
+        return items.some((item) => {
+            if (item?.hasOwnProperty("items")) {
+                return this.containsItem(item, callback);
+            }
+            return callback(item);
+        });
+    }
+
+    public async automaticallyDecideRequest(event: IncomingRequestStatusChangedEvent): Promise<Result<LocalRequestDTO>> {
```
jkoenig134, line 105: why is this public?

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -22,18 +76,269 @@ export class DeciderModule extends RuntimeModule {
...
+                if (element["@type"] === "RequestItemGroup") {
+                    const responseItems = this.createEmptyDecideRequestItemParameters(element.items);
+                    return responseItems;
+                }
+                return undefined;
+            })
+        };
+    }
+
+    public checkGeneralRequestCompatibility(requestConfigElement: RequestConfig, request: LocalRequestDTO): boolean {
```
jkoenig134, line 152: ... please re-check all methods here :D

packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -22,18 +76,269 @@ export class DeciderModule extends RuntimeModule {
...
+                parametersToDecideRequest.items[i] = responseConfigElement;
+            }
+        }
+        return parametersToDecideRequest;
+    }
+
+    public checkRequestItemCompatibility(requestConfigElement: RequestItemDerivationConfig, requestItem: RequestItemJSONDerivations): boolean {
+        const requestItemPartOfConfigElement = this.filterConfigElementByPrefix(requestConfigElement, true);
+        return this.checkCompatibility(requestItemPartOfConfigElement, requestItem);
+    }
```
jkoenig134, line 290: It feels like this code should not be inside the module itself - it's really blown up now.

packages/runtime/test/modules/DeciderModule.test.ts
```diff
@@ -1,126 +1,3228 @@
...
 const runtimeServiceProvider = new RuntimeServiceProvider();
 
-let sender: TestRuntimeServices;
-let recipient: TestRuntimeServices;
+afterAll(async () => await runtimeServiceProvider.stop());
 
-beforeAll(async () => {
-    const runtimeServices = await runtimeServiceProvider.launch(2, { enableDeciderModule: true });
+describe("DeciderModule", () => {
+    describe("Unit tests", () => {
```
jkoenig134, line 61: you could extract the unit tests to a file called `DeciderModule.unit.test.ts` to make the test file less than 12371298321381 lines :D

jkoenig134: I'm not a big fan of having

1. that deeply nested describes
2. describes that explain what kind of testing is done

packages/runtime/src/modules/DeciderModule.ts
Milena-Czierlinski: For testing 😬

packages/runtime/src/modules/DeciderModule.ts
Milena-Czierlinski: @jkoenig134 I see what you mean, however I'm not sure how to properly separate what should be part of the module and what should be done outside. E.g. why didn't you mark the function `createEmptyDecideRequestItemParameters`? Furthermore, would you put the part that should be moved outside of the module just below it in the same file?

packages/runtime/src/modules/DeciderModule.ts
jkoenig134: never! ever! expose things for testing!

packages/runtime/src/modules/DeciderModule.ts
jkoenig134: because the overview is a bit strange, I've should have marked that :)

packages/runtime/src/useCases/common/RuntimeErrors.ts
Milena-Czierlinski: Actually, we don't seem to need the requestId.

packages/runtime/test/modules/DeciderModule.test.ts
Milena-Czierlinski: > 12371298321381 lines

Bye bye 😢

jkoenig134: :D

jkoenig134 (approved):
LGTM

PR #270, opened 2024-09-13
Personalized RelationshipTemplates

Magnus-Kuhn:
# Readiness checklist

-   [X] I labeled the PR.


Magnus-Kuhn:
This PR has no changes to the reference, so there's no runtime validation when loading a template with the truncated reference.

jkoenig134:
> This PR has no changes to the reference, so there's no runtime validation when loading a template with the truncated reference.

we should introduce the change to the reference in an earlier PR and not merge this or #249 before. I'd suggest to make this a draft again until we figured out what we do with truncated references.

packages/runtime/src/useCases/transport/relationshipTemplates/LoadPeerRelationshipTemplate.ts
```diff
@@ -67,11 +67,11 @@ export class LoadPeerRelationshipTemplateUseCase extends UseCase<LoadPeerRelatio
         }
 
         const content = token.cache.content;
-        return await this.loadTemplate(content.templateId, content.secretKey);
+        return await this.loadTemplate(content.templateId, content.secretKey, content.forIdentity);
     }
 
-    private async loadTemplate(id: CoreId, key: CryptoSecretKey) {
-        const template = await this.templateController.loadPeerRelationshipTemplate(id, key);
+    private async loadTemplate(id: CoreId, key: CryptoSecretKey, forIdentity?: CoreAddress) {
```
stnmtz, line 73: I do not get this one - how could you load a peer template with a different identity? I guess loading a peer template would not need the forIdentity parameter.

packages/runtime/src/useCases/common/Schemas.ts
```diff
@@ -21972,6 +21975,10 @@ export const CreateOwnRelationshipTemplateRequest: any = {
             "type": "string",
             "errorMessage": "must match ISO8601 datetime format",
             "pattern": "^([+-]?\\d{4}(?!\\d{2}\\b))((-?)((0[1-9]|1[0-2])(\\3([12]\\d|0[1-9]|3[01]))?|W([0-4]\\d|5[0-2])(-?[1-7])?|(00[1-9]|0[1-9]\\d|[12]\\d{2}|3([0-5]\\d|6[1-6])))([T\\s]((([01]\\d|2[0-3])((:?)[0-5]\\d)?|24:?00)([.,]\\d+(?!:))?)?(\\17[0-5]\\d([.,]\\d+)?)?([zZ]|([+-])([01]\\d|2[0-3]):?([0-5]\\d)?)?)?)?$"
+        },
+        "AddressString": {
+            "type": "string",
+            "pattern": "did:e:[a-zA-Z0-9.-]+:dids:[0-9a-f]{22}"
```
stnmtz, line 21981: This regexp is in my opinion not the best for returning valid domain names, as it allowes `.....`, `----` or `a-.-` as examples.

stnmtz (changes requested):

packages/runtime/src/useCases/transport/relationshipTemplates/LoadPeerRelationshipTemplate.ts
Magnus-Kuhn: This is the personalization information from the template - similar to id and key

packages/runtime/src/useCases/common/Schemas.ts
Magnus-Kuhn: Do you have one in mind already? Before we go down the rabbithole of URL regexes

Also I don't think it's crucial to get this right - (hopefully) no one will manually enter an address and for copy-paste the check is not needed

jkoenig134: +1

jkoenig134: also - IF we want to change it it would definitely not be in this PR. This AddressString was introduced months before.

packages/runtime/src/useCases/transport/relationshipTemplates/CreateTokenForOwnRelationshipTemplate.ts
```diff
@@ -42,7 +42,8 @@ export class CreateTokenForOwnTemplateUseCase extends UseCase<CreateTokenForOwnT
 
         const tokenContent = TokenContentRelationshipTemplate.from({
             templateId: template.id,
-            secretKey: template.secretKey
+            secretKey: template.secretKey,
+            forIdentity: template.cache?.forIdentity
```
jkoenig134, line 46: we should validate that this forIdentity matches the forIdentity of the Token, same for other CreateTokenFor*UseCases

packages/runtime/src/useCases/transport/relationshipTemplates/LoadPeerRelationshipTemplate.ts
jkoenig134: @Magnus-Kuhn maybe it's beneficial to remove this private method and put the content inside the other method.

packages/runtime/test/dataViews/RelationshipTemplateDVO.test.ts
```diff
@@ -142,12 +143,19 @@ describe("RelationshipTemplateDVO", () => {
...
+                content: templateContent,
+                forIdentity: requestor.address
+            })
+        ).value;
         templateId = templatorTemplate.id;
     });
 
     test("TemplateDVO for templator", async () => {
-        const dto = templatorTemplate;
+        const dto = templatorTemplate as RelationshipTemplateDTO & { content: RelationshipTemplateContentJSON };
```
jkoenig134, line 158: I don't see a reason for this change.

packages/transport/test/modules/tokens/TokenController.test.ts
```diff
@@ -108,11 +108,12 @@ describe("TokenController", function () {
         expect((receivedToken.cache?.content as TokenContentFile).secretKey.toBase64()).toBe((sentToken.cache?.content as TokenContentFile).secretKey.toBase64());
     });
 
-    test("should send and receive a TokenContentRelationshipTemplate", async function () {
+    test("should send and receive a personalized TokenContentRelationshipTemplate", async function () {
```
jkoenig134, line 111: also here - ADD a test instead of replacing another one.

packages/transport/test/modules/relationshipTemplates/RelationshipTemplateController.test.ts
```diff
@@ -115,15 +115,37 @@ describe("RelationshipTemplateController", function () {
         }).rejects.toThrow(/SendRelationshipTemplateParameters.maxNumberOfAllocations/);
     });
 
-    test("should send and receive a RelationshipTemplate using a RelationshipTemplateReference", async function () {
-        tempDate = CoreDate.utc().subtract(TestUtil.tempDateThreshold);
-        const sentRelationshipTemplate = await TestUtil.sendRelationshipTemplate(sender);
+    test("should create and load a personalized template", async function () {
```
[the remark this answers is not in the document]
Magnus-Kuhn: I removed the option of loading a template by reference, leaving only loading the template by a truncated reference

packages/runtime/test/dataViews/RelationshipTemplateDVO.test.ts
Magnus-Kuhn: Without it, I get "Property 'title' does not exist on type 'RelationshipTemplateContentDerivation'.
  Property 'title' does not exist on type 'ArbitraryRelationshipTemplateContentJSON'." in line 167

jkoenig134: you can re-add it above.

jkoenig134: (where you removed it :) )

packages/transport/test/modules/relationshipTemplates/RelationshipTemplateController.test.ts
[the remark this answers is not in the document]
jkoenig134: I see. You see how things like these should be made in separate PRs to not confuse reviewers?

packages/transport/test/modules/tokens/TokenController.test.ts
Magnus-Kuhn: I don't have a strong opinion in this case, but on a general basis: Would this approach make test files larger and thus make it more difficult to keep track of what is currently tested?

jkoenig134: you could screw up sending and receiving un-personalized TokenContentRelationshipTemplate and do not recognize it, because you deleted the test for it.

I have a really strong opinion on that.

packages/runtime/src/useCases/transport/relationshipTemplates/CreateTokenForOwnRelationshipTemplate.ts
Magnus-Kuhn: I'd rather take the personalization from the Template rather than validating a personalization given from the user

jkoenig134: Yeah, if there is a personalization. But what you could do is having an unpersonalized template and putting a personalization on with a token.

jkoenig134 (approved):
LGTM

jkoenig134:
@stnmtz you are blocking this PR 'case you have changes requested. Any addition to comments you had?

packages/runtime/src/useCases/common/Schemas.ts
jkoenig134: Fixed in #294

stnmtz (approved):

PR #271, opened 2024-09-13
Change the message of "error.consumption.validation.inheritedFromItem" when using canCreate

RuthDiG:
# Readiness checklist

-   [X] I labeled the PR.

### DIFF
--- a/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
+++ b/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
@@ -306,6 +306,10 @@ class Requests {
         return new CoreError("error.consumption.requests.missingRelationship", message);
     }
 
+    public inheritedFromItem(message: string) {
+        return new ApplicationError("error.consumption.validation.inheritedFromItem", message);
+    }
+
     private static readonly _decideValidation = class {
         public invalidNumberOfItems(message: string) {
             return new ApplicationError("error.consumption.requests.decide.validation.invalidNumberOfItems", message);
--- a/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
+++ b/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
@@ -307,7 +307,7 @@ class Requests {
     }
 
     public inheritedFromItem(message: string) {
-        return new ApplicationError("error.consumption.validation.inheritedFromItem", message);
+        return new ApplicationError("error.consumption.requests.validation.inheritedFromItem", message);
     }
 
     private static readonly _decideValidation = class {
--- a/packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
+++ b/packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
@@ -115,11 +115,7 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
     private async _create(id: CoreId, content: Request, peer: CoreAddress) {
         const canCreateResult = await this.canCreate({ content, peer });
 
-        if (canCreateResult.isError()) {
-            throw ConsumptionCoreErrors.requests.inheritedFromItem(
-                "Some child items have errors. If this error occurred during the specification of a Request, call 'canCreate' to get more information."
-            );
-        }
+        if (canCreateResult.isError()) throw canCreateResult.error;
 
         const request = LocalRequest.from({
             id: id,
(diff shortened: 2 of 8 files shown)


packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -115,7 +115,15 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
     private async _create(id: CoreId, content: Request, peer: CoreAddress) {
         const canCreateResult = await this.canCreate({ content, peer });
 
-        if (canCreateResult.isError()) throw canCreateResult.error;
+        if (canCreateResult.isError()) {
+            if (canCreateResult.error.code === "error.consumption.validation.inheritedFromItem") {
+                throw ConsumptionCoreErrors.requests.inheritedFromItem(
+                    "Some child items have errors. If this error occurred during the specification of a Request, call 'canCreate' to get more information."
```
jkoenig134, line 121: we actually know that this occurred during _create, so we can simplify this

```suggestion
                    "Some child items have errors. Call 'canCreate' to get more information."
```

packages/runtime/test/modules/DeciderModule.test.ts
Milena-Czierlinski, line 1: This doesn't seem to be the right place for such a test as it has nothing to do with the decider module.

RuthDiG: Yes so I write a test in OutgoingRequestsController.test.ts. Do you think the test there is enough or should I add another test on another place?

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -115,7 +115,15 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
...
+                );
+            }
```
jkoenig134, line 123:
```suggestion
            const error = ConsumptionCoreErrors.requests.inheritedFromItem("Some child items have errors. Call 'canCreate' to get more information.");
            if (canCreateResult.error.equals(error)) throw error;
```

I'd prefer to make this bulletproof and not to hardcode this error code.

packages/runtime/test/modules/DeciderModule.test.ts
Milena-Czierlinski: In my opinion the other tests are sufficient.

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
RuthDiG: There is the problem that I want that when the error is  ConsumptionCoreErrors.requests.inheritedFromItem("Some child items have errors.") that I want to change it to  ConsumptionCoreErrors.requests.inheritedFromItem("Some child items have errors.  Call 'canCreate' to get more information.").

jkoenig134: I actually do exactly the same. `CoreError#equals` only compares the error code.

packages/runtime/test/modules/DeciderModule.test.ts
jkoenig134: Yep, it's enough to only test this in consumption :)

packages/consumption/src/consumption/ConsumptionCoreErrors.ts
```diff
@@ -306,6 +306,10 @@ class Requests {
         return new CoreError("error.consumption.requests.missingRelationship", message);
     }
 
+    public inheritedFromItem(message: string) {
+        return new ApplicationError("error.consumption.validation.inheritedFromItem", message);
```
britsta, line 310: `.requests` missing in the error code if you place the error here in the file under the class `Requests`, for example, `"error.consumption.requests.validation.inheritedFromItem"` instead of `"error.consumption.validation.inheritedFromItem"`.

RuthDiG: Yes that's true I didn't want to change the docu... But I will do it then.

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -115,7 +115,11 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
     private async _create(id: CoreId, content: Request, peer: CoreAddress) {
         const canCreateResult = await this.canCreate({ content, peer });
 
-        if (canCreateResult.isError()) throw canCreateResult.error;
+        if (canCreateResult.isError()) {
+            const error = ConsumptionCoreErrors.requests.inheritedFromItem("Some child items have errors. Call 'canCreate' to get more information.");
+            if (canCreateResult.error.equals(error)) throw error;
+            throw canCreateResult.error;
```
jkoenig134, line 121:
```suggestion
            if (canCreateResult.error.equals(error)) throw error;

            throw canCreateResult.error;
```

as this results in very different things I'd love to see a newline here to make this easier to process.

jkoenig134 (approved):

PR #274, opened 2024-09-19
Make relationship setting usable

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

# Description

As the App wants to be able to override names for contacts we have to update the functionality that @stnmtz build with new possibilities from the settings module, to be usable again.

As it is also a requirement to revert the contact name to the runtime default I also return the original name.

### DIFF
--- a/packages/runtime/src/dataViews/DataViewExpander.ts
+++ b/packages/runtime/src/dataViews/DataViewExpander.ts
@@ -1698,15 +1698,7 @@ export class DataViewExpander {
     }
 
     private async createRelationshipDVO(relationship: RelationshipDTO): Promise<RelationshipDVO> {
-        let relationshipSetting: RelationshipSettingDVO;
-        const settingResult = await this.consumption.settings.getSettings({ query: { reference: relationship.id } });
-        if (settingResult.value.length > 0) {
-            relationshipSetting = settingResult.value[0].value;
-        } else {
-            relationshipSetting = {
-                isPinned: false
-            };
-        }
+        const relationshipSetting = await this.getRelationshipSettingDVO(relationship);
 
         const stringByType: Record<string, undefined | string> = {};
         const relationshipAttributesResult = await this.consumption.attributes.getPeerSharedAttributes({ onlyValid: true, peer: relationship.peer });
@@ -1797,6 +1789,25 @@ export class DataViewExpander {
             auditLog: relationship.auditLog,
             creationContent: relationship.creationContent
         };
+    }
+
+    private async getRelationshipSettingDVO(relationship: RelationshipDTO): Promise<RelationshipSettingDVO> {
+        const settingResult = await this.consumption.settings.getSettings({
+            query: {
+                scope: "Relationship",
+                reference: relationship.id
+            }
+        });
+
+        const defaultSetting = { isPinned: false };
+
+        if (settingResult.value.length === 0) return defaultSetting;
+
+        const value = settingResult.value.reduce((prev, current) => (prev.createdAt > current.createdAt ? prev : current));
+
+        if (typeof value !== "object") return defaultSetting;
+
+        return { ...defaultSetting, ...value };
     }
 
     public async expandRelationshipDTO(relationship: RelationshipDTO): Promise<IdentityDVO> {
--- a/packages/runtime/src/dataViews/DataViewExpander.ts
+++ b/packages/runtime/src/dataViews/DataViewExpander.ts
@@ -1804,7 +1804,8 @@ export class DataViewExpander {
 
         if (settingResult.value.length === 0) return defaultSetting;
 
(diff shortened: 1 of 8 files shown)


jkoenig134:
@Siolto FYI

packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -1799,6 +1792,26 @@ export class DataViewExpander {
...
 
+    private async getRelationshipSettingDVO(relationship: RelationshipDTO): Promise<RelationshipSettingDVO> {
+        const settingResult = await this.consumption.settings.getSettings({
+            query: {
+                scope: "Relationship",
+                reference: relationship.id
+            }
+        });
+
+        const defaultSetting = { isPinned: false };
```
Milena-Czierlinski, line 1803: Out of interest: What does this `isPinned` do?

packages/runtime/test/consumption/settings.test.ts
```diff
@@ -134,6 +134,19 @@ describe("Settings", () => {
...
+        const key = "a-key";
+        const value = { aKey: "a-value" };
+        const reference = (await TransportIds.generic.generate()).toString();
+        const scope = "Identity";
+
+        const upsertSettingResult = await consumptionServices.settings.upsertSettingByKey({ key, reference, scope, value });
+        expect(upsertSettingResult).toBeSuccessful();
+
+        const result = await consumptionServices.settings.getSettingByKey({ key, reference, scope });
+        expect(result).toBeSuccessful();
```
Milena-Czierlinski, line 148: You could check the result given that you don't have a separate "upsert a setting including reference and type" test.

packages/runtime/test/dataViews/RelationshipDVO.test.ts
```diff
@@ -105,4 +102,24 @@ describe("RelationshipDVO", () => {
 
         expect(dvo.relationship!.templateId).toBe(dto.template.id);
     });
+
+    test("check the relationship dvo for the templator wit active relationshipSetting", async () => {
```
Milena-Czierlinski, line 106: wit typo

packages/runtime/src/useCases/consumption/settings/GetSettingByKey.ts
```diff
@@ -1,12 +1,14 @@
...
+import { Setting, SettingsController, SettingScope } from "@nmshd/consumption";
 import { Inject } from "typescript-ioc";
 import { SettingDTO } from "../../../types";
 import { RuntimeErrors, SchemaRepository, SchemaValidator, UseCase } from "../../common";
 import { SettingMapper } from "./SettingMapper";
 
 export interface GetSettingByKeyRequest {
     key: string;
+    reference?: string;
+    scope?: "Identity" | "Device" | "Relationship";
```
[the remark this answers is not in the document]
jkoenig134: this was a bug. A Setting can never be defined only by the key but is also defined by the scope (identity is the default)

packages/runtime/src/dataViews/DataViewExpander.ts
jkoenig134: I don't know :D

packages/runtime/src/useCases/consumption/settings/GetSettingByKey.ts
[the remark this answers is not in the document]
Milena-Czierlinski: I mean why don't you use the property `SettingScope` here and instead write `"Identity" | "Device" | "Relationship"`? :)

packages/runtime/src/dataViews/DataViewExpander.ts
Milena-Czierlinski: According to @Siolto it is nowhere used in the App. So probably it doesn't make sense for it to be here.

jkoenig134: true, but this is a refactoring PR and has nothing to do with isPinned.

jkoenig134: Maybe we can use this for contact favorites :)

packages/runtime/src/useCases/consumption/settings/GetSettingByKey.ts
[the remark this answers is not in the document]
jkoenig134: ah, b/c we don't use enums in Requests.

packages/runtime/src/dataViews/DataViewExpander.ts
Milena-Czierlinski: Ok, let's review this and discuss it somewhere else. 👍

Milena-Czierlinski (approved):

PR #278, opened 2024-09-19
Datawallet modifications are not executed in the right order and wrong objects are deleted

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

# Description

When we first have written the Datawallet logic we were convinced that we can execute modifications in the order 

`creates -> updates -> deletes -> cacheChanges`.

With the Relationship termination now more in use we have encountered  bugs with this approach. This PR changes the behavior to run the modifications (except cacheChanges) in the **exact** order the arrived.

Some of the optimizations done here will be implemented by the backbone team later, but should not break as soon the optimize it.


jkoenig134:
@Siolto FYI, we should be able to test this soon in the App :)

packages/transport/src/modules/sync/DatawalletModificationsProcessor.ts
```diff
@@ -61,41 +56,60 @@ export class DatawalletModificationsProcessor {
...
-            const newObject = Serializable.fromUnknown(mergedPayload);
+            let resultingObject: any = {};
+            for (const modification of currentModifications) {
+                switch (modification.type) {
+                    case DatawalletModificationType.Create:
+                        resultingObject = { ...resultingObject, ...modification.payload };
+                        break;
+                    case DatawalletModificationType.Update:
+                        resultingObject = { ...resultingObject, ...modification.payload };
+                        break;
```
tnotheis, line 88: Could be written as

```suggestion
                    case DatawalletModificationType.Create:
                    case DatawalletModificationType.Update:
                        resultingObject = { ...resultingObject, ...modification.payload };
                        break;
```

This way you make it clear that the two are actually intended to be the same and that it is not a copy&paste error.

packages/transport/test/modules/sync/SyncController.ordered.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    afterAll(async () => {
+        await sender?.close();
+        await recipient?.close();
+        await recipientSecondDevice?.close();
+
+        await connection.close();
+    });
+
+    // eslint-disable-next-line jest/expect-expect
+    test("should execute datawallet modifications in the correct order", async function () {
```
tnotheis, line 33: I don't understand how the current implementation of this test should assert that the datawallet modifications are executed in the correct order. Is the fact that no exception is thrown  sufficient for you?

stnmtz (dismissed):

tnotheis (changes requested):

packages/transport/src/modules/sync/DatawalletModificationsProcessor.ts
```diff
@@ -117,57 +131,35 @@ export class DatawalletModificationsProcessor {
...
-
     private async applyCacheChanges() {
         if (this.cacheChanges.length === 0) {
             return;
         }
 
         this.ensureAllItemsAreCacheable();
 
-        const cacheChangesWithoutDeletes = this.cacheChanges.filter((c) => !this.deletes.some((d) => d.objectIdentifier.equals(c.objectIdentifier)));
+        const cacheChangesWithoutDeletes = this.cacheChanges.filter((c) => !this.deletedObjects.some((d) => c.objectIdentifier.equals(d)));
```
stnmtz, line 141: This looks like a bug to me.

packages/transport/src/modules/sync/DatawalletModificationsProcessor.ts
```diff
@@ -61,41 +56,60 @@ export class DatawalletModificationsProcessor {
...
 
-            let mergedPayload = { id: objectIdentifier };
+            const lastModification = currentModifications.at(-1)!;
+            if (lastModification.type === DatawalletModificationType.Delete) {
+                await targetCollection.delete({ id: objectIdentifier });
+                this.deletedObjects.push(objectIdentifier);
 
-            for (const create of currentCreates) {
-                mergedPayload = { ...mergedPayload, ...create.payload };
+                continue;
```
stnmtz, line 77: Afaik, this continue continues the inner for, but the outer for should be continued. I guess the code is running fine currently - it just checks a deleted item multiple times.

stnmtz: Nvm, was confused by a deleted line with a for loop above ;)

packages/transport/test/modules/sync/SyncController.ordered.test.ts
jkoenig134: I have added a comment

packages/transport/src/modules/sync/DatawalletModificationsProcessor.ts
jkoenig134: But I didn't deleted the object Identifiers :D

tnotheis (approved):

PR #279, opened 2024-09-20
Decrypting Messages fails when Relationship is not active anymore

Magnus-Kuhn:
# Readiness checklist

-   [x] I labeled the PR.


Magnus-Kuhn:
I'm not sure about how much sense it makes to create a relationship challenge after the relationship has been terminated - but then sending or requesting challenges via Messages is not supported, so you are supposed to use a side-channel for which termination doesn't matter?? idk

jkoenig134:
> I'm not sure about how much sense it makes to create a relationship challenge after the relationship has been terminated

IMO it doesn't

packages/transport/src/modules/challenges/ChallengeController.ts
```diff
@@ -31,9 +31,9 @@ export class ChallengeController extends TransportController {
     private async validateChallengeLocally(challenge: Challenge, signedChallenge: ChallengeSigned): Promise<{ isValid: boolean; correspondingRelationship?: Relationship }> {
         if (!challenge.createdBy) return { isValid: false };
 
-        const relationship = await this.parent.relationships.getActiveRelationshipToIdentity(challenge.createdBy);
+        const relationship = await this.parent.relationships.getOnceActiveRelationshipToIdentity(challenge.createdBy);
         if (!relationship) {
-            throw TransportCoreErrors.general.recordNotFound(Relationship, challenge.createdBy.toString());
+            throw TransportCoreErrors.general.recordNotFound("Once active Relationship to Identity", challenge.createdBy.toString());
```
jkoenig134, line 36: I think you forgot to undo this

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -507,10 +507,10 @@ export class MessageController extends TransportController {
                 );
             }
         } else {
-            relationship = await this.relationships.getActiveRelationshipToIdentity(envelope.createdBy);
+            relationship = await this.relationships.getOnceActiveRelationshipToIdentity(envelope.createdBy);
 
             if (!relationship) {
-                throw TransportCoreErrors.messages.missingOrInactiveRelationship(envelope.createdBy.toString());
+                throw TransportCoreErrors.general.recordNotFound("Once active Relationship to Identity", envelope.createdBy.toString());
```
jkoenig134, line 513: I really don't like the wording "once active Relationship"..

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -507,10 +507,10 @@ export class MessageController extends TransportController {
                 );
             }
         } else {
-            relationship = await this.relationships.getActiveRelationshipToIdentity(envelope.createdBy);
+            relationship = await this.relationships.getOnceActiveRelationshipToIdentity(envelope.createdBy);
```
jkoenig134, line 510:
```suggestion
            relationship = await this.relationships.getRelationshipToIdentity(envelope.createdBy);
```
this would be the way easier version without needing the new strange wording

packages/transport/src/modules/messages/MessageController.ts
jkoenig134: this would be easier like this when you simplify the above
```suggestion
                throw TransportCoreErrors.general.recordNotFound(Relationship, envelope.createdBy.toString());
```

packages/transport/src/modules/relationships/RelationshipsController.ts
```diff
@@ -127,6 +127,17 @@ export class RelationshipsController extends TransportController {
...
+        const relationship = await this.getRelationshipToIdentity(address);
+        if (!relationship || !this.relationshipWasOnceActive(relationship)) return;
+
+        return relationship;
+    }
+
+    public relationshipWasOnceActive(relationship: Relationship): boolean {
+        return !!relationship.cache?.auditLog.find((entry) => entry.newStatus === RelationshipStatus.Active);
+    }
+
```
jkoenig134, line 140: don't forget to remove this when you apply my idea

packages/transport/test/modules/challenges/Challenges.test.ts
jkoenig134, line 1: as you didn't changed challenges, I don't think this test should be edited.

packages/transport/test/modules/messages/MessageController.test.ts
```diff
@@ -219,12 +219,19 @@ describe("MessageController", function () {
     });
 
     describe("Relationship Termination", function () {
+        let messageId: CoreId;
         beforeAll(async function () {
```
jkoenig134, line 223: newline missing

jkoenig134 (changes requested):

jkoenig134:
@Magnus-Kuhn please always assign everyone available from the runtime team.

packages/transport/src/modules/challenges/ChallengeController.ts
Magnus-Kuhn: I'm not sure if we want to prevent validating challenges after relationship termination - usually challenges are very short-lived but idk

jkoenig134: maybe we should also discuss this to @stnmtz on monday.

jkoenig134: as discussed you should remove this.

jkoenig134 (changes requested):

jkoenig134 (approved):

PR #281, opened 2024-09-23
Remove object loading by id and key

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

# Description

When loading objects was introduced this was the main way to do it. As we now introduce personalized and password protected objects loading by id and key becomes very unsafe.

This will result in runtime@v6.


packages/runtime/test/transport/files.test.ts
```diff
@@ -333,18 +333,11 @@ describe("Load peer file with file id and secret", () => {
...
-        const response = await transportServices2.files.getOrLoadFile({ id: file.id, secretKey: file.secretKey });
-
-        expect(response).toBeSuccessful();
-        expect(response.value).toMatchObject({ ...file, isOwn: false });
-    });
-
     test("after peer file is loaded the file can be accessed under /Files/{id}", async () => {
         expect(file).toBeDefined();
 
+        await transportServices2.files.getOrLoadFile({ reference: file.truncatedReference });
```
Magnus-Kuhn, line 339: The describe-block `Load peer file with file id and secret` can go

Magnus-Kuhn:
The controllers `TokenController` etc. could also drop some loading methods or change them to private, but I don't think it's necessary and it would mean having to change tests

jkoenig134:
> The controllers TokenController etc. could also drop some loading methods or change them to private, but I don't think it's necessary and it would mean having to change tests

yes, out of scope.

jkoenig134:
> So what is the strategy for v6?

No strategy. Just release. We expect minimal impact by this change so no long-running communication needed.

packages/runtime/test/transport/files.test.ts
```diff
@@ -315,83 +315,7 @@ describe("Load peer file with token reference", () => {
...
-    test.each([
-        [null, "id must be string"],
-        [undefined, "must have required property 'id'"],
-        ["", "id must match pattern FIL.*"]
-    ])("cannot pass %p as file id", async (fileId, expectedMessage) => {
-        const response = await transportServices2.files.getOrLoadFile({ id: fileId!, secretKey: file.secretKey });
-
-        expect(response).toBeAnError(expectedMessage, "error.runtime.validation.invalidPropertyValue");
+        expect(response).toBeAnError("token / file reference invalid", "error.runtime.validation.invalidPropertyValue");
     });
```
Magnus-Kuhn, line 319: The tests "after peer file is loaded it can be accessed under /Files", "after peer file is loaded the file can be accessed under /Files/{id}", "before the peer file is loaded another client cannot access it" did not have to be removed

jkoenig134: I thought they were duplicated

jkoenig134: done.

Magnus-Kuhn (approved):

PR #284, opened 2024-09-24
Extend Truncated References

Magnus-Kuhn:
# Readiness checklist

-   [x] I labeled the PR.


packages/transport/src/core/Reference.ts
```diff
@@ -6,43 +6,67 @@ import { TransportCoreErrors } from "./TransportCoreErrors";
...
     public static fromTruncated(value: string): Reference {
         const truncatedBuffer = CoreBuffer.fromBase64URL(value);
         const splitted = truncatedBuffer.toUtf8().split("|");
 
-        if (splitted.length !== 3) {
+        if (![3, 5].includes(splitted.length)) {
             throw TransportCoreErrors.general.invalidTruncatedReference();
         }
 
         try {
```
jkoenig134, line 52: I am not a fan of this try...catch, you should try..catch every parseInt individually to give a better error message

packages/transport/src/core/Reference.ts
```diff
@@ -6,43 +6,67 @@ import { TransportCoreErrors } from "./TransportCoreErrors";
...
         return truncatedReference.toBase64URL();
     }
 
     public static fromTruncated(value: string): Reference {
         const truncatedBuffer = CoreBuffer.fromBase64URL(value);
         const splitted = truncatedBuffer.toUtf8().split("|");
 
-        if (splitted.length !== 3) {
+        if (![3, 5].includes(splitted.length)) {
             throw TransportCoreErrors.general.invalidTruncatedReference();
```
jkoenig134, line 49: you should give a better error message here

jkoenig134 (changes requested):

packages/transport/src/core/Reference.ts
```diff
@@ -6,43 +6,67 @@ import { TransportCoreErrors } from "./TransportCoreErrors";
...
+    @serialize()
+    public backboneBaseUrl?: string;
+
     @validate()
     @serialize()
     public key: CryptoSecretKey;
 
+    @validate({ nullable: true, regExp: new RegExp("^[0-9a-f]{4}$") })
+    @serialize()
+    public forIdentityTruncated?: string;
```
jkoenig134, line 30: why not 
```
    public forIdentity?: string;
```

I don't really understand where the word Truncated comes from here.

Magnus-Kuhn: because we only give the last digits. Wasn't sure if there would be misunderstandings, am fine with both

jkoenig134: > last digits

chars and digits, right?

Magnus-Kuhn: it's a hexadecimal - so you can spin it both ways

jkoenig134: True. I'm ok with truncated in this case, now I understand it.

jkoenig134 (approved):

PR #286, opened 2024-09-30
Block sending Messages to Identities which are in deletion or for which the Relationship is terminated

RuthDiG:
# Readiness checklist

-   [X] I labeled the PR.


packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,48 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+            for (const recipient of recipientsCoreAddress) {
+                const relationship = await this.relationshipsController.getActiveRelationshipToIdentity(recipient);
+                if (!relationship) {
+                    return TransportCoreErrors.messages.missingOrInactiveRelationship(recipient.toString());
+                }
+                if (typeof relationship.peerDeletionInfo?.deletionStatus !== "undefined") {
+                    peerInDeletionAddressArray.push(recipient.address);
+                }
+            }
+            if (peerInDeletionAddressArray.toString() !== "") {
```
Magnus-Kuhn, line 107: You can use `peerInDeletionAddressArray.length`

packages/runtime/test/lib/testUtils.ts
```diff
@@ -298,6 +299,26 @@ export async function sendMessageToMultipleRecipients(transportServices: Transpo
 
     return response.value;
 }
+export async function failToSendMessageToMultipleRecipients(
```
Magnus-Kuhn, line 302: it's simpler to refactor the existing `sendMessageToMultipleRecipients` to work even when failing

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -942,3 +948,111 @@ describe("Relationship existence check", () => {
...
+                        "@type": "TestRequestItem",
+                        mustBeAccepted: false
+                    }
+                ]
+            },
+            peer: services1.address
+        };
+        const result = await createRequest(services2, services1, requestContent);
+        expect(result).toBeAnError(
+            `You cannot create a request to '${services1.address.toString()}' since the peer is in status '${updatedRelationship.peerDeletionInfo?.deletionStatus}'.`,
```
Magnus-Kuhn, line 995: On general grounds it's cleaner to explicitly state the `deletionStatus` - doesn't matter here though

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -942,3 +948,111 @@ describe("Relationship existence check", () => {
...
+
+    test("messages with multiple recipients should fail with another ErrorMessage if more than one of the recipients has an active IdentityDeletionProcess", async () => {
+        const result = await failToSendMessageToMultipleRecipients(services2.transport, [services3.address, services1.address]);
+        expect(result).toBeAnError(
+            `The recipients with the following addresses '${services3.address.toString()},${services1.address.toString()}' have an active IdentityDeletionProcess so you cannot send a message to them.`,
+            "error.transport.messages.peerInDeletion"
+        );
+    });
+
+    test("sending Notification to an Identity which is in status 'ToBeDeleted' results in an error redirected from the Backbone", async () => {
```
Magnus-Kuhn, line 1049: In the long term, the backbone should not be treating this an error, this test can go

packages/runtime/test/lib/testUtils.ts
```diff
@@ -315,6 +336,15 @@ export async function sendMessageWithRequest(
     return sendMessageResult.value as MessageDTO & { content: RequestJSON };
 }
 
+export async function createRequest(
```
Magnus-Kuhn, line 339: You don't save anything by introducing a helper function for a one-line function

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -942,3 +948,111 @@ describe("Relationship existence check", () => {
...
+
+        const relationship2 = await establishRelationship(services2.transport, services3.transport);
+        relationshipId2 = relationship2.id;
+
+        await establishRelationship(services2.transport, services4.transport);
+    }, 30000);
+
+    afterAll(() => serviceProvider.stop());
+
+    test("returns error if the peer of the Request has an active IdentityDeletionProcess", async () => {
```
Magnus-Kuhn, line 981: The Requests test should be in consumption, I think they aren't even needed in the runtime - but I won't ask you to remove tests :)

Those tests generally feel more at home in the `identityDeletionPeerProcessing.test.ts`

Magnus-Kuhn (commented):
Some consumption tests

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -174,10 +174,10 @@ export class IncomingRequestsController extends ConsumptionBaseController {
         const relationship = await this.relationshipResolver.getRelationshipToIdentity(request.peer);
         // It is safe to decide an incoming Request when no Relationship is found as this is the case when the Request origins from onNewRelationship of the RelationshipTemplateContent
         if (relationship) {
-            if (typeof relationship.peerDeletionInfo?.deletionStatus !== "undefined") {
+            if (relationship.peerDeletionInfo?.deletionStatus) {
```
Magnus-Kuhn, line 177: I mean that you don't need two `if`s here, they can be combined into one `relationship && relationship.peerDeletionInfo?.deletionStatus` and this can be further abbreviated into `relationship?.peerDeletionInfo?.deletionStatus`

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,48 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+        if (transformedContent instanceof Mail || transformedContent instanceof ResponseWrapper || transformedContent instanceof ArbitraryMessageContent) {
+            const peerInDeletionAddressArray: string[] = [];
+            const recipientsCoreAddress = recipients.map((r) => CoreAddress.from(r));
+            for (const recipient of recipientsCoreAddress) {
+                const relationship = await this.relationshipsController.getActiveRelationshipToIdentity(recipient);
+                if (!relationship) {
+                    return TransportCoreErrors.messages.missingOrInactiveRelationship(recipient.toString());
+                }
+                if (typeof relationship.peerDeletionInfo?.deletionStatus !== "undefined") {
+                    peerInDeletionAddressArray.push(recipient.address);
```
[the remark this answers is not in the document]
RuthDiG: I won't remove it from the MessageController as the MessageController.sendMessage is also used directly for example by DeleteOwnSharedAttributeAndNotifyPeer.ts.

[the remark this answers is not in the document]
Magnus-Kuhn: Do you want to make the version where the backbone is queueing messages or the version where the backbone blocks everything? You exclude notifications in the validation here, but they would still be caught in the `MessageController.sendMessage`.

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -744,7 +748,7 @@ describe("RelationshipDecomposition", () => {
         templateId2 = relationship2.template.id;
 
         await createRelationshipData(services1, services3);
-        multipleRecipientsMessageId = (await sendMessageToMultipleRecipients(services1.transport, [services2.address, services3.address])).id;
+        multipleRecipientsMessageId = await sendMessageToMultipleRecipients(services1.transport, [services2.address, services3.address]);
```
Magnus-Kuhn, line 751: There was no need no remove the id (and to type `multipleRecipientsMessageId` to any)

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,53 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+        }
+        if (peerInDeletionAddressArray.length > 0) {
+            if (!peerInDeletionAddressArray[1]) {
+                return TransportCoreErrors.messages.peerInDeletion(
+                    `The recipient with the address '${peerInDeletionAddressArray[0]}' has an active IdentityDeletionProcess, so you cannot send them a Message.`
+                );
+            }
+            return TransportCoreErrors.messages.peerInDeletion(
+                `The recipients with the following addresses '${peerInDeletionAddressArray}' have an active IdentityDeletionProcess, so you cannot send them a Message.`
+            );
```
Magnus-Kuhn, line 122: It would be more consequent to return both the peers in deletion and the peers with an inactive relationship in the error message.

jkoenig134:
@RuthDiG could you please use full words instead of the abbreviation in the PR title? (`resp.`)

jkoenig134:
I had a hard time processing your PR title. What do you think about `Block sending messages to identities which are in deletion`?

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
```diff
@@ -1,22 +1,27 @@
...
 const serviceProvider = new RuntimeServiceProvider();
 let services1: TestRuntimeServices;
 let services2: TestRuntimeServices;
+let services3: TestRuntimeServices;
 let relationshipId: string;
+let relationshipId2: string;
 
 beforeAll(async () => {
-    const runtimeServices = await serviceProvider.launch(2);
+    const runtimeServices = await serviceProvider.launch(3, { enableRequestModule: true, enableDeciderModule: true, enableNotificationModule: true });
```
jkoenig134, line 13: We should be very careful with enabling modules for test suites. Is there no other possibility?

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,81 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+                `The recipient with the address '${peerInDeletionAddressArray[0]}' has an active IdentityDeletionProcess, so you cannot send them a Message.`
+            );
+        }
 
-        const recipient = CoreAddress.from(recipients[0]);
-        if (!recipient.equals(localRequest.peer)) return RuntimeErrors.general.invalidPropertyValue("The recipient does not match the Request's peer.");
+        if (peerInDeletionAddressArray.length > 0 && peerInDeletionAddressArray[1]) {
+            return TransportCoreErrors.messages.missingOrWrongRelationshipStatusOrPeerInDeletion(
+                `The recipients with the following addresses '${peerInDeletionAddressArray}' have an active IdentityDeletionProcess, so you cannot send them a Message.`
+            );
```
jkoenig134, line 150: This code has to be refactored massively, it is way to complex. 
I think you should start by extracting this to a method and then trying to apply the bouncer / early-return pattern.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,81 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
 
-        if (!_.isEqual(transformedContent.toJSON(), localRequest.content.toJSON())) {
-            return RuntimeErrors.general.invalidPropertyValue("The sent Request must have the same content as the LocalRequest.");
+            if (!_.isEqual(transformedContent.toJSON(), localRequest.content.toJSON())) {
+                return RuntimeErrors.general.invalidPropertyValue("The sent Request must have the same content as the LocalRequest.");
+            }
+
+            if (recipients.length > 1) return RuntimeErrors.general.invalidPropertyValue("Only one recipient is allowed for sending Requests.");
+
+            const recipient = CoreAddress.from(recipients[0]);
```
jkoenig134, line 91: I think you have forgotten to check if this recipient is in deletion.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,81 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+            if (!recipient.equals(localRequest.peer)) return RuntimeErrors.general.invalidPropertyValue("The recipient does not match the Request's peer.");
+            return;
```
jkoenig134, line 93: you should extract this to another private method as it's now too indented

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -489,7 +493,7 @@ describe("RelationshipTermination", () => {
                 to: [services2.address]
             }
         });
-        expect(result).toBeAnError(/.*/, "error.transport.messages.missingOrInactiveRelationship");
+        expect(result).toBeAnError(/.*/, "error.transport.messages.missingOrWrongRelationshipStatusOrPeerInDeletion");
```
jkoenig134, line 496: To many or's in one error message. We need meaningful error codes that let the user know what to do.

packages/runtime/test/transport/relationships.test.ts
jkoenig134: actually the backbone will not block any messages anymore. But yes, I second that this test can be removed as we don't want to test backbone functionalities here. Instead we actually should add a test that shows that you CAN send messages while the other side is marked as ToBeDeleted.

You can write this test and then mark it with `test.skip("...`

packages/transport/src/core/TransportCoreErrors.ts
```diff
@@ -83,6 +83,10 @@ class Messages {
     public missingOrInactiveRelationship(address: string) {
         return new CoreError("error.transport.messages.missingOrInactiveRelationship", `An active Relationship with the given address '${address}' does not exist.`);
     }
+
+    public missingOrWrongRelationshipStatusOrPeerInDeletion(message: string) {
+        return new CoreError("error.transport.messages.missingOrWrongRelationshipStatusOrPeerInDeletion", message);
```
jkoenig134, line 88: as written in the test this should be splitted into multiple errors.

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -293,20 +293,74 @@ export class MessageController extends TransportController {
...
+                addressArray.push(recipient);
             }
+            if (!(relationship && (relationship.status === RelationshipStatus.Terminated || relationship.status === RelationshipStatus.Active))) {
+                missingOrInactiveRelationshipAddressArray.push(recipient.address);
+            }
+            if (relationship?.peerDeletionInfo?.deletionStatus === "Deleted") {
+                peerDeletedAddressArray.push(recipient.address);
+            }
+        }
+        if (missingOrInactiveRelationshipAddressArray.length > 0) {
```
jkoenig134, line 321: looks like massive copy-paste from the other code I already commented on. Also here it is way to complex.

jkoenig134 (dismissed):

RuthDiG:
> I had a hard time processing your PR title. What do you think about `Block sending messages to identities which are in deletion`?

I can do so but perhaps it would be better `Block sending Messages to peer with Relationship blockade` to integrate also the case when the Relationship is terminated?

packages/transport/src/modules/messages/MessageController.ts
RuthDiG: I will do the whole validation here inclusive this from the 'Notification' because then it isn't anymore necessary to do it in `SendMessage.ts` and I will refactore it like you said above.

jkoenig134: Actually you can not do the whole validation here as the Transport library does not know Notifications or Requests.

I think it would be better to just remove the validation HERE, as transport should only transport the stuff that the runtime wants to transport. The only thing we could check here if one of the recipients is already deleted. This is a real transport issue as it will never be accepted by the backbone.

jkoenig134:
Then you completely leave out the identity deletion. Maybe you find a title that includes both.

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -287,26 +287,48 @@ export class MessageController extends TransportController {
...
+        if (!parsedParams.attachments) parsedParams.attachments = [];
+        /*
+        if (parameters.content.constructor.name === "Notification")
+        {
+*/
         const secret = await CoreCrypto.generateSecretKey();
         const serializedSecret = secret.serialize(false);
         const addressArray: ICoreAddress[] = [];
         const envelopeRecipients: MessageEnvelopeRecipient[] = [];
+        const peerDeletedAddressArray: string[] = [];
```
britsta, line 300: Consider a renaming of these variables as described in another comments.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,53 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
             );
         }
 
-        if (!(transformedContent instanceof Request)) return;
+        if (transformedContent instanceof Notification) return;
+        if (!(transformedContent instanceof Notification)) {
+            const peerDeletedAddressArray: string[] = [];
+            const peerToBeDeletedAddressArray: string[] = [];
+            const missingOrInactiveRelationshipAddressArray: string[] = [];
+            const recipientsCoreAddress = recipients.map((r) => CoreAddress.from(r));
```
Milena-Czierlinski, line 82: Why do you transform the strings to CoreAddresses and then later only use the strings again?

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,53 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+                if (!relationship) {
+                    missingOrInactiveRelationshipAddressArray.push(recipient.address);
+                }
+                if (relationship?.peerDeletionInfo?.deletionStatus === "Deleted") {
+                    peerDeletedAddressArray.push(recipient.address);
+                }
+                if (relationship?.peerDeletionInfo?.deletionStatus === "ToBeDeleted") {
+                    peerToBeDeletedAddressArray.push(recipient.address);
+                }
+            }
```
Milena-Czierlinski, line 94: I am missing newlines here.

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -287,26 +287,48 @@ export class MessageController extends TransportController {
...
 
-            const cipherForRecipient = await this.secrets.encrypt(relationship.relationshipSecretId, serializedSecret);
-            envelopeRecipients.push(
-                MessageEnvelopeRecipient.from({
-                    address: recipient,
-                    encryptedKey: cipherForRecipient
-                })
-            );
-            addressArray.push(recipient);
+        if (peerDeletedAddressArray.length > 0) {
```
britsta, line 326: Isn't it necessary to consider the case that the peer is "ToBeDeleted" in addition to the case that the peer is "Deleted" as well, as we do in the `SendMessage.ts`?

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,53 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+        if (transformedContent instanceof Notification) return;
+        if (!(transformedContent instanceof Notification)) {
+            const peerDeletedAddressArray: string[] = [];
+            const peerToBeDeletedAddressArray: string[] = [];
+            const missingOrInactiveRelationshipAddressArray: string[] = [];
+            const recipientsCoreAddress = recipients.map((r) => CoreAddress.from(r));
+            for (const recipient of recipientsCoreAddress) {
+                const relationship = await this.relationshipsController.getActiveRelationshipToIdentity(recipient);
+                if (!relationship) {
+                    missingOrInactiveRelationshipAddressArray.push(recipient.address);
```
Milena-Czierlinski, line 86: If you add a `continue` hereafter, you will be allowed to remove the ? after `relationship` in the following.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,53 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+            if (peerToBeDeletedAddressArray.length > 0) {
+                return TransportCoreErrors.messages.peerToBeDeleted(peerToBeDeletedAddressArray);
+            }
 
-        if (!_.isEqual(transformedContent.toJSON(), localRequest.content.toJSON())) {
-            return RuntimeErrors.general.invalidPropertyValue("The sent Request must have the same content as the LocalRequest.");
+            if (missingOrInactiveRelationshipAddressArray.length > 0) {
+                return TransportCoreErrors.messages.missingOrInactiveRelationship(missingOrInactiveRelationshipAddressArray);
+            }
         }
```
Milena-Czierlinski, line 107: There is a newline missing after this.

Milena-Czierlinski (commented):
Just got through a part of the PR so far.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,53 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
             );
         }
 
-        if (!(transformedContent instanceof Request)) return;
+        if (transformedContent instanceof Notification) return;
+        if (!(transformedContent instanceof Notification)) {
+            const peerDeletedAddressArray: string[] = [];
+            const peerToBeDeletedAddressArray: string[] = [];
```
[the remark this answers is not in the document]
jkoenig134: Or peersInDeletion

packages/transport/src/core/TransportCoreErrors.ts
```diff
@@ -79,8 +79,40 @@ class Messages {
...
+    public missingOrInactiveRelationship(address: string[] | string) {
+        if (typeof address !== "string" && address.length > 1) {
+            return new CoreError(
+                "error.transport.messages.missingOrInactiveRelationship",
+                `An active Relationship with the given addresses '${address}' do not exist, so you cannot send them a Message.`
+            );
+        }
+        return new CoreError(
+            "error.transport.messages.missingOrInactiveRelationship",
+            `An active Relationship with the given address '${address[0]}' does not exist, so you cannot send them a Message.`
```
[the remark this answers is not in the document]
RuthDiG: Good point, I have overlooked it.

packages/transport/src/modules/messages/MessageController.ts
RuthDiG: No, because here I can't distinguish between Notification and Message etc. and so I only block what needs to be blocked for Notifications. The method sendMessage from MessageController is only called directly in cases where a notification is sent or  in case of Requests you call canCreate before with the necessary validations.

packages/runtime/test/lib/testUtils.ts
```diff
@@ -294,9 +299,8 @@ export async function sendMessageToMultipleRecipients(transportServices: Transpo
         },
         attachments
     });
-    expect(response).toBeSuccessful();
```
britsta, line 297: Why was this removed?

packages/consumption/test/modules/requests/testHelpers/TestObjectFactory.ts
```diff
@@ -162,6 +164,102 @@ export class TestObjectFactory {
...
+                        publicKey: CoreBuffer.from("L1sPFQgS5CxgGs1ejBcWCQLCpeFXbRc1TQnSpuHQqDQ")
+                    })
+                }),
+            peerDeletionInfo:
+                properties?.peerDeletionInfo ??
+                PeerDeletionInfo.from({
+                    deletionStatus: PeerDeletionStatus.Deleted,
+                    deletionDate: CoreDate.from("2022-01-03T00:00:00.000Z")
+                }),
+            status: properties?.status ?? RelationshipStatus.Active,
```
britsta, line 185: Is the Relationship not set to the 'DeletionProposed' status after the peer has deleted themself?

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
```diff
@@ -64,4 +72,116 @@ describe("IdentityDeletionProcess", () => {
...
         expect(updatedRelationship.peerDeletionInfo).toBeUndefined();
     });
+
+    test("messages with multiple recipients should fail if one of the recipients has an active IdentityDeletionProcess", async () => {
+        await services1.transport.identityDeletionProcesses.initiateIdentityDeletionProcess();
+
+        await syncUntilHasEvent(services2, PeerToBeDeletedEvent, (e) => e.data.id === relationshipId);
+        await services2.eventBus.waitForRunningEventHandlers();
+
+        const result = await sendMessageToMultipleRecipients(services2.transport, [services1.address, services4.address]);
```
britsta, line 82: Use `services3.address` instead of `services4.address` here, because there is only a Relationship between `services2` and `services3` and not between `services2` and `services4`. Then, you can simplify by removing the variable `services4` from the file.

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
```diff
@@ -64,4 +72,116 @@ describe("IdentityDeletionProcess", () => {
...
+        const requestContent = {
+            content: {
+                items: [
+                    {
+                        "@type": "TestRequestItem",
+                        mustBeAccepted: false
+                    }
+                ]
+            },
+            peer: services2.address
```
britsta, line 168: Shouldn't be `services1.address` be the `peer`? (Because `services2` calls `canAccept` below.)

packages/transport/test/modules/messages/MessageController.test.ts
```diff
@@ -226,8 +226,9 @@ describe("MessageController", function () {
             await TestUtil.terminateRelationship(sender, recipient);
         });
 
-        test("should not send a message on a terminated relationship", async function () {
-            await expect(TestUtil.sendMessage(sender, recipient)).rejects.toThrow("error.transport.messages.missingOrInactiveRelationship");
+        // eslint-disable-next-line jest/no-disabled-tests
+        test.skip("should not send a message on a terminated relationship", async function () {
```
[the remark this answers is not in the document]
jkoenig134: @britsta I think this test was kind of intended to be here. We want to prepare a test that you can send messages to peers in deletion (which the backbone will allow soon)

@RuthDiG you should use test.todo for that.

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
```diff
@@ -64,4 +72,116 @@ describe("IdentityDeletionProcess", () => {
...
+        };
+        const rRequestMessage = await exchangeMessageWithRequestAndRequireManualDecision(services1, services2, requestContent);
+
+        const requestIds = rRequestMessage.request.content.id!;
```
britsta, line 172: `requestId` instead of `requestIds`.

packages/transport/test/modules/messages/MessageController.test.ts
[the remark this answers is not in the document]
jkoenig134: same for all other comments

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
```diff
@@ -64,4 +72,116 @@ describe("IdentityDeletionProcess", () => {
...
+        const canAcceptResult = (await services2.consumption.incomingRequests.canAccept({ requestId: requestIds, items: [{ accept: true }] })).value;
+        expect(canAcceptResult.isSuccess).toBe(true);
+
+        await services1.transport.identityDeletionProcesses.initiateIdentityDeletionProcess();
+        await syncUntilHasEvent(services2, PeerToBeDeletedEvent, (e) => e.data.id === relationshipId);
+        await services2.eventBus.waitForRunningEventHandlers();
+        const updatedRelationship2 = (await services2.transport.relationships.getRelationship({ id: relationshipId })).value;
```
britsta, line 179: `updatedRelationship` instead of `updatedRelationship2`.

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -492,6 +497,17 @@ describe("RelationshipTermination", () => {
         expect(result).toBeAnError(/.*/, "error.transport.messages.missingOrInactiveRelationship");
     });
 
+    // eslint-disable-next-line jest/no-disabled-tests
+    test.skip("sending Notification results in an error redirected from the Backbone", async () => {
+        const id = await ConsumptionIds.notification.generate();
+        const notificationToSend = Notification.from({ id, items: [TestNotificationItem.from({})] });
+        const result = await services1.transport.messages.sendMessage({ recipients: [services2.address], content: notificationToSend.toJSON() });
+        expect(result).toBeAnError(
+            `Cannot send message to ${services1.address.toString()} because the relationship to it is not active. In order to be able to send messages again, you have to reactivate the relationship.`,
```
britsta, line 506: I am wondering why this test does not throw an error, because `${services2.address.toString()}` instead of `${services1.address.toString()}` should be used (`services2` is the recipient instead of `services1`).

packages/runtime/src/useCases/transport/messages/SendMessage.ts
RuthDiG: I have changed it like @britsta suggested.

packages/transport/src/core/TransportCoreErrors.ts
```diff
@@ -79,8 +79,46 @@ class Messages {
...
+        if (address.length > 1) {
+            return new CoreError(
+                "error.transport.messages.peerDeleted",
+                `The recipients with the following addresses '${address}' are deleted, so you cannot send them a Message.`
+            );
+        }
+        return new CoreError("error.transport.messages.peerDeleted", `The recipient with the address '${address[0]}' is deleted, so you cannot send them a Message.`);
+    }
+
+    public peerToBeDeleted(address: string[]) {
```
britsta, line 111: I'd be in favor of `peerInDeletion`.

packages/transport/src/core/TransportCoreErrors.ts
```diff
@@ -79,8 +79,46 @@ class Messages {
...
+                `An active Relationship with the given address '${address}' does not exist, so you cannot send them a Message.`
+            );
+        }
+        return new CoreError(
+            "error.transport.messages.missingOrInactiveRelationship",
+            `An active Relationship with the given address '${address[0]}' does not exist, so you cannot send them a Message.`
+        );
+    }
+
+    public peerDeleted(address: string[]) {
```
britsta, line 101: I'd be in favor of `deletedPeer`.

packages/runtime/test/transport/relationships.test.ts
RuthDiG: Yes I have noticed this also, I think there the Backbone mixed something up.

packages/runtime/test/lib/testUtils.ts
RuthDiG: Otherwise I cannot use it for both cases the sucessful one and the one where it throws an error.

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
RuthDiG: yea it is before I combined the errorMessage (so you could see that to one you have no relationship and the other is toBeDeleted etc) but this wasn't desired, so now I can refactor this.

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
RuthDiG: No, because services1 send the request which services2 wants to accept and can only do so when services1 is not in Deletion.

packages/transport/src/core/TransportCoreErrors.ts
RuthDiG: but if we use `peerInDeletion` and `peerHasDeletionInfo` wouldn't be more consequent to use `peerDeleted`?

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
britsta: Ahh you're right!

packages/transport/src/core/TransportCoreErrors.ts
britsta: Mhh `deletedPeer` seems grammatically more fitting than `peerDeleted` to me, but no strong opinion on this.

packages/runtime/test/transport/relationships.test.ts
britsta: When we came to the same conclusion, we should report this to the Backbone.

RuthDiG: I had a look on it in the Backbone and the reason is because they only look for the relationship between to peers and there you have 'from' and 'to' depending who created the relationship and then they put 'to' in the errorMessage so it has nothing to do with the peer who wants to send the message.

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -287,26 +287,45 @@ export class MessageController extends TransportController {
...
+                    })
+                );
+                addressArray.push(recipient);
             }
+            if (!(relationship && (relationship.status === RelationshipStatus.Terminated || relationship.status === RelationshipStatus.Active))) {
+                peersWithMissingOrInactiveRelationship.push(recipient.address);
+            }
+            if (relationship?.peerDeletionInfo?.deletionStatus === "Deleted") {
+                deletedPeers.push(recipient.address);
+            }
```
Milena-Czierlinski, line 320: I feel like this section could be simplified, e.g. by restructuring or introducing variables with speaking names for repeated boolean expressions.

packages/runtime/test/transport/relationships.test.ts
tnotheis: > I had a look on it in the Backbone and the reason is because they only look for the relationship between to peers and there you have 'from' and 'to' depending who created the relationship and then they put 'to' in the errorMessage so it has nothing to do with the peer who wants to send the message.

Whoops. :) That's completely right. 🙈 I will fix it. 

@Milena-Czierlinski Regarding your suggested error message: that doesn't really work out, because when a message has multiple recipients, there are multiple relationships. So you wouldn't know which is the one that causes the problem.
I would simply use the correct address in the error message. Or is there a problem with that?

And lastly: I don't recommend you to check for error messages. At least not in their entirety. If you want to assert that the Backbone complains for the correct reason (=> the correct recipient), then checking for the string to contain the address of the recipient should be sufficient.
The problem with checking for the whole message is that this makes the test brittle. Once the backbone changes the error messages, your test will fail for no real reason.

britsta: > I would simply use the correct address in the error message. Or is there a problem with that?

Perfectly fine, there is no problem with that. The idea of the suggestion was to reduce work when fixing the error message, but if we could get the address it would be great.

> And lastly: I don't recommend you to check for error messages. At least not in their entirety. If you want to assert that the Backbone complains for the correct reason (=> the correct recipient), then checking for the string to contain the address of the recipient should be sufficient.
> The problem with checking for the whole message is that this makes the test brittle. Once the backbone changes the error messages, your test will fail for no real reason.

Good point!

tnotheis: Version 6.13.2 contains the fix for this.

RuthDiG: It was clear that this test will fail when you won't block Notifications anymore, that's why it was test.skip and only for debugging, but in general you are right!

packages/transport/src/core/TransportCoreErrors.ts
```diff
@@ -79,8 +79,46 @@ class Messages {
         );
     }
 
-    public missingOrInactiveRelationship(address: string) {
-        return new CoreError("error.transport.messages.missingOrInactiveRelationship", `An active Relationship with the given address '${address}' does not exist.`);
+    public missingOrInactiveRelationship(address: string[] | string) {
+        if (typeof address !== "string" && address.length > 1) {
+            return new CoreError(
+                "error.transport.messages.missingOrInactiveRelationship",
+                `An active Relationship with the given addresses '${address}' do not exist, so you cannot send them a Message.`
```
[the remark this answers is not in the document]
RuthDiG: do not exist because they are more than one.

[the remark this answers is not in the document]
britsta: I think "do not exist" is referring to "an active Relationship" and not to the addresses, which is why it should be changed as suggested

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -942,3 +993,38 @@ describe("Relationship existence check", () => {
...
+    const serviceProvider = new RuntimeServiceProvider();
+    let services1: TestRuntimeServices;
+    let services2: TestRuntimeServices;
+    let services4: TestRuntimeServices;
+
+    beforeAll(async () => {
+        const runtimeServices = await serviceProvider.launch(4, { enableRequestModule: true, enableDeciderModule: true, enableNotificationModule: true });
+        services1 = runtimeServices[0];
+        services2 = runtimeServices[1];
+        services4 = runtimeServices[3];
```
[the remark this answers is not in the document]
RuthDiG: It doesn't matter because I will delete this test, first of all it isn't the right place and second I have to delete the test which are signed as test.skip

packages/transport/src/core/TransportCoreErrors.ts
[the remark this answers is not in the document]
RuthDiG: ups you are right!

packages/consumption/test/modules/requests/testHelpers/TestObjectFactory.ts
RuthDiG: Yes when the Backbone has finished with the implementation, so I changed it now like it will be in the future...

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -297,12 +297,8 @@ export class MessageController extends TransportController {
...
         const peersWithMissingOrInactiveRelationship: string[] = [];
         for (const recipient of parsedParams.recipients) {
-            const relationship = await this.relationships.getRelationshipToIdentity(recipient);
-            if (
-                relationship &&
-                (relationship.status === RelationshipStatus.Terminated || relationship.status === RelationshipStatus.Active) &&
-                relationship.peerDeletionInfo?.deletionStatus !== "Deleted"
-            ) {
+            const relationship = await this.relationships.getActiveOrTerminatedRelationshipToIdentity(recipient);
+            if (relationship && relationship.peerDeletionInfo?.deletionStatus !== "Deleted") {
```
Milena-Czierlinski, line 301: I think you can make it even simpler. I suggest
- get Relationship without specifying the status
- check if peer is deleted, if so push to `deletedPeers` and `continue`
- check if Relationship doesn't exist, if so push to `peersWithMissingOrInactiveRelationship` and `continue`
- otherwise process for usual case (without the need for `if` or `else`)

Doing so, you don't need a new function `getActiveOrTerminatedRelationshipToIdentity.

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -287,18 +287,25 @@ export class MessageController extends TransportController {
...
+            const relationship = await this.relationships.getRelationshipToIdentity(recipient);
+            if (relationship?.peerDeletionInfo?.deletionStatus === "Deleted") {
+                deletedPeers.push(recipient.address);
+                continue;
+            }
+            if (!relationship || !(relationship.status === "Terminated" || relationship.status === "Active")) {
+                peersWithMissingOrInactiveRelationship.push(recipient.address);
+                continue;
             }
-
```
Milena-Czierlinski, line 301: I'd re-add this blank line and also one after line 304 for better readability.

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -333,19 +348,16 @@ export class MessageController extends TransportController {
...
 
         for (const recipient of parsedParams.recipients) {
-            const relationship = await this.relationships.getActiveRelationshipToIdentity(CoreAddress.from(recipient));
-            if (!relationship) {
-                throw TransportCoreErrors.messages.missingOrInactiveRelationship(recipient.toString());
-            }
+            const relationship = await this.relationships.getRelationshipToIdentity(CoreAddress.from(recipient));
 
-            const signature = await this.secrets.sign(relationship.relationshipSecretId, plaintextBuffer);
+            const signature = await this.secrets.sign(relationship!.relationshipSecretId, plaintextBuffer);
```
Milena-Czierlinski, line 353: Usually we shouldn't use `!` in productive coding. I'm not sure if it okay here, since we do know that the relationship must exists. Does anyone else have an opinion on this?

Milena-Czierlinski (commented):
The rest looks good to me. @britsta Will you also do a final review?

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -102,7 +102,7 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
         }
 
         if (peersWithMissingOrInactiveRelationship.length > 0) {
-            return TransportCoreErrors.messages.missingOrInactiveRelationship(peersWithMissingOrInactiveRelationship);
+            return TransportCoreErrors.messages.missingRelationshipOrWrongRelationshipStatus(peersWithMissingOrInactiveRelationship);
```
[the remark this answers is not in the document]
RuthDiG: Okay then I divide it into two errors.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,54 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+                peersWithMissingOrInactiveRelationship.push(recipient);
+                continue;
+            }
+            if (relationship.peerDeletionInfo?.deletionStatus === "Deleted") {
+                deletedPeers.push(recipient);
+            }
+            if (relationship.peerDeletionInfo?.deletionStatus === "ToBeDeleted") {
+                peersInDeletion.push(recipient);
+            }
+        }
```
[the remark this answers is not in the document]
jkoenig134: (you validate that the recipients are correct)

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -309,6 +319,14 @@ export class MessageController extends TransportController {
             addressArray.push(recipient);
         }
 
+        if (deletedPeers.length > 0) {
+            throw TransportCoreErrors.messages.peerDeleted(deletedPeers);
+        }
+
+        if (peersWithMissingOrInactiveRelationship.length > 0) {
+            throw TransportCoreErrors.messages.missingRelationshipOrWrongRelationshipStatus(peersWithMissingOrInactiveRelationship);
+        }
```
jkoenig134, line 328: see in SendMessageUseCase

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -287,16 +287,26 @@ export class MessageController extends TransportController {
...
     public async sendMessage(parameters: ISendMessageParameters): Promise<Message> {
         const parsedParams = SendMessageParameters.from(parameters);
+
         if (!parsedParams.attachments) parsedParams.attachments = [];
 
         const secret = await CoreCrypto.generateSecretKey();
         const serializedSecret = secret.serialize(false);
         const addressArray: ICoreAddress[] = [];
         const envelopeRecipients: MessageEnvelopeRecipient[] = [];
+        const deletedPeers: string[] = [];
```
jkoenig134, line 297: all this new logic should be extracted to a new **private** method. sendMessage gets way to cluttered here.

packages/transport/test/modules/messages/MessageController.test.ts
```diff
@@ -226,9 +226,7 @@ describe("MessageController", function () {
             await TestUtil.terminateRelationship(sender, recipient);
         });
 
-        test("should not send a message on a terminated relationship", async function () {
-            await expect(TestUtil.sendMessage(sender, recipient)).rejects.toThrow("error.transport.messages.missingOrInactiveRelationship");
-        });
+        test.todo("should be able to send a Notification on a terminated relationship");
```
jkoenig134, line 229: you speak about Notifications in the transport library - the runtime does not know Notifications. should just be named message as the transport libs let's you send EVERY message regardless of the content.

jkoenig134 (dismissed):

packages/consumption/src/consumption/ConsumptionCoreErrors.ts
```diff
@@ -302,6 +302,10 @@ class Requests {
         return new CoreError("error.consumption.requests.wrongRelationshipStatus", message);
     }
 
+    public peerHasDeletionInfo(message: string) {
+        return new CoreError("error.consumption.requests.peerHasDeletionInfo", message);
+    }
```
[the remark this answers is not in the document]
RuthDiG: isInIdentityDeletion ? hasAnActiveIdentityDeletionProcess ?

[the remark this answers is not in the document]
RuthDiG: Or I divide also between toBeDeleted and Deleted?

packages/runtime/src/useCases/transport/messages/SendMessage.ts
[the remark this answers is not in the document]
RuthDiG: Okay but also before in the case of Requests there is a validation regarding the recipients, so I can simply rename the method to validateMessage ?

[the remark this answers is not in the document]
jkoenig134: I'm ok with that.

packages/consumption/src/consumption/ConsumptionCoreErrors.ts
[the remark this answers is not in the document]
jkoenig134: > divide also between toBeDeleted and Deleted

even better

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
```diff
@@ -64,4 +70,118 @@ describe("IdentityDeletionProcess", () => {
...
+        const canAcceptResult = (await services2.consumption.incomingRequests.canAccept({ requestId: requestId, items: [{ accept: true }] })).value;
+        expect(canAcceptResult.isSuccess).toBe(true);
+
+        await services1.transport.identityDeletionProcesses.initiateIdentityDeletionProcess();
+        await syncUntilHasEvent(services2, PeerToBeDeletedEvent, (e) => e.data.id === relationshipId);
+        await services2.eventBus.waitForRunningEventHandlers();
+        const updatedRelationship = (await services2.transport.relationships.getRelationship({ id: relationshipId })).value;
+        expect(updatedRelationship.peerDeletionInfo?.deletionStatus).toBe("ToBeDeleted");
+
+        const canAcceptResultAfterPeerDeletion = (await services2.consumption.incomingRequests.canAccept({ requestId: requestId, items: [{ accept: true }] })).value;
```
britsta, line 180: Rename variable, because peer isn't deleted yet, e.g.: `canAcceptResultAfterPeerInitiatedDeletion`

packages/transport/src/modules/messages/MessageController.ts
```diff
@@ -333,9 +339,10 @@ export class MessageController extends TransportController {
         const addressToRelationshipId: Record<string, ICoreId> = {};
 
         for (const recipient of parsedParams.recipients) {
-            const relationship = await this.relationships.getActiveRelationshipToIdentity(CoreAddress.from(recipient));
+            const relationship = await this.relationships.getRelationshipToIdentity(CoreAddress.from(recipient));
+
             if (!relationship) {
-                throw TransportCoreErrors.messages.missingOrInactiveRelationship(recipient.toString());
+                throw TransportCoreErrors.messages.hasNoActiveRelationship(recipient.toString());
```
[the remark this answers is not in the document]
RuthDiG: yea you are right!

britsta (approved):
LGTM 🐸 As discussed, let's wait for another review of @jkoenig134 and the required Backbone changes for replacing our test.todos with actual tests before merging the PR.

jkoenig134 (changes requested):
I'll block this agains an accidental merge before the backbone features drops

jkoenig134:
@RuthDiG [backbone 6.14.0](https://github.com/nmshd/backbone/releases/tag/6.14.0) is out, you can now write the tests for messages during relationship termination.

Milena-Czierlinski:
[Backbone 6.15.0](https://github.com/nmshd/backbone/releases/tag/6.15.0) is now out too, including the queueing of Message within the grace period of an IdentityDeletionProcess. So there should be nothing left blocking us.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -183,6 +183,23 @@ export class IncomingRequestsController extends ConsumptionBaseController {
 
         const relationship = await this.relationshipResolver.getRelationshipToIdentity(request.peer);
         // It is safe to decide an incoming Request when no Relationship is found as this is the case when the Request origins from onNewRelationship of the RelationshipTemplateContent
+
+        if (relationship?.peerDeletionInfo?.deletionStatus === PeerDeletionStatus.Deleted) {
+            return ValidationResult.error(
+                ConsumptionCoreErrors.requests.peerIsDeleted(
+                    `You cannot decide a Request from '${request.peer.toString()}' since the peer is in status '${relationship.peerDeletionInfo.deletionStatus}'.`
```
Milena-Czierlinski, line 190: Since you distinguish both cases here, there is no need to pass the deletion status as variable. You can simple write "... since the peer is deleted."

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -183,6 +183,23 @@ export class IncomingRequestsController extends ConsumptionBaseController {
...
+                )
+            );
+        }
+
+        if (relationship?.peerDeletionInfo?.deletionStatus === PeerDeletionStatus.ToBeDeleted) {
+            return ValidationResult.error(
+                ConsumptionCoreErrors.requests.peerIsToBeDeleted(
+                    `You cannot decide a Request from '${request.peer.toString()}' since the peer is in status '${relationship.peerDeletionInfo.deletionStatus}'.`
```
Milena-Czierlinski, line 198: Same here.

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -62,6 +62,22 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
                     )
                 );
             }
+
+            if (relationship.peerDeletionInfo?.deletionStatus === PeerDeletionStatus.Deleted) {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.peerIsDeleted(
+                        `You cannot create a Request to '${parsedParams.peer.toString()}' since the peer is in status '${relationship.peerDeletionInfo.deletionStatus}'.`
```
Milena-Czierlinski, line 69: Same here.

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -62,6 +62,22 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
...
+                    )
+                );
+            }
+
+            if (relationship.peerDeletionInfo?.deletionStatus === PeerDeletionStatus.ToBeDeleted) {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.peerIsToBeDeleted(
+                        `You cannot create a Request to '${parsedParams.peer.toString()}' since the peer is in status '${relationship.peerDeletionInfo.deletionStatus}'.`
```
Milena-Czierlinski, line 77: Same here.

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
```diff
@@ -64,4 +72,130 @@ describe("IdentityDeletionProcess", () => {
...
+        await services1.transport.identityDeletionProcesses.initiateIdentityDeletionProcess();
+        await syncUntilHasEvent(services2, PeerToBeDeletedEvent, (e) => e.data.id === relationshipId);
+        await services2.eventBus.waitForRunningEventHandlers();
+        const updatedRelationship = (await services2.transport.relationships.getRelationship({ id: relationshipId })).value;
+        expect(updatedRelationship.peerDeletionInfo?.deletionStatus).toBe("ToBeDeleted");
+        const id = await ConsumptionIds.notification.generate();
+        const notificationToSend = Notification.from({ id, items: [TestNotificationItem.from({})] });
+        const result = await services2.transport.messages.sendMessage({ recipients: [services1.address], content: notificationToSend.toJSON() });
+        expect(result).toBeSuccessful();
+    });
```
Milena-Czierlinski, line 200: Please add new lines to improve readability.

packages/consumption/src/consumption/ConsumptionCoreErrors.ts
```diff
@@ -305,6 +305,14 @@ class Requests {
         return new CoreError("error.consumption.requests.wrongRelationshipStatus", message);
     }
 
+    public peerIsDeleted(message: string) {
+        return new CoreError("error.consumption.requests.peerIsDeleted", message);
+    }
+
+    public peerIsToBeDeleted(message: string) {
+        return new CoreError("error.consumption.requests.peerIsToBeDeleted", message);
+    }
```
jkoenig134, line 314: peerIsInDeletion

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -180,6 +180,17 @@ export class IncomingRequestsController extends ConsumptionBaseController {
...
         const relationship = await this.relationshipResolver.getRelationshipToIdentity(request.peer);
         // It is safe to decide an incoming Request when no Relationship is found as this is the case when the Request origins from onNewRelationship of the RelationshipTemplateContent
+
+        if (relationship?.peerDeletionInfo?.deletionStatus === PeerDeletionStatus.Deleted) {
+            return ValidationResult.error(ConsumptionCoreErrors.requests.peerIsDeleted(`You cannot decide a Request from '${request.peer.toString()}' since the peer is deleted.`));
+        }
+
+        if (relationship?.peerDeletionInfo?.deletionStatus === PeerDeletionStatus.ToBeDeleted) {
+            return ValidationResult.error(
+                ConsumptionCoreErrors.requests.peerIsToBeDeleted(`You cannot decide a Request from '${request.peer.toString()}' since the peer has the status 'ToBeDeleted'.`)
```
jkoenig134, line 190: I would rather describe the status then name it with `'`s. The current wording is wrong. A peer (identity / Relationship) has no status named `ToBeDeleted` which can be very confusing.

jkoenig134: same everywhere else

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -187,7 +187,9 @@ export class IncomingRequestsController extends ConsumptionBaseController {
 
         if (relationship?.peerDeletionInfo?.deletionStatus === PeerDeletionStatus.ToBeDeleted) {
             return ValidationResult.error(
-                ConsumptionCoreErrors.requests.peerIsToBeDeleted(`You cannot decide a Request from '${request.peer.toString()}' since the peer has the status 'ToBeDeleted'.`)
+                ConsumptionCoreErrors.requests.peerIsInDeletion(
+                    `You cannot decide a Request from '${request.peer.toString()}' since the peer has an active IdentityDeletionProcess.`
```
jkoenig134, line 191: why not just say "the peer is in deletion"?

Milena-Czierlinski: That does sound simpler. 👍 Make sure to also adjust the error message accordingly in other places.

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -73,21 +74,48 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
...
+            if (!relationship) {
+                peersWithNoActiveRelationship.push(recipient);
+                continue;
+            }
+            if (relationship.peerDeletionInfo?.deletionStatus === PeerDeletionStatus.Deleted) {
+                deletedPeers.push(recipient);
+            }
+            if (relationship.peerDeletionInfo?.deletionStatus === PeerDeletionStatus.ToBeDeleted) {
+                peersInDeletion.push(recipient);
+            }
```
Milena-Czierlinski, line 93: Since you'll be touching the code again anyways, please add blank lines between the if-statements for better readability.

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
```diff
@@ -64,4 +80,156 @@ describe("IdentityDeletionProcess", () => {
...
+        const updatedRelationship = (await services2.transport.relationships.getRelationship({ id: relationshipId })).value;
+        expect(updatedRelationship.peerDeletionInfo?.deletionStatus).toBe("ToBeDeleted");
+
+        const id = await ConsumptionIds.notification.generate();
+        const notificationToSend = Notification.from({ id, items: [TestNotificationItem.from({})] });
+        const result = await services2.transport.messages.sendMessage({ recipients: [services1.address], content: notificationToSend.toJSON() });
+        expect(result).toBeSuccessful();
+    });
+
+    test("should be able to receive a Notification after the IdentityDeletionProcess has been cancelled.", async () => {
```
Milena-Czierlinski, line 212: Remove the period in the end here.

Milena-Czierlinski (approved):

jkoenig134 (changes requested):
blocking since I'm not finished with my review and don't want an accidental merge

packages/consumption/test/modules/requests/IncomingRequestsController.test.ts
```diff
@@ -412,9 +413,10 @@ describe("IncomingRequestsController", function () {
             await Given.aTerminatedRelationshipToIdentity();
             await Given.anIncomingRequestInStatus(LocalRequestStatus.DecisionRequired);
             const validationResult = await When.iCallCanAccept();
-            expect(validationResult).errorValidationResult({
-                code: "error.consumption.requests.wrongRelationshipStatus"
-            });
+            expect((validationResult as ErrorValidationResult).error.code).toBe("error.consumption.requests.wrongRelationshipStatus");
+            expect((validationResult as ErrorValidationResult).error.message).toContain(
+                "You cannot decide a request from 'did:e:a-domain:dids:anidentity' since the relationship is in status 'Terminated'."
+            );
```
jkoenig134, line 419: you made this code harder to read. please go back to errorValidationResult as this allows checking code and message in one call **and** additionally checks that

edit, I've done it and pushed it.

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -350,6 +355,59 @@ describe("Relationship status validations on active relationship", () => {
     });
 });
 
+describe("tests on inactive relationships", () => {
```
jkoenig134, line 358: 1. I don't like this describe name
2. shouldn't this be inside the message test file?

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -608,7 +667,18 @@ describe("RelationshipTermination", () => {
                 to: [services2.address]
             }
         });
-        expect(result).toBeAnError(/.*/, "error.transport.messages.missingOrInactiveRelationship");
+        expect(result).toBeAnError(/.*/, "error.transport.messages.hasNoActiveRelationship");
+    });
+
+    test("should be able to send a Notification", async () => {
```
jkoenig134, line 673: all this notification stuff should be in the test file for notifications

packages/transport/test/modules/messages/MessageController.test.ts
```diff
@@ -226,8 +226,8 @@ describe("MessageController", function () {
             await TestUtil.terminateRelationship(sender, recipient);
         });
 
-        test("should not send a message on a terminated relationship", async function () {
-            await expect(TestUtil.sendMessage(sender, recipient)).rejects.toThrow("error.transport.messages.missingOrInactiveRelationship");
+        test("should be able to send a message on a terminated relationship", async function () {
+            await expect(TestUtil.sendMessage(sender, recipient)).resolves.not.toThrow();
         });
```
jkoenig134, line 231: this is not enough testing. You should also check that the message is received when revived

jkoenig134: also I'd like to see a test about identityDeletion in transport

jkoenig134 (dismissed):

jkoenig134:
@RuthDiG your latest commits introduced some commented out code, please remove this, we don't merge commented out code.

packages/runtime/test/transport/relationships.test.ts
RuthDiG: But here the focus is on the RelationshipTermination and we also test, that a Message is not possible to send in this case.

jkoenig134: you are testing that the SendMessage UseCase reacts correctly on the different Relationship statuses.

But WHAT you test is the SendMessage UseCase and this decides where to put the test.

jkoenig134: Ehm, I think I messed up. In the first comment I wrote, that these tests belong to the notifications file, but as this tests the sendMessage UseCase it defenitely should be part of the messages tests. Sorry for the confusion.

RuthDiG: So also the test 'should not send a message' should be in the messages test, isn't so?

.dev/compose.backbone.env
jkoenig134, line 1: please re-add the newline :D

jkoenig134 (approved):
LGTM

PR #290, opened 2024-10-07
Add `deletionDate` to relationship's deletion info

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
+++ b/packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
@@ -1,4 +1,3 @@
-import { CoreDate } from "@nmshd/core-types";
 import { IdentityDeletionProcessStatus } from "@nmshd/transport";
 import { PeerDeletionCancelledEvent, PeerDeletionStatus, PeerToBeDeletedEvent } from "../../src";
 import { establishRelationship, RuntimeServiceProvider, syncUntilHasEvent, TestRuntimeServices } from "../lib";
@@ -50,7 +49,6 @@ describe("IdentityDeletionProcess", () => {
         const updatedRelationship = (await services2.transport.relationships.getRelationship({ id: relationshipId })).value;
         expect(updatedRelationship.peerDeletionInfo!.deletionStatus).toBe(PeerDeletionStatus.ToBeDeleted);
         expect(updatedRelationship.peerDeletionInfo!.deletionDate).toBeDefined();
-        expect(CoreDate.from(updatedRelationship.peerDeletionInfo!.deletionDate!).isAfter(CoreDate.local())).toBeTruthy();
     });
 
     test("peer should be notified about cancelled deletion process", async function () {
--- a/packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
+++ b/packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
@@ -1,3 +1,4 @@
+import { CoreDate } from "@nmshd/core-types";
 import { IdentityDeletionProcessStatus } from "@nmshd/transport";
 import { PeerDeletionCancelledEvent, PeerDeletionStatus, PeerToBeDeletedEvent } from "../../src";
 import { establishRelationship, RuntimeServiceProvider, syncUntilHasEvent, TestRuntimeServices } from "../lib";
@@ -14,13 +15,13 @@ beforeAll(async () => {
     relationshipId = (await establishRelationship(services1.transport, services2.transport)).id;
 }, 30000);
 
+afterAll(async () => {
+    return await serviceProvider.stop();
+});
+
 beforeEach(() => {
     services1.eventBus.reset();
     services2.eventBus.reset();
-});
-
-afterAll(async () => {
-    return await serviceProvider.stop();
 });
 
 afterEach(async () => {
@@ -34,6 +35,7 @@ afterEach(async () => {
     } else if (activeIdentityDeletionProcess.value.status === IdentityDeletionProcessStatus.WaitingForApproval) {
         abortResult = await services1.transport.identityDeletionProcesses.rejectIdentityDeletionProcess();
     }
+    await syncUntilHasEvent(services2, PeerDeletionCancelledEvent, (e) => e.data.id === relationshipId);
 
     if (abortResult?.isError) throw abortResult.error;
 });
@@ -47,6 +49,8 @@ describe("IdentityDeletionProcess", () => {
 
         const updatedRelationship = (await services2.transport.relationships.getRelationship({ id: relationshipId })).value;
         expect(updatedRelationship.peerDeletionInfo!.deletionStatus).toBe(PeerDeletionStatus.ToBeDeleted);
+        expect(updatedRelationship.peerDeletionInfo!.deletionDate).toBeDefined();
+        expect(CoreDate.from(updatedRelationship.peerDeletionInfo!.deletionDate!).isAfter(CoreDate.local())).toBeTruthy();
     });
 
     test("peer should be notified about cancelled deletion process", async function () {
--- a/packages/transport/src/modules/relationships/local/PeerDeletionInfo.ts
+++ b/packages/transport/src/modules/relationships/local/PeerDeletionInfo.ts
@@ -28,13 +28,15 @@ export class PeerDeletionInfo extends Serializable implements IPeerDeletionInfo 
     public deletionDate: CoreDate;
 
     public static override preFrom(value: any): any {
-        switch (value.deletionStatus as PeerDeletionStatus) {
-            case PeerDeletionStatus.ToBeDeleted:
-                value.deletionDate = CoreDate.local().add({ days: 14 }).toString();
-                break;
-            case PeerDeletionStatus.Deleted:
-                value.deletionDate = CoreDate.local().toString();
-                break;
+        if (!value.deletionDate) {
+            switch (value.deletionStatus as PeerDeletionStatus) {
+                case PeerDeletionStatus.ToBeDeleted:
+                    value.deletionDate = CoreDate.local().add({ days: 14 }).toString();
+                    break;
+                case PeerDeletionStatus.Deleted:
+                    value.deletionDate = CoreDate.local().toString();
+                    break;
+            }
         }
 
         return value;
--- a/packages/transport/src/modules/relationships/local/PeerDeletionInfo.ts
+++ b/packages/transport/src/modules/relationships/local/PeerDeletionInfo.ts
@@ -3,8 +3,7 @@ import { CoreDate } from "@nmshd/core-types";
 
 export enum PeerDeletionStatus {
     ToBeDeleted = "ToBeDeleted",
-    Deleted = "Deleted",
-    NeverDeletable = "NeverDeletable"
+    Deleted = "Deleted"
 }
 
 export interface PeerDeletionInfoJSON {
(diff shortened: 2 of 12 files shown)


packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
```diff
@@ -14,15 +14,15 @@ beforeAll(async () => {
     relationshipId = (await establishRelationship(services1.transport, services2.transport)).id;
 }, 30000);
 
+afterAll(async () => {
+    return await serviceProvider.stop();
```
jkoenig134, line 18: Why did you move this? We always do afterAll after beforeAll, I don't see a reason to do it differently here.

packages/transport/src/modules/relationships/local/PeerDeletionInfo.ts
```diff
@@ -20,6 +23,10 @@ export class PeerDeletionInfo extends Serializable implements IPeerDeletionInfo
     })
     public deletionStatus: PeerDeletionStatus;
 
+    @serialize()
+    @validate({ nullable: true })
+    public deletionDate?: CoreDate;
```
jkoenig134, line 28: This should not be nullable and instead default to now in 14 Days when the contact is to be deleted or now if the contact is already deleted.

jkoenig134 (changes requested):
I am missing one other external event processor regarding the identity deletion to be changed.

packages/runtime/test/transport/identityDeletionPeerProcessing.test.ts
sebbi08: Because before it was like

beforeAll
beforeEach
afterAll
afterEach

now it is 

beforeAll
afterAll
beforeEach
afterEach

we also have that in other tests.

packages/transport/src/modules/relationships/local/PeerDeletionInfo.ts
jkoenig134: > and instead default to now in 14 Days when the contact is to be deleted or now if the contact is already deleted

I think you forgot this whole part. Needs to be in preFrom.

sebbi08: I don't think we need the default, as we always get it from the corresponding External Event (PeerToBeDeleted and PeerDeleted). Or is there any other place where this default would be actually used ?

jkoenig134: Backwards compatibility.

packages/transport/src/modules/relationships/local/PeerDeletionInfo.ts
```diff
@@ -20,6 +23,17 @@ export class PeerDeletionInfo extends Serializable implements IPeerDeletionInfo
     })
     public deletionStatus: PeerDeletionStatus;
 
+    @serialize()
+    @validate()
+    public deletionDate: CoreDate;
+
+    public static override preFrom(value: any): any {
+        if (!value.deletionDate) {
+            value.deletionDate = value.deletionStatus === PeerDeletionStatus.ToBeDeleted ? CoreDate.local().add({ days: 14 }) : CoreDate.local();
```
jkoenig134, line 32: two things

1. switch over `value.deletionStatus as PeerDeletionStatus` to make sure the switch is exhaustive if we add another status
2. in each block do `CoreDate#toString()` to be closer to having the thing in the DB.

packages/transport/test/modules/sync/externalEventProcessors/PeerDeletedExternalEventProcessor.test.ts
```diff
@@ -35,7 +35,14 @@ describe("PeerDeletedExternalEventProcessor", function () {
...
+        await eventProcessor.execute({
+            id: "anId",
+            createdAt: "aDate",
+            index: 1,
+            syncErrorCount: 0,
+            type: "PeerDeleted",
+            payload: { relationshipId: relationshipId.toString(), deletionDate: CoreDate.local().add({ days: 24 }) }
+        });
         const relationship = await recipient.relationships.getRelationship(relationshipId);
         expect(relationship!.peerDeletionInfo!.deletionStatus).toBe(PeerDeletionStatus.Deleted);
```
jkoenig134, line 47: please add a test if the date was correctly saved

packages/transport/test/modules/sync/externalEventProcessors/PeerToBeDeletedExternalEventProcessor.test.ts
```diff
@@ -41,7 +41,7 @@ describe("PeerToBeDeletedExternalEventProcessor", function () {
             index: 1,
             syncErrorCount: 0,
             type: "PeerToBeDeleted",
-            payload: { relationshipId: relationshipId.toString() }
+            payload: { relationshipId: relationshipId.toString(), deletionDate: CoreDate.local().add({ days: 14 }) }
         });
         const relationship = await recipient.relationships.getRelationship(relationshipId);
         expect(relationship!.peerDeletionInfo!.deletionStatus).toBe(PeerDeletionStatus.ToBeDeleted);
```
jkoenig134, line 47: please add a test if the date was correctly saved

jkoenig134 (changes requested):

packages/transport/test/modules/relationships/PeerDeletionInfo.test.ts
```diff
@@ -0,0 +1,16 @@
...
+        const peerDeletionInfoToBeDeleted = PeerDeletionInfo.fromAny({
+            deletionStatus: PeerDeletionStatus.ToBeDeleted
+        });
+        const peerDeletionInfoDeleted = PeerDeletionInfo.fromAny({
+            deletionStatus: PeerDeletionStatus.Deleted
+        });
+
+        expect(peerDeletionInfoToBeDeleted.deletionDate.isAfter(CoreDate.local())).toBeTruthy();
+        expect(peerDeletionInfoDeleted.deletionDate.isSameOrBefore(CoreDate.local())).toBeTruthy();
+    });
```
jkoenig134, line 15: these are two tests.

packages/transport/test/modules/relationships/PeerDeletionInfo.test.ts
```diff
@@ -0,0 +1,16 @@
+import { CoreDate } from "@nmshd/core-types";
+import { PeerDeletionInfo, PeerDeletionStatus } from "../../../src";
+
+describe("PeerDeletionInfo", () => {
+    test("PeerDeletionInfo default values", () => {
```
jkoenig134, line 5: this doesn't match our test naming and does not explain what should be tested here. I you would have written a proper test name you would have noticed that this test should be splitted in two tests.

packages/transport/src/modules/relationships/local/PeerDeletionInfo.ts
```diff
@@ -20,6 +23,23 @@ export class PeerDeletionInfo extends Serializable implements IPeerDeletionInfo
...
+
+    public static override preFrom(value: any): any {
+        switch (value.deletionStatus as PeerDeletionStatus) {
+            case PeerDeletionStatus.ToBeDeleted:
+                value.deletionDate = CoreDate.local().add({ days: 14 }).toString();
+                break;
+            case PeerDeletionStatus.Deleted:
+                value.deletionDate = CoreDate.local().toString();
+                break;
+        }
```
jkoenig134, line 38: you introduced a bug by just deleting the if

packages/transport/test/modules/sync/externalEventProcessors/PeerDeletedExternalEventProcessor.test.ts
```diff
@@ -35,8 +35,16 @@ describe("PeerDeletedExternalEventProcessor", function () {
...
+            id: "anId",
+            createdAt: "aDate",
+            index: 1,
+            syncErrorCount: 0,
+            type: "PeerDeleted",
+            payload: { relationshipId: relationshipId.toString(), deletionDate: CoreDate.local() }
+        });
         const relationship = await recipient.relationships.getRelationship(relationshipId);
         expect(relationship!.peerDeletionInfo!.deletionStatus).toBe(PeerDeletionStatus.Deleted);
+        expect(relationship!.peerDeletionInfo!.deletionDate.isSameOrBefore(CoreDate.local())).toBeTruthy();
```
jkoenig134, line 48: why do you test it like this? you can put the date in a variable and check the exact date. Same for other test.

sebbi08:
For this one yea in the other the date is create in the preFrom

jkoenig134:
> For this one yea in the other the date is create in the preFrom

???

packages/transport/test/modules/sync/externalEventProcessors/PeerDeletedExternalEventProcessor.test.ts
sebbi08: For this one yea in the other the date is create in the preFrom

sebbi08:
> > For this one yea in the other the date is create in the preFrom
> 
> ???

wow the reply function from the teams bot is not really working. Put the comment to the right thread

packages/transport/test/modules/sync/externalEventProcessors/PeerDeletedExternalEventProcessor.test.ts
jkoenig134: i meant the other processor. ANd you have already changed this.

packages/transport/test/modules/sync/externalEventProcessors/PeerToBeDeletedExternalEventProcessor.test.ts
```diff
@@ -35,15 +35,20 @@ describe("PeerToBeDeletedExternalEventProcessor", function () {
...
             id: "anId",
             createdAt: "aDate",
             index: 1,
             syncErrorCount: 0,
             type: "PeerToBeDeleted",
-            payload: { relationshipId: relationshipId.toString() }
+            payload: {
+                relationshipId: relationshipId.toString(),
+                deletionDate: deletionDate
+            }
```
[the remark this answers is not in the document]
jkoenig134: you made the other one inline. Please decide for one...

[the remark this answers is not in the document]
jkoenig134: I am a big fan of inline in this case and it shouldn't be a problem with very small monitors.

jkoenig134 (dismissed):

packages/transport/test/modules/sync/externalEventProcessors/PeerDeletedExternalEventProcessor.test.ts
```diff
@@ -35,8 +35,17 @@ describe("PeerDeletedExternalEventProcessor", function () {
...
         const eventProcessor = new PeerDeletedExternalEventProcessor(recipient.identityDeletionProcess.eventBus, recipient);
-        await eventProcessor.execute({ id: "anId", createdAt: "aDate", index: 1, syncErrorCount: 0, type: "PeerDeleted", payload: { relationshipId: relationshipId.toString() } });
+        const deletionDate = CoreDate.local();
+        await eventProcessor.execute({
+            id: "anId",
+            createdAt: "aDate",
+            index: 1,
+            syncErrorCount: 0,
+            type: "PeerDeleted",
+            payload: { relationshipId: relationshipId.toString(), deletionDate: deletionDate }
```
jkoenig134, line 45: sorry, I didn't saw this. 

```suggestion
            payload: { relationshipId: relationshipId.toString(), deletionDate }
```

we always use the shorthand if possible.

packages/transport/test/modules/sync/externalEventProcessors/PeerToBeDeletedExternalEventProcessor.test.ts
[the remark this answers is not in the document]
sebbi08: yea now i remebered why i did it because before the CoreDate creation was in there and that was way longer.

jkoenig134 (approved):

PR #293, opened 2024-10-10
Upgrade the `form-data` lib

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

# Description

As described in https://github.com/form-data/form-data/issues/563 we want to make sure to be ready for node 23. As `form-data` 4.0.1 is now released this should do the job.

### DIFF
--- a/packages/transport/src/core/TransportCoreErrors.ts
+++ b/packages/transport/src/core/TransportCoreErrors.ts
@@ -1,5 +1,4 @@
 import { CoreError } from "@nmshd/core-types";
-import stringify from "json-stringify-safe";
 import { RelationshipStatus } from "../modules";
 
 class Relationships {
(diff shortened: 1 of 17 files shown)


packages/transport/src/core/TransportCoreErrors.ts
```diff
@@ -104,20 +104,13 @@ class Challenges {
 }
 
 class Datawallet {
-    public unsupportedModification(type: "unsupportedCacheChangedModificationCollection", data: any) {
-        const errorCode = "error.transport.datawallet.unsupportedModification";
+    public cacheChangeForUnsupportedCollectionDetected(data: any) {
```
Magnus-Kuhn, line 107: This renaming technically constitutes a breaking change and the code is included in the documentation, is it worth it?

jkoenig134: Docs yes, breaking no. I doubt that anyone relies on it.

jkoenig134: I think I have a solution to make it really backwards compatible.

jkoenig134: @Magnus-Kuhn changed, thanks for the heads-up!

Magnus-Kuhn (approved):

PR #295, opened 2024-10-14
Password-protected RelationshipTemplates

Magnus-Kuhn:
# Readiness checklist

-   [X] I labeled the PR.


.dev/compose.backbone.env
```diff
@@ -1 +1 @@
-BACKBONE_VERSION=6.10.0
+BACKBONE_VERSION=6.13.0-alpha.1
```
jkoenig134, line 1:
```suggestion
BACKBONE_VERSION=6.13.0
```

packages/runtime/src/useCases/transport/relationshipTemplates/CreateOwnRelationshipTemplate.ts
```diff
@@ -39,6 +40,17 @@ class Validator extends SchemaValidator<CreateOwnRelationshipTemplateRequest> {
...
+        if (input.password && /^\d+$/.test(input.password) && (input.password.length > 12 || input.password.length < 2)) {
+            validationResult.addFailure(
+                new ValidationFailure(
+                    RuntimeErrors.general.invalidPropertyValue(
+                        `Your chosen '${nameof<CreateOwnRelationshipTemplateRequest>((r) => r.password)}' is a PIN (consists of numbers only) and PINs must be at least 2 and at most 12 digits long`
+                    ),
+                    nameof<CreateOwnRelationshipTemplateRequest>((r) => r.password)
+                )
+            );
+        }
```
jkoenig134, line 52: plase put this to a private method to be able to do early return :)

packages/runtime/src/useCases/transport/relationshipTemplates/LoadPeerRelationshipTemplate.ts
```diff
@@ -29,16 +30,16 @@ export class LoadPeerRelationshipTemplateUseCase extends UseCase<LoadPeerRelatio
...
     }
 
-    private async loadRelationshipTemplateFromReference(reference: string): Promise<Result<RelationshipTemplateDTO>> {
+    private async loadRelationshipTemplateFromReference(reference: string, password?: string): Promise<Result<RelationshipTemplateDTO>> {
         if (reference.startsWith(Base64ForIdPrefix.RelationshipTemplate)) {
-            return await this.loadRelationshipTemplateFromRelationshipTemplateReference(reference);
+            return await this.loadRelationshipTemplateFromRelationshipTemplateReference(reference, password);
         }
 
         if (reference.startsWith(Base64ForIdPrefix.Token)) {
```
jkoenig134, line 45: you forgot to pass the password (next line)

I think you miss a test, this should have popped up in tests.

packages/runtime/src/useCases/transport/relationshipTemplates/CreateOwnRelationshipTemplate.ts
Magnus-Kuhn: What do you mean? This is part of the UseCase validation, what's supposed to be the improvement?

jkoenig134: readability.

jkoenig134: a bunch of boolean logic stuffed in one line can be overwhelming if not written yourself.

Magnus-Kuhn: The minimum PIN length was for the technical reason that we wanted passwordType 1 for non-PINs. Using passwordType 0 for non-PINs would solve that. The PIN characterization was always just for app displaying purposes which shouldn't depend on the length, then there would just be no upper bound on the passwordType.

I  don't know about max or min length - iirc the upper bound on the PIN was because at that point it would be user-friendlier to also use letters. IMO people should know what length constitutes a secure password/PIN for their use and we don't need to enforce it.

jkoenig134: ^ with the above information the password `1234567891011121314` must be of type `1`, you just throw instead.

jkoenig134: I'm ok with that.

Magnus-Kuhn: And regarding the lower bound? I'd be happy with a lower bound for the length (e. g. 2) as well as allowing 1-digit PINs that are then not treated as passwords :)

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -122,22 +131,37 @@ export class RelationshipTemplateController extends TransportController {
...
     }
 
-    public async fetchCaches(ids: CoreId[]): Promise<{ id: CoreId; cache: CachedRelationshipTemplate }[]> {
-        if (ids.length === 0) return [];
-
-        const backboneRelationships = await (await this.client.getRelationshipTemplates({ ids: ids.map((id) => id.id) })).value.collect();
-
-        const decryptionPromises = backboneRelationships.map(async (t) => {
-            const templateDoc = await this.templates.read(t.id);
+    private async readTemplates(ids: string[]): Promise<RelationshipTemplate[]> {
```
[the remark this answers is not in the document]
jkoenig134: and this method has to be blow it's last caller (fetchCaches), if you would have done this at the beginning it would have made the review way easier.

packages/transport/src/modules/relationshipTemplates/backbone/RelationshipTemplateClient.ts
```diff
@@ -6,11 +6,14 @@ import { BackbonePostRelationshipTemplatesRequest, BackbonePostRelationshipTempl
 
 export class RelationshipTemplateClient extends RESTClientAuthenticate {
     public async getRelationshipTemplates(request: BackboneGetRelationshipTemplatesRequest): Promise<ClientResult<Paginator<BackboneGetRelationshipTemplatesResponse>>> {
-        return await this.getPaged<BackboneGetRelationshipTemplatesResponse>("/api/v1/RelationshipTemplates", request);
+        return await this.getPaged<BackboneGetRelationshipTemplatesResponse>("/api/v1/RelationshipTemplates", {
+            templates: JSON.stringify(request.templates)
```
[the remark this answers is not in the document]
jkoenig134: also you are not future-proof for params added to the `BackboneGetRelationshipTemplatesRequest` you would simply ignore them

packages/transport/src/modules/relationshipTemplates/local/SendRelationshipTemplateParameters.ts
```diff
@@ -27,6 +28,18 @@ export class SendRelationshipTemplateParameters extends Serializable implements
     @serialize()
     public forIdentity?: CoreAddress;
 
+    @validate({
+        nullable: true,
+        customValidator: (input) => {
+            if (/^\d+$/.test(input) && (input.length > 12 || input.length < 2)) {
+                return "PINs must be at least 2 and at most 12 digits long";
+            }
```
jkoenig134, line 36: this validator is broken.

jkoenig134: (see the other discussion we had)

packages/transport/src/modules/tokens/transmission/TokenContentRelationshipTemplate.ts
```diff
@@ -22,6 +23,10 @@ export class TokenContentRelationshipTemplate extends Serializable implements IT
     @serialize()
     public forIdentity?: CoreAddress;
 
+    @validate({ nullable: true, min: 1, max: 12, customValidator: (v) => (!Number.isInteger(v) ? "must be an integer" : undefined) })
```
jkoenig134, line 26: also broken

packages/transport/test/modules/tokens/TokenController.test.ts
```diff
@@ -176,11 +176,83 @@ describe("TokenController", function () {
...
+        expect((receivedToken.cache?.content as TokenContentRelationshipTemplate).secretKey).toBeInstanceOf(CryptoSecretKey);
+        expect((sentToken.cache?.content as TokenContentRelationshipTemplate).templateId.toString()).toBe(content.templateId.toString());
+        expect((sentToken.cache?.content as TokenContentRelationshipTemplate).secretKey.toBase64()).toBe(content.secretKey.toBase64());
+        expect((receivedToken.cache?.content as TokenContentRelationshipTemplate).templateId.toString()).toBe(
+            (sentToken.cache?.content as TokenContentRelationshipTemplate).templateId.toString()
+        );
+        expect((receivedToken.cache?.content as TokenContentRelationshipTemplate).secretKey.toBase64()).toBe(
+            (sentToken.cache?.content as TokenContentRelationshipTemplate).secretKey.toBase64()
+        );
+        expect((receivedToken.cache?.content as TokenContentRelationshipTemplate).passwordType).toBe((sentToken.cache?.content as TokenContentRelationshipTemplate).passwordType);
```
jkoenig134, line 210: you could make this so much more readable by extracting `receivedToken.cache?.content as TokenContentRelationshipTemplate` to a variable

jkoenig134: same for the other occurrences

packages/runtime/test/transport/relationshipTemplates.test.ts
```diff
@@ -208,6 +209,99 @@ describe("Template Tests", () => {
...
+                content: emptyRelationshipTemplateContent,
+                expiresAt: DateTime.utc().plus({ minutes: 1 }).toString(),
+                password: "password"
+            });
+            expect(createResult).toBeSuccessful();
+            const createTokenResult = await runtimeServices1.transport.relationshipTemplates.createTokenQRCodeForOwnTemplate({
+                templateId: createResult.value.id
+            });
+            expect(createTokenResult).toBeSuccessful();
+        });
```
jkoenig134, line 303: this test is meaningless, what do you expect to go wrong here?

packages/runtime/test/transport/relationshipTemplates.test.ts
```diff
@@ -208,6 +209,99 @@ describe("Template Tests", () => {
...
+
+        test("validation error when creating a template with a too short PIN", async () => {
+            const createResult = await runtimeServices1.transport.relationshipTemplates.createOwnRelationshipTemplate({
+                content: emptyRelationshipTemplateContent,
+                expiresAt: DateTime.utc().plus({ minutes: 1 }).toString(),
+                password: "1"
+            });
+            expect(createResult).toBeAnError(
+                "Your chosen 'password' is a PIN \\(consists of numbers only\\) and PINs must be at least 2 and at most 12 digits long",
```
jkoenig134, line 271: what are these backslashes?

packages/runtime/test/transport/relationshipTemplates.test.ts
```diff
@@ -208,6 +209,99 @@ describe("Template Tests", () => {
...
+                password: "password"
+            });
+            expect(createResult).toBeSuccessful();
+            const createTokenResult = await runtimeServices1.transport.relationshipTemplates.createTokenForOwnTemplate({
+                templateId: createResult.value.id
+            });
+            expect(createTokenResult).toBeSuccessful();
+            const loadTokenResult = await runtimeServices2.transport.tokens.loadPeerToken({ reference: createTokenResult.value.truncatedReference, ephemeral: true });
+            expect(loadTokenResult).toBeSuccessful();
+            expect((loadTokenResult.value.content as TokenContentRelationshipTemplate).passwordType).toBe(1);
```
jkoenig134, line 289:
```suggestion
            expect((loadTokenResult.value.content as TokenContentRelationshipTemplateJSON).passwordType).toBe(1);
```

please check that if you are using transport types that you are only using JSON representations. it's just luck that this typecast works.

packages/runtime/test/transport/relationshipTemplates.test.ts
Magnus-Kuhn: escapes, since this will become a regex

packages/runtime/test/transport/relationshipTemplates.test.ts
Magnus-Kuhn: I might have (very unlikely) messed something up when changing it - not complaining about removing a test though :)

packages/transport/src/modules/relationshipTemplates/backbone/RelationshipTemplateClient.ts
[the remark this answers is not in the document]
Magnus-Kuhn: I originally thought axios would do it too, but it doesn't work without the stringify

packages/runtime/src/useCases/transport/relationshipTemplates/CreateOwnRelationshipTemplate.ts
Magnus-Kuhn: Should the bound be extended to passwords? Seems like it would be easier for us

packages/runtime/test/transport/relationshipTemplates.test.ts
Magnus-Kuhn: The JSON doesn't exist for this type and creating it for a key is iffy, I'll remove the typecast

packages/transport/src/modules/relationshipTemplates/backbone/RelationshipTemplateClient.ts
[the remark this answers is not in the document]
jkoenig134: Maybe you have to configure axios differently.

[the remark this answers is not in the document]
Magnus-Kuhn: Hmm... the backbone serializes this array as a JSON, but in other instances I looks like it serializes arrays in the conventional way, i. e. something like `/api/v1/RelationshipTemplates?templates.0.id=RLTU19d0lpIVbNWfoB7h&templates.0.password=password&templates.1.id=RLTZ4JPzmgd50eSDU6p6&templates.1.password=password` what axios currently creates.

I'd like there to be a clear stylistic direction before I write a custom configuration for it @tnotheis .

packages/runtime/src/useCases/transport/relationshipTemplates/CreateOwnRelationshipTemplate.ts
jkoenig134: I think it's way easier to  brute-force if you have the length, so I would avoid that.

Magnus-Kuhn: You would only know that the length is at least 2

packages/transport/src/modules/relationshipTemplates/backbone/RelationshipTemplateClient.ts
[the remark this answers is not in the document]
tnotheis: I will try to change the implementation of the backbone to support the format you mentioned

packages/transport/src/modules/relationshipTemplates/local/RelationshipTemplate.ts
```diff
@@ -50,7 +55,16 @@ export class RelationshipTemplate extends CoreSynchronizable implements IRelatio
...
-        return RelationshipTemplateReference.from({ id: this.id, key: this.secretKey, forIdentityTruncated: this.cache!.forIdentity?.toString().slice(-4) });
+        return RelationshipTemplateReference.from({
+            id: this.id,
+            key: this.secretKey,
+            forIdentityTruncated: this.cache!.forIdentity?.toString().slice(-4),
+            passwordType: this.passwordType
+        });
+    }
+
+    public get passwordType(): number | undefined {
```
britsta, line 66: In principle, it is good to have a distinction between PINs and "regular" passwords, especially when it comes to their representation in the UI. However, I find that the chosen approach can be problematic if my specified `password` is used to decide for me whether it is a PIN or not and what the UI representation is. Implicitly deriving the password type from my chosen password gives the App user the opportunity to draw conclusions about my password via the UI, without me necessarily having been aware of this.

I would find it more transparent and secure for Integrators of Connectors to have two optional fields `password` and `pin`, which both have an individual validation and representation in the UI.

Even if we don't need to tackle this in this PR, as it would change a lot, I just wanted to share my impression.

packages/transport/src/modules/relationshipTemplates/backbone/RelationshipTemplateClient.ts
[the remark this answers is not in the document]
tnotheis: Version 6.13.2 contains the fix for this.

packages/runtime/test/transport/relationshipTemplates.test.ts
jkoenig134: if it's a regex, please make it a regex.

and DONT resolve comments that are not resolved.

Magnus-Kuhn: While looking into ways to not have people be surprised that the message in this matcher is currently a regex, I noticed that for runtime errors the error code is included in the error message, which is not the case for transport/... errors. I think this should be unified.

packages/runtime/src/useCases/transport/relationshipTemplates/CreateOwnRelationshipTemplate.ts
```diff
@@ -19,6 +19,8 @@ export interface CreateOwnRelationshipTemplateRequest {
      */
     maxNumberOfAllocations?: number;
     forIdentity?: AddressString;
+    password?: string;
+    pin?: string;
```
jkoenig134, line 23:
```suggestion
    pin?: PINString;
```

packages/runtime/src/useCases/transport/relationshipTemplates/CreateOwnRelationshipTemplate.ts
```diff
@@ -19,6 +19,8 @@ export interface CreateOwnRelationshipTemplateRequest {
      */
     maxNumberOfAllocations?: number;
     forIdentity?: AddressString;
+    password?: string;
```
jkoenig134, line 22: missing a validation, the string should not be empty

packages/runtime/src/useCases/transport/relationshipTemplates/CreateOwnRelationshipTemplate.ts
```diff
@@ -61,7 +67,8 @@ export class CreateOwnRelationshipTemplateUseCase extends UseCase<CreateOwnRelat
             content: request.content,
             expiresAt: CoreDate.from(request.expiresAt),
             maxNumberOfAllocations: request.maxNumberOfAllocations,
-            forIdentity: request.forIdentity ? CoreAddress.from(request.forIdentity) : undefined
+            forIdentity: request.forIdentity ? CoreAddress.from(request.forIdentity) : undefined,
+            password: request.password
```
jkoenig134, line 71: pin is completely ignored here.

packages/runtime/src/useCases/transport/relationshipTemplates/LoadPeerRelationshipTemplate.ts
```diff
@@ -10,6 +10,7 @@ import { RelationshipTemplateMapper } from "./RelationshipTemplateMapper";
  */
 export interface LoadPeerRelationshipTemplateRequest {
     reference: TokenReferenceString | RelationshipTemplateReferenceString;
+    passwordOrPin?: string;
```
jkoenig134, line 13: a pin is also a password, I would rename this to password

packages/runtime/src/useCases/transport/relationshipTemplates/RelationshipTemplateMapper.ts
```diff
@@ -17,6 +17,8 @@ export class RelationshipTemplateMapper {
             createdByDevice: template.cache.createdByDevice.toString(),
             createdAt: template.cache.createdAt.toString(),
             forIdentity: template.cache.forIdentity?.toString(),
+            password: template.password,
+            pin: template.pin,
```
jkoenig134, line 21: this should be combined to only password

- "password" for passwords
- "pin:1234" for pins

packages/transport/test/modules/tokens/TokenReference.test.ts
```diff
@@ -166,7 +166,7 @@ describe("TokenReference", function () {
             TokenReference.from({
                 key: await CryptoEncryption.generateKey(),
                 id: await BackboneIds.token.generateUnsafe(),
-                passwordType: 20
+                passwordType: "pin20"
```
jkoenig134, line 169: havent we said that 16 is the maximum?

packages/runtime/src/useCases/transport/relationshipTemplates/RelationshipTemplateMapper.ts
Magnus-Kuhn: Do you mean "pw:password" (or something like that)? Also would integrators like this if they want to do something with the password in their own systems?

jkoenig134: nah, password can be the default.

packages/transport/test/modules/tokens/TokenReference.test.ts
Magnus-Kuhn: this is expected to fail :)

packages/runtime/src/useCases/transport/relationshipTemplates/RelationshipTemplateMapper.ts
jkoenig134: argh

jkoenig134: yeah, then password:

packages/runtime/src/useCases/transport/relationshipTemplates/LoadPeerRelationshipTemplate.ts
```diff
@@ -1,21 +1,54 @@
...
+} from "../../common";
 import { RelationshipTemplateMapper } from "./RelationshipTemplateMapper";
 
 /**
  * @errorMessage token / relationship template reference invalid
  */
 export interface LoadPeerRelationshipTemplateRequest {
     reference: TokenReferenceString | RelationshipTemplateReferenceString;
+    password?: string;
+    pin?: string;
```
jkoenig134, line 25: as there has to be made a difference between pin and password when **creating** an object to make it obvious for the consumer to enter a correct pin/password we should simplify the behavior here, by simply asking for a password. Same for other `load*` UseCases.

packages/transport/src/modules/tokens/transmission/TokenContentRelationshipTemplate.ts
```diff
@@ -22,7 +25,19 @@ export class TokenContentRelationshipTemplate extends Serializable implements IT
...
+
+    @validate({ nullable: true, customValidator: (v: ICoreBuffer) => (v.buffer.byteLength === 16 ? undefined : "must be 16 bytes long") })
+    @serialize()
+    public salt?: CoreBuffer;
+
     public static from(value: ITokenContentRelationshipTemplate): TokenContentRelationshipTemplate {
-        return this.fromAny(value);
+        const content = this.fromAny(value);
+        if (!content.passwordType !== !content.salt) {
+            throw TransportCoreErrors.general.invalidTruncatedReference("It's not possible to have only one of passwordType and salt set..");
```
jkoenig134, line 39:
```suggestion
            throw TransportCoreErrors.general.invalidTruncatedReference("It's not possible to have only one of passwordType and salt set.");
```
you don't have to be sad about wrong usage

packages/transport/src/modules/tokens/transmission/TokenContentRelationshipTemplate.ts
```diff
@@ -1,11 +1,14 @@
 import { ISerializable, Serializable, serialize, type, validate } from "@js-soft/ts-serval";
 import { CoreAddress, CoreId, ICoreAddress, ICoreId } from "@nmshd/core-types";
-import { CryptoSecretKey, ICryptoSecretKey } from "@nmshd/crypto";
+import { CoreBuffer, CryptoSecretKey, ICoreBuffer, ICryptoSecretKey } from "@nmshd/crypto";
+import { TransportCoreErrors } from "../../../core";
```
jkoenig134, line 4: to big barrel imported here - this can lead to circular references

packages/transport/src/core/Reference.ts
```diff
@@ -9,9 +9,12 @@ export interface IReference extends ISerializable {
     backboneBaseUrl?: string;
     key: ICryptoSecretKey;
     forIdentityTruncated?: string;
-    passwordType?: number;
+    passwordType?: string;
+    version?: number;
```
jkoenig134, line 13: what do we use this version for? I don't get it.

packages/transport/src/core/Reference.ts
```diff
@@ -29,14 +32,22 @@ export class Reference extends Serializable implements IReference {
...
+
+    @validate({ nullable: true, customValidator: (v: ICoreBuffer) => (v.buffer.byteLength === 16 ? undefined : "must be 16 bytes long") })
+    @serialize()
+    public salt?: CoreBuffer;
 
     public truncate(): string {
         const idPart = this.backboneBaseUrl ? `${this.id.toString()}@${this.backboneBaseUrl}` : this.id.toString();
         const truncatedReference = CoreBuffer.fromUtf8(
-            `${idPart}|${this.key.algorithm}|${this.key.secretKey.toBase64URL()}|${this.forIdentityTruncated ? this.forIdentityTruncated : ""}|${this.passwordType ? this.passwordType.toString() : ""}`
+            `${idPart}|${this.key.algorithm}|${this.key.secretKey.toBase64URL()}|${this.forIdentityTruncated ? this.forIdentityTruncated : ""}|${this.version ? this.version : ""}|${this.salt ? this.salt.toBase64() : ""}|${this.passwordType ? this.passwordType : ""}`
```
jkoenig134, line 50: now you introduced something we declared as a **definite** non-goal, having a 3,5,7 length. I remember that we talked about to introduce 3 for backwards compat, but all new features have to be empty or filled, so why 5 or 7 instead of just 7?

packages/transport/src/core/TransportCoreErrors.ts
```diff
@@ -206,6 +206,14 @@ class General {
...
         return new CoreError("error.transport.accountControllerInitialSyncFailed", "The initial sync of the AccountController failed.");
     }
+
+    public noPasswordProvided() {
+        return new CoreError("error.transport.noPasswordProvided", "You need to provide a password. Maybe you have given a PIN instead.");
+    }
+
+    public noPINProvided() {
+        return new CoreError("error.transport.noPINProvided", "You need to provide a PIN. Maybe you have given a password instead.");
+    }
```
jkoenig134, line 216: transport should not care about pw / pin. This is only a consumer information

packages/transport/src/modules/files/local/File.ts
```diff
@@ -49,7 +49,7 @@ export class File extends CoreSynchronizable implements IFile {
     }
 
     public toFileReference(): FileReference {
-        return FileReference.from({ id: this.id, key: this.secretKey });
+        return FileReference.from({ id: this.id, key: this.secretKey, version: 1 });
```
jkoenig134, line 52: I don't get why we need this.

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -40,6 +40,9 @@ export class RelationshipTemplateController extends TransportController {
 
     public async sendRelationshipTemplate(parameters: ISendRelationshipTemplateParameters): Promise<RelationshipTemplate> {
         parameters = SendRelationshipTemplateParameters.from(parameters);
+        if (!!parameters.password && !!parameters.pin) {
+            throw new CoreError("error.transport.notBothPasswordAndPin", "It's not possible to protect a RelationshipTemplate with both a password and a PIN.");
```
jkoenig134, line 44: CoreError in a Controller file?!

jkoenig134: also - this should not be transport logic, instead runtime

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -62,11 +65,15 @@ export class RelationshipTemplateController extends TransportController {
...
 
+        const password = parameters.password ?? parameters.pin;
+        const salt = password ? await CoreCrypto.random(16) : undefined;
+
         const backboneResponse = (
             await this.client.createRelationshipTemplate({
                 expiresAt: parameters.expiresAt.toString(),
                 maxNumberOfAllocations: parameters.maxNumberOfAllocations,
                 forIdentity: parameters.forIdentity?.address.toString(),
+                password: password ? (await CoreCrypto.deriveHashOutOfPassword(password, salt!)).toBase64() : undefined,
```
jkoenig134, line 76: very unreadable, please extract this to a variable.

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -112,8 +122,18 @@ export class RelationshipTemplateController extends TransportController {
         if (ids.length < 1) {
             return [];
         }
-
-        const resultItems = (await this.client.getRelationshipTemplates({ ids })).value;
+        const templates = await this.readRelationshipTemplates(ids);
```
jkoenig134, line 125: missing newline

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -112,8 +122,18 @@ export class RelationshipTemplateController extends TransportController {
...
+        const resultItems = (
+            await this.client.getRelationshipTemplates({
+                templates: await Promise.all(
+                    templates.map(async (t) => {
+                        const password = t.password ?? t.pin;
+                        return { id: t.id.toString(), password: password ? (await CoreCrypto.deriveHashOutOfPassword(password, t.salt!)).toBase64() : undefined };
+                    })
+                )
+            })
+        ).value;
```
jkoenig134, line 136: missing newline

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -112,8 +122,18 @@ export class RelationshipTemplateController extends TransportController {
...
-
-        const resultItems = (await this.client.getRelationshipTemplates({ ids })).value;
+        const templates = await this.readRelationshipTemplates(ids);
+
+        const resultItems = (
+            await this.client.getRelationshipTemplates({
+                templates: await Promise.all(
+                    templates.map(async (t) => {
+                        const password = t.password ?? t.pin;
+                        return { id: t.id.toString(), password: password ? (await CoreCrypto.deriveHashOutOfPassword(password, t.salt!)).toBase64() : undefined };
```
jkoenig134, line 132: hash should be a separate variable

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -124,25 +144,43 @@ export class RelationshipTemplateController extends TransportController {
...
+                    })
+                )
+            })
+        ).value.collect();
 
-        const backboneRelationships = await (await this.client.getRelationshipTemplates({ ids: ids.map((id) => id.id) })).value.collect();
+        const decryptionPromises = backboneRelationshipTemplates.map(async (t) => {
+            const template = templates.find((template) => template.id.toString() === t.id);
+            if (!template) return;
+            return { id: CoreId.from(t.id), cache: await this.decryptRelationshipTemplate(t, template.secretKey) };
```
jkoenig134, line 163: see above

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -124,25 +144,43 @@ export class RelationshipTemplateController extends TransportController {
...
+        return caches.filter((c) => c !== undefined);
+    }
+
+    private async readRelationshipTemplates(ids: string[]): Promise<RelationshipTemplate[]> {
+        const templatePromises = ids.map(async (id) => {
+            const templateDoc = await this.templates.read(id);
             if (!templateDoc) {
                 this._log.error(
-                    `Template '${t.id}' not found in local database and the cache fetching was therefore skipped. This should not happen and might be a bug in the application logic.`
+                    `Template '${id}' not found in local database. This should not happen and might be a bug in the application logic. Skipping further processing of that Template.`
```
jkoenig134, line 175: you introduced a function called `readRelationshipTemplates` in which body you log an error that if not found `Skipping further processing of that Template`. **Nothing** in the method naming indicates that there is some kind of processing that could be skipped.

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -161,7 +199,13 @@ export class RelationshipTemplateController extends TransportController {
 
     private async updateCacheOfTemplate(template: RelationshipTemplate, response?: BackboneGetRelationshipTemplatesResponse) {
         if (!response) {
-            response = (await this.client.getRelationshipTemplate(template.id.toString())).value;
+            const password = template.password ?? template.pin;
+            response = (
+                await this.client.getRelationshipTemplate(
+                    template.id.toString(),
+                    password ? (await CoreCrypto.deriveHashOutOfPassword(password, template.salt!)).toBase64() : undefined
```
jkoenig134, line 206: see above

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -227,12 +271,22 @@ export class RelationshipTemplateController extends TransportController {
         return template;
     }
 
-    public async loadPeerRelationshipTemplateByTruncated(truncated: string): Promise<RelationshipTemplate> {
+    public async loadPeerRelationshipTemplateByTruncated(truncated: string, password?: string, pin?: string): Promise<RelationshipTemplate> {
```
jkoenig134, line 274:
```suggestion
    public async loadPeerRelationshipTemplateByTruncated(truncated: string, password?: string): Promise<RelationshipTemplate> {
```

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -227,12 +271,22 @@ export class RelationshipTemplateController extends TransportController {
...
 
-    public async loadPeerRelationshipTemplate(id: CoreId, secretKey: CryptoSecretKey, forIdentityTruncated?: string): Promise<RelationshipTemplate> {
+    public async loadPeerRelationshipTemplate(
+        id: CoreId,
+        secretKey: CryptoSecretKey,
+        forIdentityTruncated?: string,
+        password?: string,
+        pin?: string,
+        salt?: CoreBuffer
+    ): Promise<RelationshipTemplate> {
```
jkoenig134, line 289: I think this function should be private as we heavily advise against loading a template by id and key.

packages/transport/src/modules/relationshipTemplates/local/RelationshipTemplate.ts
```diff
@@ -18,7 +21,7 @@ export interface IRelationshipTemplate extends ICoreSynchronizable {
 @type("RelationshipTemplate")
 export class RelationshipTemplate extends CoreSynchronizable implements IRelationshipTemplate {
     public override readonly technicalProperties = ["@type", "@context", nameof<RelationshipTemplate>((r) => r.secretKey), nameof<RelationshipTemplate>((r) => r.isOwn)];
-
+    public override readonly userdataProperties = [nameof<RelationshipTemplate>((r) => r.password)];
```
jkoenig134, line 24: pin and salt are not datawallet synced

packages/transport/src/modules/relationshipTemplates/local/RelationshipTemplate.ts
```diff
@@ -50,7 +65,20 @@ export class RelationshipTemplate extends CoreSynchronizable implements IRelatio
...
 
     public toRelationshipTemplateReference(): RelationshipTemplateReference {
-        return RelationshipTemplateReference.from({ id: this.id, key: this.secretKey, forIdentityTruncated: this.cache!.forIdentity?.toString().slice(-4) });
+        return RelationshipTemplateReference.from({
+            id: this.id,
+            key: this.secretKey,
+            forIdentityTruncated: this.cache!.forIdentity?.toString().slice(-4),
+            passwordType: this.passwordType,
+            salt: this.salt,
+            version: 1
```
jkoenig134, line 74: what the heck is this version popping up everywhere?!

jkoenig134 (changes requested):
I have so many issues here that I stop my review for now and would ask you to heavily rework this, before I continue.

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
Magnus-Kuhn: It's used when loading a template with a token

jkoenig134: Then you should add a function that allows to pass a TokenContentRelationshipTemplate instead.

packages/transport/src/core/Reference.ts
Magnus-Kuhn: What do we do if someone has created correct References with an older length (in this case 5)?

packages/transport/src/core/Reference.ts
Magnus-Kuhn: Giving a version to the reference was for easier updating of the reference (originally for if the password algorithm changes, but we decided to apply the version to the entire reference instead of just the password)

packages/transport/src/core/Reference.ts
jkoenig134: Also, IMO we can archive to bring this back to 5, but that's a next step.

packages/transport/src/core/Reference.ts
jkoenig134: This feels so unrelated to the other changes here, this should defenitely not be made in this PR.

packages/transport/src/core/Reference.ts
jkoenig134: misspelled. I meant generate.

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
Magnus-Kuhn: I'll just remove that part from the log then - otherwise I'd have to return the failed templates and then handle them, which seems too much effort for something that shouldn't happen

jkoenig134: great.

packages/runtime/src/useCases/transport/account/LoadItemFromTruncatedReference.ts
```diff
@@ -93,7 +93,7 @@ export class LoadItemFromTruncatedReferenceUseCase extends UseCase<LoadItemFromT
         const tokenContent = token.cache.content;
 
         if (tokenContent instanceof TokenContentRelationshipTemplate) {
-            const template = await this.templateController.loadPeerRelationshipTemplate(tokenContent.templateId, tokenContent.secretKey, tokenContent.forIdentity?.toString());
+            const template = await this.templateController.loadPeerRelationshipTemplateByTokenContent(tokenContent);
```
Magnus-Kuhn, line 96: Making this use case clean requires that tokens for password-protected templates are protected with the same password - I'd like to only implement this in the password-protected tokens PR.

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -124,25 +144,41 @@ export class RelationshipTemplateController extends TransportController {
...
+        const backboneRelationshipTemplates = await (
+            await this.client.getRelationshipTemplates({
+                templates: await Promise.all(
+                    templates.map(async (t) => {
+                        const hashedPassword = t.password ? (await CoreCrypto.deriveHashOutOfPassword(t.password, t.salt!)).toBase64() : undefined;
+                        return { id: t.id.toString(), password: hashedPassword };
+                    })
+                )
+            })
+        ).value.collect();
```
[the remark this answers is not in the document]
jkoenig134: IDK.

packages/runtime/src/useCases/transport/account/LoadItemFromTruncatedReference.ts
jkoenig134: This means we have to merge password-protected tokens first?

packages/runtime/test/dataViews/RelationshipTemplateDVO.test.ts
```diff
@@ -69,7 +70,7 @@ describe("RelationshipTemplateDVO", () => {
             key: "surname",
             confidentiality: "protected" as RelationshipAttributeConfidentiality
         };
-        const templateContent = RelationshipTemplateContent.from({
+        templateContent = RelationshipTemplateContent.from({
```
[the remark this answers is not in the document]
jkoenig134: you thumbed up my comment, resolved it and ignored it.

packages/transport/src/core/Reference.ts
```diff
@@ -29,14 +31,18 @@ export class Reference extends Serializable implements IReference {
...
 
-    @validate({ nullable: true, min: 1, max: 12, customValidator: (v) => (!Number.isInteger(v) ? "must be an integer" : undefined) })
+    @validate({ nullable: true, regExp: /^(pw|pin(4|5|6|7|8|9|10|11|12|13|14|15|16))$/ })
     @serialize()
-    public passwordType?: number;
+    public passwordType?: string;
+
+    @validate({ nullable: true, customValidator: (v: ICoreBuffer) => (v.buffer.byteLength === 16 ? undefined : "must be 16 bytes long") })
+    @serialize()
+    public salt?: CoreBuffer;
```
jkoenig134, line 40: maybe salt and password type should be one parameter, this would help us to not have to have the validation in the from method.

packages/transport/src/core/Reference.ts
```diff
@@ -103,8 +111,10 @@ export class Reference extends Serializable implements IReference {
     }
 
     public static from(value: IReference | string): Reference {
-        if (typeof value === "string") return this.fromTruncated(value);
-
-        return this.fromAny(value);
+        const reference = typeof value === "string" ? this.fromTruncated(value) : this.fromAny(value);
+        if (!reference.passwordType !== !reference.salt) {
+            throw TransportCoreErrors.general.invalidTruncatedReference("It's not possible to have only one of passwordType and salt set.");
+        }
```
jkoenig134, line 117: this is the wrong place for the validation. you can use postFrom for that.

packages/transport/src/modules/relationshipTemplates/local/SendRelationshipTemplateParameters.ts
```diff
@@ -27,7 +30,20 @@ export class SendRelationshipTemplateParameters extends Serializable implements
     @serialize()
     public forIdentity?: CoreAddress;
 
+    @validate({ nullable: true, disallowedValues: [""] })
+    @serialize()
+    public password?: string;
+
+    @validate({ nullable: true, regExp: /^(pw|pin(4|5|6|7|8|9|10|11|12|13|14|15|16))$/ })
+    @serialize()
+    public passwordType?: string;
```
jkoenig134, line 39: see other comment, we should combine them. This now counts for everywhere.

packages/transport/src/modules/relationshipTemplates/local/SendRelationshipTemplateParameters.ts
```diff
@@ -27,7 +30,20 @@ export class SendRelationshipTemplateParameters extends Serializable implements
...
+
     public static from(value: ISendRelationshipTemplateParameters): SendRelationshipTemplateParameters {
-        return this.fromAny(value);
+        const parameters = this.fromAny(value);
+        if (!parameters.password !== !parameters.passwordType) {
+            throw TransportCoreErrors.general.onlyOneOfPasswordTypeAndPasswordSet();
+        }
```
jkoenig134, line 45: also here - wrong place for the validation + please use serval validation errors instead of core errors for validation issues here.

packages/transport/src/modules/tokens/transmission/TokenContentRelationshipTemplate.ts
jkoenig134, line 1: see others

packages/transport/test/modules/files/FileController.test.ts
```diff
@@ -62,8 +62,8 @@ describe("FileController", function () {
         const content = CoreBuffer.fromUtf8("Test");
         const sentFile = await TestUtil.uploadFile(sender, content);
 
-        const reference = sentFile.toFileReference().truncate();
-        const receivedFile = await recipient.files.getOrLoadFileByTruncated(reference);
+        const reference = sentFile.toFileReference();
+        const receivedFile = await recipient.files.getOrLoadFileByTruncated(reference.truncate());
```
jkoenig134, line 66: why?

jkoenig134 (changes requested):
again some structural issues

packages/runtime/src/useCases/transport/relationshipTemplates/RelationshipTemplateMapper.ts
```diff
@@ -28,6 +29,12 @@ export class RelationshipTemplateMapper {
         return responseItems.map((i) => this.toRelationshipTemplateDTO(i));
     }
 
+    private static toPassword(password?: string, passwordType?: string) {
```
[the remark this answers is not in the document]
Magnus-Kuhn: You requested this format, I'm happy to change it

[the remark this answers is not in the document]
jkoenig134: Maybe we talked past each other.. I'll think about it.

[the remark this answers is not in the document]
jkoenig134: maybe an object

```json
passwordProtection: {
  password: string
  passwordType: string
}
```

would be better

packages/runtime/src/useCases/transport/account/LoadItemFromTruncatedReference.ts
jkoenig134: Maybe it's enough to release both things together as this PR seems to come to an end.

packages/transport/test/modules/files/FileController.test.ts
Magnus-Kuhn: IDK how this happened

jkoenig134 (dismissed):

jkoenig134 (dismissed):
LGTM. @britsta and @Milena-Czierlinski please re-check the tests.

packages/runtime/test/transport/relationshipTemplates.test.ts
```diff
@@ -247,6 +248,173 @@ describe("RelationshipTemplate Tests", () => {
...
+                    password: "password"
+                }
+            });
+            expect(createResult).toBeSuccessful();
+
+            const loadResult = await runtimeServices2.transport.relationshipTemplates.loadPeerRelationshipTemplate({
+                reference: createResult.value.truncatedReference,
+                password: "wrong-password"
+            });
+            expect(loadResult).toBeAnError(/.*/, "error.runtime.recordNotFound");
```
Milena-Czierlinski, line 358: If the password is wrong, we'll throw a `recordNotFound` error with the message `'RelationshipTemplate not found. Make sure the ID exists and the record is not expired.'`. We should be more precise with the error, such that someone doesn't think the RelationshipTemplate is the problem if they provide a wrong password. If no password is provided, we'll throw an error `error.transport.noPasswordProvided` with message `"You need to provide a password to perform this operation."`. That means, we already expose the information that the RelationshipTemplate requires a password. Thus, I suggest to add an error that informs about a wrong password.

jkoenig134: We can not do this. The backbone doesn't give more information..

Magnus-Kuhn: What about expanding the error message to also include a failed password check as a cause?

jkoenig134: Hmmmmmm, forget about my comment. 

1. we know that the thing is pw protected
2. we get a 404 from the bkb
3. we COULD say that there is a possibility that
  a. the thing is expired
  b. the password is incorrect

But we should tell about both issues.

jkoenig134: > What about expanding the error message to also include a failed password check as a potential cause?

Yeah, but please only expand if we think that is's password protected.

Milena-Czierlinski: > We can not do this. The backbone doesn't give more information..

Should we request this from the Backbone then? It would be good to be able to differentiate these cases.

> What about expanding the error message to also include a failed password check as a potential cause?

That would be an alternative, if the Backbone has valid reasons not to give us more information. We could in fact only expand the error message if we know that the reference has a passwordProtection.

jkoenig134: And - please check the message instead of doing `/.*/` then.

jkoenig134: > Should we request this from the Backbone then? It would be good to be able to differentiate these cases.
> 
> >

It's a FEATURE, that we (and noone else) don't get the information.

Milena-Czierlinski: I thought the Backbone blocks the RelationshipTemplate after a certain amount of failed attempts to avoid bruteforce.

jkoenig134: Not yet. But this would only be an additional security mechanism. It's an antipattern to say a password is wrong.

Milena-Czierlinski: Alright, then let's expand the error message in case we have a passwordProtection and get an error from the Backbone. 👍

jkoenig134: alrighty schmighty!

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -251,9 +329,28 @@ export class RelationshipTemplateController extends TransportController {
...
+                    e.method,
+                    e.path,
+                    e.platformParameters,
+                    e.code,
+                    "RelationshipTemplate not found. Make sure the ID exists, the record is not expired and the entered password is correct.",
+                    e.requestId,
+                    e.status,
+                    e.time,
+                    e.object
+                );
```
jkoenig134, line 350: I don't like this approach. Why do you throw a RequestError instead of a record not found by the runtime? Just altering one thing from the backbone and make it look like the backbone outputs another error message is very weird looking and very hard to understand.

jkoenig134 (approved):
LGTM

PR #300, opened 2024-10-18
Password-protected Tokens

Magnus-Kuhn:
# Readiness checklist

-   [X] I labeled the PR.


packages/runtime/src/useCases/transport/relationshipTemplates/CreateTokenQRCodeForOwnRelationshipTemplate.ts
```diff
@@ -52,7 +53,7 @@ export class CreateTokenQRCodeForOwnTemplateUseCase extends UseCase<CreateTokenQ
             return Result.fail(RuntimeErrors.relationshipTemplates.personalizationMustBeInherited());
         }
 
-        if (template.passwordProtection && request.passwordProtection !== template.passwordProtection) {
+        if (template.passwordProtection && !_.isEqual(request.passwordProtection, template.passwordProtection)) {
```
jkoenig134, line 56: I'd prefer an isEqual method on the PasswordProtection object. Comparing classes with lodash's isEqual should be avoided.

Magnus-Kuhn: I see, then some isEqual checks in the AttributeController need a JSON conversion first, I'm on it

Magnus-Kuhn:
Some TODOs not in this PR are:
* Allowing querying for password-protection in the runtime (I forgot it with the templates, so I thought to do them separately for both templates and tokens)
* Completing the refactoring where validations for `expiresAt` and `passwordProtection` are moved into the SchemaValidator by renaming the SchemaValidator

packages/runtime/src/useCases/common/validation/SchemaValidator.ts
```diff
@@ -1,17 +1,34 @@
...
+            const passwordProtection = input.passwordProtection as { password: string; passwordIsPin?: true };
+
+            if (passwordProtection.passwordIsPin) {
+                if (!/^[0-9]{4,16}$/.test(passwordProtection.password)) {
+                    validationResult.addFailure(new ValidationFailure(RuntimeErrors.general.invalidPin()));
+                }
+            }
+        }
+
+        return validationResult;
```
jkoenig134, line 31: What's going on here? Why do you put logic for some usescases in the baseclass?! It would be way cleaner if you just extend the SchemaValidator and put it there any only use the new extended version where required.

packages/runtime/src/useCases/transport/account/LoadItemFromTruncatedReference.ts
jkoenig134, line 1: you forgot to pass the password down to the `loadPeerRelationshipTemplateByTokenContent` and `getOrLoadFile` methods. I'm **very** curious why this didn't popped up in a test, please double check that you test this usecase with a double-password-protected item.

packages/runtime/src/useCases/transport/files/UploadOwnFile.ts
```diff
@@ -53,15 +53,6 @@ class Validator extends SchemaValidator<UploadOwnFileValidatableRequest> {
...
 
-        if (input.expiresAt && CoreDate.from(input.expiresAt).isSameOrBefore(CoreDate.utc())) {
-            validationResult.addFailure(
-                new ValidationFailure(
-                    RuntimeErrors.general.invalidPropertyValue(`'${nameof<UploadOwnFileValidatableRequest>((r) => r.expiresAt)}' must be in the future`),
-                    nameof<UploadOwnFileValidatableRequest>((r) => r.expiresAt)
-                )
-            );
-        }
-
```
jkoenig134, line 64: I don't like, that you squeezed in a major refactoring into this PR. This has nothing to do with the actual PR.

packages/runtime/src/useCases/transport/relationshipTemplates/CreateTokenForOwnRelationshipTemplate.ts
```diff
@@ -51,13 +59,17 @@ export class CreateTokenForOwnTemplateUseCase extends UseCase<CreateTokenForOwnT
             return Result.fail(RuntimeErrors.relationshipTemplates.personalizationMustBeInherited());
         }
 
-        const passwordProtection = template.passwordProtection ? SharedPasswordProtection.from(template.passwordProtection) : undefined;
+        if (template.passwordProtection && !template.passwordProtection.matchesInputForNewPassword(request.passwordProtection)) {
+            return Result.fail(RuntimeErrors.relationshipTemplates.passwordProtectionMustBeInherited());
+        }
+
+        const templatePasswordProtection = template.passwordProtection ? SharedPasswordProtection.from(template.passwordProtection) : undefined;
```
jkoenig134, line 66: why did you rename this? This results in a way less clean usage below.

packages/runtime/src/useCases/transport/relationshipTemplates/CreateTokenQRCodeForOwnRelationshipTemplate.ts
```diff
@@ -44,13 +58,17 @@ export class CreateTokenQRCodeForOwnTemplateUseCase extends UseCase<CreateTokenQ
             return Result.fail(RuntimeErrors.relationshipTemplates.personalizationMustBeInherited());
         }
 
-        const passwordProtection = template.passwordProtection ? SharedPasswordProtection.from(template.passwordProtection) : undefined;
+        if (template.passwordProtection && !template.passwordProtection.matchesInputForNewPassword(request.passwordProtection)) {
+            return Result.fail(RuntimeErrors.relationshipTemplates.passwordProtectionMustBeInherited());
+        }
+
+        const templatePasswordProtection = template.passwordProtection ? SharedPasswordProtection.from(template.passwordProtection) : undefined;
```
jkoenig134, line 65: see above

packages/runtime/src/useCases/transport/relationshipTemplates/LoadPeerRelationshipTemplate.ts
```diff
@@ -55,7 +55,7 @@ export class LoadPeerRelationshipTemplateUseCase extends UseCase<LoadPeerRelatio
     }
 
     private async loadRelationshipTemplateFromTokenReference(tokenReference: string, password?: string): Promise<Result<RelationshipTemplateDTO>> {
-        const token = await this.tokenController.loadPeerTokenByTruncated(tokenReference, true);
+        const token = await this.tokenController.loadPeerTokenByTruncated(tokenReference, true, password);
```
jkoenig134, line 58: please write a test for this, it's a pity that you didn't discovered this in the other PR by writing proper tests.

packages/runtime/test/lib/testUtils.ts
```diff
@@ -160,12 +160,13 @@ export async function syncUntilHasEvent<TEvent extends Event>(
     return event;
 }
 
-export async function uploadOwnToken(transportServices: TransportServices, forIdentity?: string): Promise<TokenDTO> {
+export async function uploadOwnToken(transportServices: TransportServices, forIdentity?: string, password?: string): Promise<TokenDTO> {
```
jkoenig134, line 163: when just calling this password, you completely loose the possibility to properly test this for example with PINs. Why not allow to pass an object with password and passwordIsPin? Would also make the usage below cleaner.

packages/runtime/test/transport/account.test.ts
```diff
@@ -81,11 +81,18 @@ describe("LoadItemFromTruncatedReference", () => {
...
         beforeAll(async () => {
             const file = await uploadFile(sTransportServices);
             fileReference = file.truncatedReference;
             fileTokenReference = (await sTransportServices.files.createTokenForFile({ fileId: file.id })).value.truncatedReference;
+            passwordProtectedFileTokenReference = (
+                await sTransportServices.files.createTokenForFile({
+                    fileId: file.id,
+                    passwordProtection: { password: "password" }
+                })
+            ).value.truncatedReference;
```
jkoenig134, line 95: Tests should be independent, please create a reference for each test and don't clutter up the test more than it is currently.

packages/runtime/test/transport/relationshipTemplates.test.ts
jkoenig134, line 1: I said it many times before - when adding a new feature it's an antipattern to just stuff in that new feature in old tests and weaken them by this.

We should test all combinations, what you do doesn't help with that.

jkoenig134: what you do:

1. test a password-protected template in a password-protected token

what you should do:

1. test a non-protected template in a password-protected token
2. test a password-protected template in a password-protected token
3. test a password-protected template in a non-protected token (is that even possible? how should we handle this?)

packages/runtime/test/transport/tokens.test.ts
```diff
@@ -94,3 +95,244 @@ describe("Personalized tokens", () => {
...
+        expect(createResult).toBeAnError(/.*/, "error.runtime.validation.invalidPin");
+    });
+});
+
+describe("Password-protected tokens for files", () => {
+    let fileId: string;
+
+    beforeAll(async () => {
+        fileId = (await uploadFile(runtimeServices1.transport)).id;
+    });
```
jkoenig134, line 195: please create a file for each test. Then you are way more flexible to test all combinations (see above for more context)

packages/transport/src/core/types/PasswordProtection.ts
```diff
@@ -31,4 +32,15 @@ export class PasswordProtection extends Serializable implements IPasswordProtect
             salt: this.salt
         });
     }
+
+    private matchesCreationParameters(compare: PasswordProtectionCreationParameters): boolean {
```
jkoenig134, line 36: `compare` is not a good wording. wtb just `creationParameters` to match the function naming?

packages/transport/src/core/types/PasswordProtection.ts
```diff
@@ -31,4 +32,15 @@ export class PasswordProtection extends Serializable implements IPasswordProtect
...
+        return this.passwordType === compare.passwordType && this.password === compare.password;
+    }
+
+    public matchesInputForNewPassword(compare: { password: string; passwordIsPin?: true } | undefined): boolean {
```
jkoenig134, line 40: see above: `newPasswordProtection` would be a better naming for `compare`. Also the method name could benefit from the word `Protection` at the end to make clearer that this is not ONLY about passwords and also about pins.

packages/transport/src/core/types/PasswordProtection.ts
```diff
@@ -31,4 +32,15 @@ export class PasswordProtection extends Serializable implements IPasswordProtect
...
+        const compareCreationParameters = PasswordProtectionCreationParameters.create(compare);
+        if (!compareCreationParameters) return false;
+
+        return this.matchesCreationParameters(compareCreationParameters);
```
jkoenig134, line 44: as this function calls the other private function the private function has to be BELOW this. You can read more about this in our coding guidelines.

jkoenig134 (changes requested):

packages/runtime/src/useCases/transport/account/LoadItemFromTruncatedReference.ts
Magnus-Kuhn: Password protection for files only works via tokens, so `getOrLoadFile` doesn't get a password - and the test did not have the double password-protection, so the error in the `relationshipTemplateTokenContent` was uncaught

jkoenig134: I see.

packages/runtime/src/useCases/common/validation/SchemaValidator.ts
```diff
@@ -5,7 +5,7 @@ import { IValidator } from "./IValidator";
 import { ValidationFailure } from "./ValidationFailure";
 import { ValidationResult } from "./ValidationResult";
 
-export class SchemaValidator<T> implements IValidator<T> {
+export class SchemaValidator<T extends object> implements IValidator<T> {
```
jkoenig134:
```suggestion
export class SchemaValidator<T> implements IValidator<T> {
```

packages/runtime/src/useCases/common/validation/GenericInputValidator.ts
```diff
@@ -0,0 +1,27 @@
+import { RuntimeErrors } from "../RuntimeErrors";
+import { JsonSchema } from "../SchemaRepository";
+import { SchemaValidator } from "./SchemaValidator";
+import { ValidationFailure } from "./ValidationFailure";
+import { ValidationResult } from "./ValidationResult";
+
+export class GenericInputValidator<T extends object> extends SchemaValidator<T> {
```
jkoenig134: I'm very confused, could you elaborate the naming here? Why is this InputValidator more GENERIC than the class that it extends?

jkoenig134: Also, Schema is missing in the name.

packages/runtime/src/useCases/transport/files/CreateTokenForFile.ts
```diff
@@ -1,22 +1,57 @@
...
+        if (!validationResult.isValid()) return validationResult;
+
+        if (input.expiresAt && DateTime.fromISO(input.expiresAt) <= DateTime.utc()) {
+            validationResult.addFailure(
+                new ValidationFailure(
+                    RuntimeErrors.general.invalidPropertyValue(`'${nameof<CreateTokenForFileRequest>((r) => r.expiresAt)}' must be in the future`),
+                    nameof<CreateTokenForFileRequest>((r) => r.expiresAt)
+                )
+            );
+        }
```
jkoenig134: so this is perfectly fine - but does not belong in this PR.

jkoenig134: as it's done already keep it.

jkoenig134: Hm. Now I see another issue with the implementation - which is **exactly** the reason why this should be in another PR

1. use CoreDate instead of DateTime
2. use `CoreDate#isExpired()` to check against `DateTime.now()`

packages/runtime/src/useCases/transport/files/CreateTokenQRCodeForFile.ts
```diff
@@ -1,23 +1,59 @@
...
+        if (!validationResult.isValid()) return validationResult;
+
+        if (input.expiresAt && DateTime.fromISO(input.expiresAt) <= DateTime.utc()) {
+            validationResult.addFailure(
+                new ValidationFailure(
+                    RuntimeErrors.general.invalidPropertyValue(`'${nameof<CreateTokenQRCodeForFileRequest>((r) => r.expiresAt)}' must be in the future`),
+                    nameof<CreateTokenQRCodeForFileRequest>((r) => r.expiresAt)
+                )
+            );
+        }
```
jkoenig134: see other comment

packages/runtime/src/useCases/transport/relationshipTemplates/CreateTokenForOwnRelationshipTemplate.ts
```diff
@@ -51,6 +87,10 @@ export class CreateTokenForOwnTemplateUseCase extends UseCase<CreateTokenForOwnT
             return Result.fail(RuntimeErrors.relationshipTemplates.personalizationMustBeInherited());
         }
 
+        if (template.passwordProtection && !template.passwordProtection.matchesInputForNewPasswordProtection(request.passwordProtection)) {
+            return Result.fail(RuntimeErrors.relationshipTemplates.passwordProtectionMustBeInherited());
+        }
```
jkoenig134, line 64: Hm, a more broader question that get's in my head while reading this:

Theoretically it's possible now to put a passwordProtected template inside of a non-protected token

1. do we want that?
2. will we run into issues with that? I don't really know how to handle this..

packages/runtime/src/useCases/common/validation/TokenAndTemplateCreationValidator.ts
```diff
@@ -0,0 +1,34 @@
+import { DateTime } from "luxon";
+import { RuntimeErrors } from "../RuntimeErrors";
+import { JsonSchema } from "../SchemaRepository";
+import { SchemaValidator } from "./SchemaValidator";
+import { ValidationFailure } from "./ValidationFailure";
+import { ValidationResult } from "./ValidationResult";
+
+export class TokenAndTemplateCreationValidator<T extends object> extends SchemaValidator<T> {
```
jkoenig134, line 8:
```suggestion
export class TokenAndTemplateCreationValidator<
    T extends {
        expiresAt?: ISO8601DateTimeString;
        passwordProtection?: {
            password: string;
            passwordIsPin?: true;
        };
    }
> extends SchemaValidator<T> {
```

packages/runtime/src/useCases/common/validation/TokenAndTemplateCreationValidator.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public override validate(input: T): ValidationResult {
+        const validationResult = super.validate(input);
+
+        if ("expiresAt" in input && input.expiresAt) {
+            if (DateTime.fromISO(input.expiresAt as string) <= DateTime.utc()) {
+                validationResult.addFailure(new ValidationFailure(RuntimeErrors.general.invalidPropertyValue(`'expiresAt' must be in the future`), "expiresAt"));
+            }
+        }
```
jkoenig134, line 20:
```suggestion
        if (input.expiresAt && CoreDate.from(input.expiresAt).isExpired()) {
            validationResult.addFailure(new ValidationFailure(RuntimeErrors.general.invalidPropertyValue(`'expiresAt' must be in the future`), "expiresAt"));
        }
```

jkoenig134 (changes requested):

jkoenig134:
It seems like you have committet some things during my review - please make sure to read all comments, also if they are marked as outdated as this can be a false positive.

jkoenig134:
Also this PR conflicts.

packages/runtime/src/useCases/transport/relationshipTemplates/CreateTokenForOwnRelationshipTemplate.ts
Magnus-Kuhn: `matchesInputForNesPasswordProtection` will return false if the new password protection is undefined - and for a protected template inside an unprotected token, the reference from the token would not reveal that a password is required. Hence I want to block that

jkoenig134: Oh, I misunderstood this. Looks good then.

packages/runtime/src/useCases/common/validation/TokenAndTemplateCreationValidator.ts
```diff
@@ -0,0 +1,41 @@
...
+        if (input.expiresAt && CoreDate.from(input.expiresAt).isExpired()) {
+            validationResult.addFailure(new ValidationFailure(RuntimeErrors.general.invalidPropertyValue(`'expiresAt' must be in the future`), "expiresAt"));
+        }
+
+        if (input.passwordProtection) {
+            const passwordProtection = input.passwordProtection;
+
+            if (passwordProtection.passwordIsPin) {
+                if (!/^[0-9]{4,16}$/.test(passwordProtection.password)) {
+                    validationResult.addFailure(new ValidationFailure(RuntimeErrors.general.invalidPin()));
```
jkoenig134, line 34: if I look up at the other error I think you should use `RuntimeErrors.general.invalidPropertyValue` instead of a new error

packages/runtime/test/transport/tokens.test.ts
jkoenig134: I think you misunderstood this comment. I really meant just an **enmeshed file** not test.ts files, but I think it's ok like it is now?
@Milena-Czierlinski wdym?

packages/runtime/src/useCases/common/validation/GenericInputValidator.ts
jkoenig134: This is not fully resolved.

jkoenig134 (approved):
LGTM. Please wait for another review by @britsta or @Milena-Czierlinski

Milena-Czierlinski (approved):
Please make sure to also test for the error message of `recordNotFound` once the Backbone has adjusted it.

jkoenig134:
@Milena-Czierlinski typically we don't test error messages, only codes.

jkoenig134 (approved):

jkoenig134:
@Magnus-Kuhn anything that blocks this from merging?

PR #307, opened 2024-10-23
Provide backbone-defined valid tags for Attributes

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.


sebbi08:
You can already review. The backbone route is not yet finished, so this PR will still be a draft

Magnus-Kuhn:
Getting the list of valid tags doesn't really require being logged in, on the other hand they're only supposed to be used for attributes which currently can only be shared between Identities. So I think it's ok that they require a login.

But I'd expect that the list is not fetched from the backbone every time you want to do something with a tag (which is often), but rather that the backbone sends the list whenever it is changed (e. g. via an event) and it's then stored locally.

jkoenig134:
I'll not review a Draft. Please re-assign when finished.

packages/runtime/src/extensibility/facades/transport/TagsFacade.ts
```diff
@@ -0,0 +1,12 @@
+import { Result } from "@js-soft/ts-utils";
+import { BackboneGetTag } from "@nmshd/transport";
```
jkoenig134, line 2: this is very lazy, why do you use a transport type instead of creating a DTO like everywhere else?

sebbi08: Is the mapper enough, or should I add a serval class ?

packages/transport/test/modules/tags/TagsController.test.ts
```diff
@@ -0,0 +1,33 @@
...
+        account = accounts[0];
+    });
+
+    afterAll(async function () {
+        await account.close();
+
+        await connection.close();
+    });
+
+    test("should receive the legal tags", async function () {
```
britsta, line 28: Emphasize that the legal tags are determined by the Backbone.

packages/transport/test/modules/tags/TagsController.test.ts
```diff
@@ -0,0 +1,33 @@
+import { IDatabaseConnection } from "@js-soft/docdb-access-abstractions";
+import { AccountController, Transport } from "../../../src";
+import { TestUtil } from "../../testHelpers/TestUtil";
+
+describe("AccountController", function () {
```
britsta, line 5: "TagsController" instead of "AccountController".

packages/transport/src/modules/tags/TagController.ts
britsta, line 1: TagsController or TagController? The test file is called TagsController.test.ts, so there is an inconsistency here. Also keep in mind that you have defined a TagsFacade and not a TagFacade. 

When making this more consistent (decide for singular form or plural form), please ensure that the file names and the class names are adapted accordingly.

jkoenig134: Tags please :)

packages/transport/src/modules/tags/TagController.ts
```diff
@@ -0,0 +1,25 @@
...
+    public override async init(): Promise<this> {
+        await super.init();
+
+        this.client = new TagClient(this.config, this.parent.authenticator, this.transport.correlator);
+
+        return this;
+    }
+
+    public async getTags(): Promise<BackboneTagList> {
+        const tags = (await this.client.getTags()).value;
```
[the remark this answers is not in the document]
sebbi08: The value getter of the ClientResult will do this on his own

packages/runtime/src/useCases/transport/tags/TagMapper.ts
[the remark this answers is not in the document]
sebbi08: hi yea as the BackboneTagList and TagListDTO are typewise the exactly the same we do not actually do any mapping.

packages/transport/src/modules/tags/TagController.ts
Milena-Czierlinski: I noticed this inconsistency, too, but saw that it is consistent with the way we have things right now. All Facades use the plural form and all Controllers the singular form. But of course, if we don't refactor this right now, the test for the Controller should then also use the singular form.

Milena-Czierlinski: As stated, now we have an inconsistency with all the other Controllers that use the singular form. @jkoenig134 Do we really want this?

jkoenig134: You mean AttributeListener**s**Controller, Attribute**s**Controller, Draft**s**Controller, Notification**s**Controller, OutgoingRequest**s**Controller and Setting**s**Controller


TagsController is perfectly fine.

Milena-Czierlinski: As this is in the transport package, I was of course referring to the TokenController, RelationshipTemplateController, MessageController etc. . But obviously we are not consistent everywhere right now, so I don't have a problem with deciding for the plural adding new Controllers.

jkoenig134: Transport is older and uses the old Version (without s), Consumption is consistently using the correct form.

packages/transport/test/modules/tags/TagsController.test.ts
Milena-Czierlinski: should receive the legal tags **the B**ackbone

Milena-Czierlinski (commented):
Just a tiny thing, the rest LGTM. 

@jkoenig134 Can you please take a look at your last open comment?

packages/transport/test/modules/tags/TagsController.test.ts
jkoenig134: ehm, what?

Milena-Czierlinski: Argh copy and paste error, but @sebbi08 got it. :D

packages/transport/src/modules/tags/TagsController.ts
```diff
@@ -18,8 +18,8 @@ export class TagsController extends TransportController {
         return this;
     }
 
-    public async getTags(): Promise<BackboneTagList> {
-        const tags = (await this.client.getTags()).value;
-        return tags;
+    public async getTags(): Promise<TagList> {
+        const backBoneTagList = (await this.client.getTags()).value;
```
Milena-Czierlinski, line 22: "Backbone" is one word, so you shouldn't apply camelCasing within it.

Milena-Czierlinski (commented):
I'm not completely sure if this is what @jkoenig134 intended, but maybe the fault lies with me. @jkoenig134 Can you please take another look. :)

packages/runtime/src/extensibility/facades/transport/TagsFacade.ts
```diff
@@ -0,0 +1,12 @@
+import { Result } from "@js-soft/ts-utils";
+import { Inject } from "@nmshd/typescript-ioc";
+import { TagListDTO } from "../../../types/transport/TagDTO";
+import { GetTagsUseCase } from "../../../useCases/transport/tags/GetTags";
```
jkoenig134, line 4: barrel please

packages/runtime/src/useCases/transport/tags/GetTags.ts
```diff
@@ -0,0 +1,17 @@
...
+import { UseCase } from "../../common";
+import { TagMapper } from "./TagMapper";
+
+export class GetTagsUseCase extends UseCase<void, TagListDTO> {
+    public constructor(@Inject private readonly tagController: TagsController) {
+        super();
+    }
+
+    protected async executeInternal(): Promise<Result<TagListDTO>> {
+        const tagListDTO = TagMapper.toTagListDTO(await this.tagController.getTags());
```
jkoenig134, line 14: 2 lines or mapping in Result.ok

packages/runtime/src/useCases/transport/tags/TagMapper.ts
```diff
@@ -0,0 +1,8 @@
+import { TagList } from "@nmshd/transport/src/modules/tags/data/TagList";
+import { TagListDTO } from "../../../types/transport/TagDTO";
+
+export class TagMapper {
+    public static toTagListDTO(tagList: TagList): TagListDTO {
+        return tagList;
```
jkoenig134, line 6: this is all the mapping required? I doubt it.

packages/transport/src/core/TransportController.ts
```diff
@@ -29,7 +29,8 @@ export enum ControllerName {
     RelationshipTemplator = "RelationshipTemplator",
     Secret = "Secret",
     Sync = "Sync",
-    Token = "Token"
+    Token = "Token",
+    Tag = "Tag"
```
jkoenig134, line 33:
```suggestion
    Tag = "Tag",
    Token = "Token"
```

.dev/appsettings.override.json
```diff
@@ -103,6 +103,73 @@
                     "ConnectionString": "User ID=tokens;Password=Passw0rd;Server=postgres;Port=5432;Database=enmeshed;"
                 }
             }
+        },
+        "Tags": {
```
jkoenig134, line 107: as in the other PR, this should be mocked instead of hardcoded.

packages/transport/test/modules/tags/TagsController.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    afterAll(async function () {
+        await account.close();
+
+        await connection.close();
+    });
+
+    test("should receive the legal tags from the Backbone", async function () {
+        const tags = await account.tags.getTags();
+
+        expect(tags).toBeDefined();
```
jkoenig134, line 31: this expect is meaningless

jkoenig134: (that whole test has to be re-written with mocks)

jkoenig134 (changes requested):

packages/transport/src/modules/tags/data/TagList.ts
```diff
@@ -0,0 +1,21 @@
+import { Serializable, serialize, validate } from "@js-soft/ts-serval";
+
+export class TagList extends Serializable {
+    @serialize()
+    @validate()
+    public supportedLanguages: string[];
+
+    @serialize()
+    @validate()
+    public tagsForAttributeValueTypes: Record<string, Record<string, Tag>>;
```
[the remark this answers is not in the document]
sebbi08: Don't underestimate your work :D seems like it is working ^^

packages/runtime/src/useCases/transport/tags/GetTags.ts
```diff
@@ -0,0 +1,17 @@
+import { Result } from "@js-soft/ts-utils";
+import { TagsController } from "@nmshd/transport";
+import { Inject } from "@nmshd/typescript-ioc";
+import { TagListDTO } from "../../../types/transport/TagDTO";
+import { UseCase } from "../../common";
+import { TagMapper } from "./TagMapper";
+
+export class GetTagsUseCase extends UseCase<void, TagListDTO> {
+    public constructor(@Inject private readonly tagController: TagsController) {
```
Milena-Czierlinski, line 9:
```suggestion
    public constructor(@Inject private readonly tagsController: TagsController) {
```
😬

packages/runtime/test/lib/RestClientMocker.ts
```diff
@@ -0,0 +1,20 @@
...
+    public mockMethod(fnName: keyof T, callback: () => ClientResult<any>): void {
+        this.mockedMethods.push({ fnName, originalProperty: this.client[fnName] });
+
+        const mockFn = jest.fn().mockImplementation(() => {
+            return Promise.resolve(callback());
+        });
+        this.client[fnName] = mockFn.bind(this.client);
+    }
+
+    public restore(): void {
```
Milena-Czierlinski, line 15:
```suggestion
    public reset(): void {
```
😇

packages/runtime/test/transport/tags.test.ts
```diff
@@ -0,0 +1,99 @@
...
+                        de: "Notfallkontakt",
+                        en: "Emergency Contact"
+                    }
+                }
+            },
+            StreetAddress: {
+                lieferung: {
+                    displayNames: {
+                        de: "Lieferadresse",
+                        en: "Deliver Address"
```
Milena-Czierlinski, line 77:
```suggestion
                        en: "Delivery Address"
```

packages/transport/test/modules/tags/TagsController.test.ts
```diff
@@ -0,0 +1,110 @@
...
+                        de: "Notfallkontakt",
+                        en: "Emergency Contact"
+                    }
+                }
+            },
+            StreetAddress: {
+                lieferung: {
+                    displayNames: {
+                        de: "Lieferadresse",
+                        en: "Deliver Address"
```
Milena-Czierlinski, line 69:
```suggestion
                        en: "Delivery Address"
```

packages/transport/test/testHelpers/RestClientMocker.ts
```diff
@@ -0,0 +1,20 @@
...
+    public mockMethod(fnName: keyof T, callback: () => ClientResult<any>): void {
+        this.mockedMethods.push({ fnName, originalProperty: this.client[fnName] });
+
+        const mockFn = jest.fn().mockImplementation(() => {
+            return Promise.resolve(callback());
+        });
+        this.client[fnName] = mockFn.bind(this.client);
+    }
+
+    public restore(): void {
```
Milena-Czierlinski, line 15:
```suggestion
    public reset(): void {
```

packages/runtime/test/transport/tags.test.ts
```diff
@@ -0,0 +1,99 @@
...
+            PhoneNumber: {
+                notfall: {
+                    displayNames: {
+                        de: "Notfallkontakt",
+                        en: "Emergency Contact"
+                    }
+                }
+            },
+            StreetAddress: {
+                lieferung: {
```
jkoenig134, line 74: why is there german wording?

packages/runtime/src/useCases/transport/tags/TagMapper.ts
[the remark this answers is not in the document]
sebbi08: How about TagList => TagsDefinition and Tag =>  TagDefinition.
Or TagList => DefinedTags and Tag => Tag.

And yes Tag and TagList only exist together, that's why I always added them to the same files. Actually we never work with Tags at all they are only used in the TagList.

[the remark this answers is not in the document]
Milena-Czierlinski: >  TagList => TagsDefinition and Tag => TagDefinition

This is even more similar. 😨 

> TagList => DefinedTags and Tag => Tag

I would be fine with this. Also, if somebody's heart is set on TagList, I would be open to discuss this again. @jkoenig134 Do you have an opinion on this matter?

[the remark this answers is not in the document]
jkoenig134: Yeah, I have quite a strong opinion about the whole structure of this PR (Tags that will be used only by Attributes are managed in the Transport layer even when there is nothing "transported")

Maybe we can move everything to the AttributesController and manage the whole tag thing in Attributes. This would hopefully also simplify the naming.

This is only an idea, please don't do anything about this until it is discussed properly.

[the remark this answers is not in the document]
Milena-Czierlinski: > This would hopefully also simplify the naming.

Oh, I overread this part. 

@jkoenig134 I see where you are coming from wanting to put this closer to the Attributes. However, I have difficulties seeing how this would simplify things. Since we get the tags from the Backbone, we'll need the TagClient in transport and presumingly also the TagsController. Consequently, I would expect to find the use case for getting the Backbone-defined tags  also in the transport folder of the runtime package. "Hiding" this within the AttributesController could be confusing for us. 

And what exactly does "transport" refer to? Transport between an Identity and the Backbone or transport between two Identities? Because we do have other things in transport that don't necessarily concern another Identity, e.g. IdentityDeletionProcess, Files and Backbone compatibility.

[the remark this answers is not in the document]
jkoenig134: IMHO its just a coincidence that all *Clients are currently in the transport lib - I don't see a reason why the TagClient shouldn't be inside the Consumption library.

> And what exactly does "transport" refer to? 

Transport between an Identity and the Backbone

packages/transport/src/modules/tags/backbone/TagClient.ts
Milena-Czierlinski, line 1: Please rename the file AttributeTagClient.

packages/transport/src/modules/tags/backbone/BackboneGetTag.ts
```diff
@@ -0,0 +1,9 @@
+export interface BackboneAttributeTagCollection {
```
Milena-Czierlinski, line 1: Can we omit the `Backbone` here in the name and simply call it `AttributeTagCollection`? Same below. And same for the function name of the Client.

packages/transport/src/modules/tags/backbone/BackboneGetTag.ts
Milena-Czierlinski, line 1: File name

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1290,4 +1293,9 @@ export class AttributesController extends ConsumptionBaseController {
 
         return ownSharedAttributeSuccessors;
     }
+
+    public async getAttributeTagCollection(): Promise<AttributeTagCollection> {
+        const backboneAttributeTagCollection = (await this.attributeTagClient.getBackboneAttributeTagCollection()).value;
+        return AttributeTagCollection.fromAny(backboneAttributeTagCollection);
```
Milena-Czierlinski, line 1299: You can add a from-method to AttributeTagCollection and use that. I think this is how we usually do it.

packages/consumption/src/modules/attributes/local/AttributeTagCollection.ts
Milena-Czierlinski, line 1: Usually, we have an interface (or two).

packages/runtime/src/useCases/consumption/attributes/GetAttributeTagCollection.ts
```diff
@@ -0,0 +1,17 @@
...
+import { AttributeTagCollectionMapper } from "./AttributeTagCollectionMapper";
+
+export class GetAttributeTagCollectionUseCase extends UseCase<void, AttributeTagCollectionDTO> {
+    public constructor(@Inject private readonly attributesController: AttributesController) {
+        super();
+    }
+
+    protected async executeInternal(): Promise<Result<AttributeTagCollectionDTO>> {
+        const tagList = await this.attributesController.getAttributeTagCollection();
+        return Result.ok(AttributeTagCollectionMapper.toAttributeTagCollectionDTO(tagList));
```
Milena-Czierlinski, line 15:
```suggestion
        const attributeTagCollection = await this.attributesController.getAttributeTagCollection();
        return Result.ok(AttributeTagCollectionMapper.toAttributeTagCollectionDTO(attributeTagCollection));
```

packages/transport/src/modules/tags/backbone/BackboneGetTag.ts
sebbi08: We have that for all the types coming from the backbone

packages/consumption/test/modules/attributes/AttributeTagCollection.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        await accountController.close();
+
+        await connection.close();
+    });
+
+    test("should receive the legal tags from the Backbone", async function () {
+        when(mockedClient.getAttributeTagCollection()).thenResolve(ClientResult.ok(mockTags));
+        const tags = await consumptionController.attributes.getAttributeTagCollection();
+
+        expect(tags.toJSON()).toStrictEqualExcluding(mockTags, "@type");
```
Milena-Czierlinski, line 75:
```suggestion
        expect(tags).toStrictEqual(AttributeTagCollection.from(mockTags));
```

This way you don't require `toStringEqualExcluding`.

Milena-Czierlinski (approved):
LGTM 

@jkoenig134 Do you want to take another look?

jkoenig134:
@sebbi08 now the things in transport are also named `AttributeTag*` - but transport doesn't know of Attributes, for transport it's just tags.

If that's undone I will do another review.

packages/transport/src/core/TransportController.ts
```diff
@@ -30,6 +30,7 @@ export enum ControllerName {
     RelationshipTemplator = "RelationshipTemplator",
     Secret = "Secret",
     Sync = "Sync",
+    Tag = "Tag",
```
jkoenig134, line 33: Either this is not used anymore - or used wrong.

jkoenig134 (approved):
LGTM

PR #309, opened 2024-10-23
Multiple open Requests that lead to a Relationship can exist for the same Identity

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/packages/runtime/src/modules/RequestModule.ts
+++ b/packages/runtime/src/modules/RequestModule.ts
@@ -114,6 +114,25 @@ export class RequestModule extends RuntimeModule {
             return;
         }
 
+        const otherRequestsThatWouldLeadToARelationship = (
+            await services.consumptionServices.incomingRequests.getRequests({
+                query: { "source.type": "RelationshipTemplate", status: [LocalRequestStatus.Open, LocalRequestStatus.DecisionRequired, LocalRequestStatus.ManualDecisionRequired] }
+            })
+        ).value;
+        if (otherRequestsThatWouldLeadToARelationship.length !== 0) {
+            this.logger.info(
+                `There is already an open Request for a RelationshipTemplate that would lead to a Relationship with the creator of the RelationshipTemplate '${template.id}'. Skipping creation of a new Request.`
+            );
+            this.runtime.eventBus.publish(
+                new RelationshipTemplateProcessedEvent(event.eventTargetAddress, {
+                    template,
+                    result: RelationshipTemplateProcessedResult.NonCompletedRequestExists,
+                    requestId: otherRequestsThatWouldLeadToARelationship[0].id
+                })
+            );
+            return;
+        }
+
         const requestCreated = await this.createIncomingRequest(services, body.onNewRelationship, template.id);
 
         if (!requestCreated) {
--- a/packages/runtime/test/modules/RequestModule.test.ts
+++ b/packages/runtime/test/modules/RequestModule.test.ts
@@ -584,3 +584,61 @@ describe("Handling the rejection and the revocation of a Relationship by the Req
         return sRelationship;
     }
 });
+
+describe("Handle Multiple RelationshipTemplate loadings", () => {
+    const runtimeServiceProvider = new RuntimeServiceProvider();
+    let sRuntimeServices: TestRuntimeServices;
+    let rRuntimeServices: TestRuntimeServices;
+
+    beforeAll(async () => {
+        const runtimeServices = await runtimeServiceProvider.launch(2, { enableRequestModule: true, enableDeciderModule: true });
+
+        sRuntimeServices = runtimeServices[0];
+        rRuntimeServices = runtimeServices[1];
+    }, 30000);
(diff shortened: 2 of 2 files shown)


jkoenig134:
@stnmtz & @tnotheis maybe you also want to look over this.

packages/runtime/src/modules/RequestModule.ts
```diff
@@ -114,6 +114,29 @@ export class RequestModule extends RuntimeModule {
             return;
         }
 
+        const otherRequestsThatWouldLeadToARelationship = (
+            await services.consumptionServices.incomingRequests.getRequests({
+                query: {
+                    "source.type": "RelationshipTemplate",
+                    status: [LocalRequestStatus.Open, LocalRequestStatus.DecisionRequired, LocalRequestStatus.ManualDecisionRequired],
```
Magnus-Kuhn, line 121: There's the very similar case above where the newly loaded template is the template with existing requests. There the status `Decided` is included as well which I prefer.

packages/runtime/src/modules/RequestModule.ts
```diff
@@ -114,6 +114,29 @@ export class RequestModule extends RuntimeModule {
...
+                    peer: template.createdBy
+                }
+            })
+        ).value;
+        if (otherRequestsThatWouldLeadToARelationship.length !== 0) {
+            this.logger.info(
+                `There is already an open Request for a RelationshipTemplate that would lead to a Relationship with the creator of the RelationshipTemplate '${template.id}'. Skipping creation of a new Request.`
```
Magnus-Kuhn, line 128: I'd assume that people know that open requests lead to relationships and simplify to `There is already an open Request with the creator of the RelationshipTemplate '${template.id}'. Skipping creation of a new Request.` Also the `Skipping creation of a new Request.` can be added above in the "existing requests from same template" case

jkoenig134: Actually the first part can not be simplified as there are also requests where answering would turn into a mesage. Second is good.

packages/runtime/src/modules/RequestModule.ts
Magnus-Kuhn: Also in the case above `Expired` is not excluded, this should be fixed

packages/runtime/src/modules/RequestModule.ts
Magnus-Kuhn: If answering the request were to turn into a message, there would be a relationship which would be caught beforehand

jkoenig134: But this is not visible in this Info message. You have to Look in the Code to know that.

Magnus-Kuhn: Alright, it's fine, but how do I put it - seems consistent with the enmeshed documentation style of immediately giving a lot of information rather than giving just enough to proceed

britsta (changes requested):
I realized a while ago that some validation takes place in the modules, although it should actually take place at a deeper level. Otherwise, if no modules are used and the methods are called manually, strange things can happen. In the case of this validation, I would expect it to be included in the `received` method of the IncomingRequestsController.

jkoenig134:
@britsta all these validations (including the new one I've added) are checking things that are important for RequestModule logic.

To make this clearer using an example: the consumption library doesn’t care at all if there is a Relationship or another open Request because it doesn't decides how the response is sent out.

britsta:
> @britsta all these validations (including the new one I've added) are checking things that are important for RequestModule logic.
> 
> To make this clearer using an example: the consumption library doesn’t care at all if there is a Relationship or another open Request because it doesn't decides how the response is sent out.

No matter we have still validation in the modules or not, how can we ensure the correctness of the processes in case an integrator isn't using the modules? We still need the option to add the appropriate validation for this case.

britsta:
@jkoenig134 Mhh alright, let's discuss this again somewhere else. :D For the moment, the newly added validation can stay with the other validation in the Module. 👼

britsta:
> Thanks master 😅

@jkoenig134 It's been a long time since anyone has referred to me as master of disaster 👀

packages/runtime/src/modules/RequestModule.ts
jkoenig134: I will not add Decided because

1. if the decision is accepted there would be a Relationship
2. if the decision is rejected I should be able to receive the Request

packages/runtime/src/modules/RequestModule.ts
jkoenig134: What the hell am I reading here.

packages/runtime/src/modules/RequestModule.ts
jkoenig134: I added the filter for decided.

I'd like to end the decision about the above check as it doesn't correlates with my PR.

Magnus-Kuhn (approved):

PR #314, opened 2024-10-28
Key pinning

Magnus-Kuhn:
# Readiness checklist

-   [X] I labeled the PR.

### DIFF
--- a/packages/transport/src/core/Transport.ts
+++ b/packages/transport/src/core/Transport.ts
@@ -6,6 +6,7 @@ import { SodiumWrapper } from "@nmshd/crypto";
 import { AgentOptions } from "http";
 import { AgentOptions as HTTPSAgentOptions } from "https";
 import _ from "lodash";
+import { PeerCertificate, checkServerIdentity } from "tls";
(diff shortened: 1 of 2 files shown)


packages/transport/src/core/Transport.ts
```diff
@@ -6,6 +6,7 @@ import { SodiumWrapper } from "@nmshd/crypto";
 import { AgentOptions } from "http";
 import { AgentOptions as HTTPSAgentOptions } from "https";
 import _ from "lodash";
+import { PeerCertificate, checkServerIdentity } from "tls";
```
jkoenig134, line 9: I think this will kill App builds, I have a feeling this PR should be in the Connector.

Magnus-Kuhn: > I think this will kill App builds

How that?

jkoenig134: the tls package is only available in nodejs and not in browsers

Magnus-Kuhn:
Moving this to the Connector https://github.com/nmshd/connector/pull/295

PR #319, opened 2024-11-04
Uniqueness of `key` for RelationshipAttributes

britsta:
# Readiness checklist

-   [x] I labeled the PR.

# Description

In each Relationship, there should only be at most one RelationshipAttribute for each Relationship context for a given trio of `value.@type`, `owner` and `key`.

## Validation To Be Added

It is probably only necessary to ensure the uniqueness of the `key` in those cases in which new RelationshipAttributes (ThirdPartyRelationshipAttributes explicitly excluded) are created. This means, validation needs to be added for the following cases:

- ReadAttributeRequestItem with RelationshipAttributeQuery as `query`.
- ProposeAttributeRequestItem with RelationshipAttributeQuery as `query`.
- CreateAttributeRequestItem with RelationshipAttribute as `attribute` to be created.

To be sure, the validation is added on both sides when the Request is created and when the Request is received. Also note the following details:

- If there are several RequestItems within a Request that would lead to the creation of RelationshipAttributes with the same `key`, the creation of such a Request is prevented by early iteration through the RequestItems. Additionally, an error would be thrown on the side of the Recipient of the Request in the unlikely event that it has received such a Request.
- If the Recipient receives a Request that it cannot accept _regardless of its specified parameters_, an error is thrown. If parameters can be specified differently so that acceptance is possible, an ErrorValidationResult is returned instead.

## Validation Not To Be Added

When it comes to existing RelationshipAttributes that are shared and lead to the creation of ThirdPartyRelationshipAttributes, no further validation needs to be added. There are three reasons for this:

- The uniqueness of the `key` is already ensured when the initial RelationshipAttributes are created.
- The re-sharing of the same RelationshipAttribute that has already been shared is already prohibited. 
- It makes sense to query a ThirdPartyRelationshipAttribute with the same `key` again, as it can happen during the succession of RelationshipAttributes that there is another existing RelationshipAttribute with the same `key` that is to be read.

## Open Questions

- We have to decide how to handle it when the two Identities involved in the Relationship send each other a Request to create a RelationshipAttribute with the same `key` and both accept it before they have received the response from their counterpart by synchronization. This can possibly be solved by forcing an early synchronization.
- Requests should not become unacceptable just because it is not possible to accept a single RequestItem by creating another RelationshipAttribute with the same `key`. This can perhaps be solved by adding a new ResponseItem.

## Further Notes

- Note that it is a priori not possible to add validations in the `applyIncomingResponseItem` method of the various RequestItemProcessors that prevent the creation of a RelationshipAttribute with an already existing `key`, as this would lead to inconsistent data sets between the two Identities of the Relationship.
- Nested RequestItemGroups are currently not fully supported, which is why some code had to be added in a simplified version.


packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -245,6 +272,150 @@ export class IncomingRequestsController extends ConsumptionBaseController {
...
+        items: (RequestItem | RequestItemGroup)[],
+        params: (DecideRequestItemParametersJSON | DecideRequestItemGroupParametersJSON)[]
+    ) {
+        const fragmentsOfMustBeAcceptedItemsOfRequest: RelationshipAttributeFragment[] = [];
+        const fragmentsOfAcceptedItemsOfRequest: RelationshipAttributeFragment[] = [];
+
+        for (let i = 0; i < params.length; i++) {
+            const item = items[i];
+            const decideItemParams = params[i];
+
```
Magnus-Kuhn, line 285: You could retrieve the relevant existing attributes here so you don't have to do it in each request item processor

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -105,6 +129,95 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
         return result;
     }
 
+    private static validateKeyUniquenessOfRelationshipAttributesWithinRequest(items: (RequestItem | RequestItemGroup)[], recipient?: CoreAddress) {
+        const fragmentsOfMustBeAcceptedItemsOfRequest: RelationshipAttributeFragment[] = [];
+        for (const item of items) {
+            if (item instanceof RequestItemGroup) {
+                const fragmentsOfMustBeAcceptedItemsOfGroup = OutgoingRequestsController.extractRelationshipAttributeFragmentsFromMustBeAcceptedItemsOfGroup(item, recipient);
```
Magnus-Kuhn, line 136: What is the exact use for restricting this to the items that must be accepted? I think that items that can't be accepted shouldn't be created. Is it about creating e. g. two ReadAttributeRequestItems for the same key with different queries where only one can be accepted?

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -105,6 +129,95 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
...
+        }
+
+        return ValidationResult.success();
+    }
+
+    private static extractRelationshipAttributeFragmentsFromMustBeAcceptedItemsOfGroup(requestItemGroup: RequestItemGroup, recipient?: CoreAddress) {
+        const fragmentsOfMustBeAcceptedItemsOfGroup: RelationshipAttributeFragment[] = [];
+
+        for (const item of requestItemGroup.items) {
+            if (item instanceof RequestItemGroup) {
```
Magnus-Kuhn, line 159: `item` is a `RequestItemDerivation`, and `RequestItemGroup` is not given as a type there, so I don't think that repeated groupings are possible

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -65,9 +85,13 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
         }
 
         const innerResults = await this.canCreateItems(parsedParams.content, parsedParams.peer);
-
         const result = ValidationResult.fromItems(innerResults);
 
+        const keyUniquenessValidationResult = OutgoingRequestsController.validateKeyUniquenessOfRelationshipAttributesWithinRequest(parsedParams.content.items, parsedParams.peer);
```
Magnus-Kuhn, line 90: This can be moved into the `if (parsedParams.peer)`, then you don't need an optional recipient in all your new methods

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
```diff
@@ -45,23 +49,67 @@ export class CreateAttributeRequestItemProcessor extends GenericRequestItemProce
...
+                requestItem.attribute.key,
+                ownerIsEmptyString ? recipient : requestItem.attribute.owner,
+                requestItem.attribute.value.toJSON()["@type"],
+                recipient
+            );
+
+            if (relationshipAttributesWithSameKey.length !== 0) {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.invalidRequestItem(
+                        "The provided RelationshipAttribute could not be created because there is already a RelationshipAttribute with the same key in the context of this Relationship."
```
Milena-Czierlinski, line 79: > The provided RelationshipAttribute could not be created...

To me this sounds like otherwise the RelationshipAttribute would have been created right away, however, this will only happen once the recipient accepts it and the sender receives the Response. Therefore, I'd rather write something like "You cannot request the creation of the RelationshipAttribute...".

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
```diff
@@ -45,23 +49,67 @@ export class CreateAttributeRequestItemProcessor extends GenericRequestItemProce
...
+            const relationshipAttributesWithSameKey = await this.consumptionController.attributes.getRelationshipAttributesOfValueTypeToPeerWithGivenKeyAndOwner(
+                requestItem.attribute.key,
+                ownerIsEmptyString ? this.currentIdentityAddress : requestItem.attribute.owner,
+                requestItem.attribute.value.toJSON()["@type"],
+                requestInfo.peer
+            );
+
+            if (relationshipAttributesWithSameKey.length !== 0 && requestItem.mustBeAccepted) {
+                throw ConsumptionCoreErrors.requests.violatedKeyUniquenessOfRelationshipAttributes(
+                    "The provided RelationshipAttribute cannot be created because there is already a RelationshipAttribute with the same key in the context of this Relationship."
```
Milena-Czierlinski, line 101: key and value type and owner (if we stick with caring about the owner)

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
```diff
@@ -45,23 +49,67 @@ export class CreateAttributeRequestItemProcessor extends GenericRequestItemProce
...
+                );
+            } else if (relationshipAttributesWithSameKey.length !== 0) {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.invalidAcceptParameters(
+                        "This CreateAttributeRequestItem cannot be accepted as the provided RelationshipAttribute cannot be created because there is already a RelationshipAttribute with the same key in the context of this Relationship."
```
Milena-Czierlinski, line 106: key and value type and owner (if we stick with caring about the owner)

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
```diff
@@ -46,6 +46,23 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
...
+                requestItem.query.key,
+                recipient,
+                requestItem.query.attributeCreationHints.valueType,
+                recipient
+            );
+
+            if (relationshipAttributesWithSameKey.length !== 0) {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.invalidRequestItem(
+                        "The queried RelationshipAttribute could not be created because there is already a RelationshipAttribute with the same key in the context of this Relationship."
```
Milena-Czierlinski, line 60: key and value type and owner (if we stick with caring about the owner)

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
```diff
@@ -157,6 +174,27 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
...
+            const relationshipAttributesWithSameKey = await this.consumptionController.attributes.getRelationshipAttributesOfValueTypeToPeerWithGivenKeyAndOwner(
+                requestItem.query.key,
+                this.currentIdentityAddress,
+                requestItem.query.attributeCreationHints.valueType,
+                requestInfo.peer
+            );
+
+            if (relationshipAttributesWithSameKey.length !== 0 && requestItem.mustBeAccepted) {
+                throw ConsumptionCoreErrors.requests.violatedKeyUniquenessOfRelationshipAttributes(
+                    "The queried RelationshipAttribute cannot be created because there is already a RelationshipAttribute with the same key in the context of this Relationship."
```
Milena-Czierlinski, line 187: key and value type and owner (if we stick with caring about the owner)

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
```diff
@@ -157,6 +174,27 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
...
+                );
+            } else if (relationshipAttributesWithSameKey.length !== 0) {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.invalidAcceptParameters(
+                        "This ProposeAttributeRequestItem cannot be accepted as the queried RelationshipAttribute cannot be created because there is already a RelationshipAttribute with the same key in the context of this Relationship."
```
Milena-Czierlinski, line 192: key and value type and owner (if we stick with caring about the owner)

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -27,27 +27,49 @@ import validateQuery from "../utility/validateQuery";
...
+                    requestItem.query.key,
+                    ownerIsEmptyString ? recipient : requestItem.query.owner,
+                    requestItem.query.attributeCreationHints.valueType,
+                    recipient
+                );
+
+                if (relationshipAttributesWithSameKey.length !== 0) {
+                    return ValidationResult.error(
+                        ConsumptionCoreErrors.requests.invalidRequestItem(
+                            "The queried RelationshipAttribute could not be created because there is already a RelationshipAttribute with the same key in the context of this Relationship."
```
Milena-Czierlinski, line 68: key and value type and owner (if we stick with caring about the owner)

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -181,6 +203,29 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+            const relationshipAttributesWithSameKey = await this.consumptionController.attributes.getRelationshipAttributesOfValueTypeToPeerWithGivenKeyAndOwner(
+                requestItem.query.key,
+                ownerOfQueriedAttributeIsEmptyString ? this.currentIdentityAddress : requestItem.query.owner,
+                requestItem.query.attributeCreationHints.valueType,
+                requestInfo.peer
+            );
+
+            if (relationshipAttributesWithSameKey.length !== 0 && requestItem.mustBeAccepted) {
+                throw ConsumptionCoreErrors.requests.violatedKeyUniquenessOfRelationshipAttributes(
+                    "The queried RelationshipAttribute cannot be created because there is already a RelationshipAttribute with the same key in the context of this Relationship."
```
Milena-Czierlinski, line 218: key and value type and owner (if we stick with caring about the owner)

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -181,6 +203,29 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+                );
+            } else if (relationshipAttributesWithSameKey.length !== 0) {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.invalidAcceptParameters(
+                        "This ReadAttributeRequestItem cannot be accepted as the queried RelationshipAttribute cannot be created because there is already a RelationshipAttribute with the same key in the context of this Relationship."
```
Milena-Czierlinski, line 223: key and value type and owner (if we stick with caring about the owner)

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -105,6 +129,95 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
...
+            } else {
+                const fragmentOfMustBeAcceptedRequestItem = OutgoingRequestsController.extractRelationshipAttributeFragmentFromMustBeAcceptedRequestItem(item, recipient);
+                if (fragmentOfMustBeAcceptedRequestItem) fragmentsOfMustBeAcceptedItemsOfRequest.push(fragmentOfMustBeAcceptedRequestItem);
+            }
+        }
+
+        if (OutgoingRequestsController.containsDuplicateRelationshipAttributeFragments(fragmentsOfMustBeAcceptedItemsOfRequest)) {
+            return ValidationResult.error(
+                ConsumptionCoreErrors.requests.violatedKeyUniquenessOfRelationshipAttributes(
+                    "The Request cannot be created because its acceptance would lead to the creation of more than one RelationshipAttribute in the context of this Relationship with the same key."
```
Milena-Czierlinski, line 147: key and value type and owner (if we stick with caring about the owner)

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
Milena-Czierlinski, line 1: Like you already pointed out yourself, we have some redundancy here. Therefore, I'd suggest you extract this validation to the `utility` folder and access it from the Incoming- and OutgoingRequestController. Some points I commented on in the IncomingRequestsController also apply here.

Milena-Czierlinski (commented):
I haven't reviewed the test, so far.

A general question: Why should it be allowed to have multiple RelationshipAttributes with the same `key` in a Relationship given different owners?

> Note that it is a priori not possible to add validations in the applyIncomingResponseItem method of the various RequestItemProcessors that prevent the creation of a RelationshipAttribute with an already existing key, as this would lead to inconsistent data sets between the two Identities of the Relationship.

I get that. However, maybe we still need to think about how we could add some kind of validation on the side of the sender of the Request, in case the recipient did some misdeed in their Response, like we always do the other way around.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -245,6 +272,150 @@ export class IncomingRequestsController extends ConsumptionBaseController {
...
+        }
+
+        if (IncomingRequestsController.containsDuplicateRelationshipAttributeFragments(fragmentsOfMustBeAcceptedItemsOfRequest)) {
+            throw ConsumptionCoreErrors.requests.violatedKeyUniquenessOfRelationshipAttributes(
+                "The Request can never be accepted because it would lead to the creation of more than one RelationshipAttribute in the context of this Relationship with the same key."
+            );
+        } else if (IncomingRequestsController.containsDuplicateRelationshipAttributeFragments(fragmentsOfAcceptedItemsOfRequest)) {
+            return ValidationResult.error(
+                ConsumptionCoreErrors.requests.invalidAcceptParameters(
+                    "The Request cannot be accepted with this parameters because it would lead to the creation of more than one RelationshipAttribute in the context of this Relationship with the same key."
```
Magnus-Kuhn, line 319: It would be good if the problematic requestItems were named here (e. g. the way it happens in requestItemProcessors)

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
Magnus-Kuhn: On a second thought: With my suggestion the validationResult would not name the problematic requestItems, so I think it's fine to keep them in the itemProcessors. Unless the validationResult is changes to also name the individual problematic requestItems (see the comment for the error message in case of a failed validation)

britsta:
> A general question: Why should it be allowed to have multiple RelationshipAttributes with the same `key` in a Relationship given different owners?

This is nothing I have decided, I followed the requirement. However, I think it's a bit about the fact that the result of the `executeRelationshipAttributeQuery` should have zero elements (if none matching RelationshipAttribute exists) or only one element. This can be achieved even if key uniqueness is only required per `owner`, as you always have to specify an `owner` in a RelationshipAttributeQuery. But of course you can consider being even stricter with the key uniqueness.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -245,6 +272,150 @@ export class IncomingRequestsController extends ConsumptionBaseController {
...
+                owner: ownerIsEmptyString ? this.identity.address.toString() : requestItem.query.owner.toString(),
+                key: requestItem.query.key,
+                value: { "@type": requestItem.query.attributeCreationHints.valueType }
+            };
+        }
+
+        return;
+    }
+
+    private static containsDuplicateRelationshipAttributeFragments(fragments: RelationshipAttributeFragment[]) {
```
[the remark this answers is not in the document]
britsta: Do you have a suggestion that works without these RelationshipAttributeFragments or similar?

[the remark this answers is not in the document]
Milena-Czierlinski: I image it in a way that you have a list of attributes (all mustBeAccepted or accepted, see my comment above). Then you check whether there are attributes that have the same values for specified keys. This way, you wouldn't have to create a separate list of RelationshipAttributeFragments.

This is only a high-level thought of me, I haven't checked how this would be implemented in detail. If you see problems with the implementation, feel free to reach out.

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
```diff
@@ -45,23 +49,67 @@ export class CreateAttributeRequestItemProcessor extends GenericRequestItemProce
...
+                    "The owner of the provided RelationshipAttribute for the `attribute` property can only be the Sender's Address or an empty string. The latter will default to the Recipient's Address."
                 )
             );
         }
 
-        return ValidationResult.error(
-            ConsumptionCoreErrors.requests.invalidRequestItem(
-                "The owner of the provided RelationshipAttribute for the `attribute` property can only be the Sender's Address or an empty string. The latter will default to the Recipient's Address."
-            )
-        );
```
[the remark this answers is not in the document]
Milena-Czierlinski: I see. Now this comments on the current state, but your changes make it obvious in the diff: Why do we throw different, but almost identical, error messages depending on whether recipient is defined or not? In my opinion it would be much easier to not check for the recipient and just throw the first error message. If the recipient is not defined and, hence, their address not known, the reader will know what to do from reading the error message. This would save us complexity in the code. Also while we're at it, we could write "address" lowercase. Maybe these changes could be done in a small separate PR. What do you think?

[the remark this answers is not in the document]
britsta: I also think that this is currently cumbersome and we should definitely do it differently :D I think we should take care of this in another PR, probably in the long run we should prevent that the address of the Recipient can be set explicitly to ensure consistency with the behavior of other RequestItems. Also I'd like your suggestion to write address lowercase.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1288,4 +1288,25 @@ export class AttributesController extends ConsumptionBaseController {
 
         return ownSharedAttributeSuccessors;
     }
+
+    public async getRelationshipAttributesOfValueTypeToPeerWithGivenKeyAndOwner(key: string, owner: CoreAddress, valueType: string, peer: CoreAddress): Promise<LocalAttribute[]> {
+        const queryForRelationshipAttributesWithSameKey = {
```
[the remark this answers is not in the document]
jkoenig134: or just inline it.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -166,7 +184,16 @@ export class IncomingRequestsController extends ConsumptionBaseController {
...
+        const canDecideResult = await this.canDecide({ ...params, accept: true });
+
+        const request = await this.getOrThrow(params.requestId);
+
+        const keyUniquenessValidationResult = this.validateKeyUniquenessOfRelationshipAttributesWithinRequest(request.content.items, params.items);
+        if (keyUniquenessValidationResult.isError() && canDecideResult.isSuccess()) {
+            return keyUniquenessValidationResult;
+        }
+
+        return canDecideResult;
```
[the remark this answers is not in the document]
britsta: Early return is possible if `canDecideResult` is an error result and I have changed it that way. Returning `keyUniquenessValidationResult` instead of `canDecideResult` in case of both results are success results is not possible as they have different formats. The tests would fail otherwise.

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
britsta: You are right, I have restricted it to `mustBeAccepted` RequestItems due to some edge cases. :D

For example, the RequestItem could become acceptable again if the existing RelationshipAttribute of this `key` has been deleted in the meantime. Even if it has not been deleted, the Request is not immediately unusable because the RequestItem can still be rejected as the RequestItem is not `mustBeAccepted`.

I think it's debatable whether this makes sense or not. In general, I have tried to validate as little restrictively as possible.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
britsta: How exactly could this be achieved? Since the problematic RequestItems cannot be identified by an ID, I find it difficult not to remain vague.

Magnus-Kuhn: Just the key will have to do then

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
Magnus-Kuhn: I understand, also agree that it's debatable. I don't know what kind of requests are commonly sent, so I don't really have an opinion there

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1288,4 +1288,25 @@ export class AttributesController extends ConsumptionBaseController {
...
+            "content.@type": "RelationshipAttribute",
+            "content.owner": owner.toString(),
+            "content.key": key,
+            "content.value.@type": valueType,
+            "shareInfo.peer": peer.toString(),
+            "shareInfo.thirdPartyAddress": { $exists: false },
+            "deletionInfo.deletionStatus": {
+                $nin: [
+                    LocalAttributeDeletionStatus.ToBeDeleted,
```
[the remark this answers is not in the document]
britsta: I think it is debatable whether to allow a RelationshipAttribute that is in deletion to exist with the same `key`. As I can imagine cases in which it would be restrictive to prohibit this, I have allowed it for the time being. For example, if a RelationshipAttribute is being deleted because it is about to expire, but a new RelationshipAttribute with the same `key` is to be created, which is then valid for a longer period of time. It would feel unnatural to then have to wait for the final deletion of the former RelationshipAttribute. However, as I said, I think it is debatable. :D

I would say that `"DeletedByPeer"` RelationshipAttributes can be described as relics that embody history, but perhaps also still have a use. I don't know to what extent such RelationshipAttributes are ignored in the App, but that we often have to consider such RelationshipAttributes as edge cases in the Runtime, because it is no longer allowed to do so many things with them.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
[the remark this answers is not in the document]
britsta: The problem why I came up with the RelationshipAttributeFragments was that only a RelationshipAttributeQuery and no RelationshipAttribute needs to be provided for the ReadAttributeRequestItem. So I can't just extract RelationshipAttributes (as I could do it with the CreateAttributeRequestItem or ProposeAttributeRequestItem) and then look for duplicates. Also, I need to handle the case that the `owner` of the RelationshipAttribute could be an empty string or the explicit `address` of the recipient. 

I'm thinking about how I could implement the whole logic, including recursions, without using such lists of RelationshipAttribute(Fragment)s and might discuss this with you again.

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
britsta: It is correct that these lines relating to nested RequestItemGroups cannot currently be reached.  However, because we already have logic regarding nested RequestItemGroups in other places, I have also inserted it consistently (which is debatable). I think it would be a separate issue to talk about nested RequestItemGroups and if we should delete such logic. :D

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
britsta: I think that this is unfortunately not possible, because even with an unknown `parsedParams.peer`, incorrect Requests can be created, which would lead to the creation of duplicate RelationshipAttributes with the same `key` regardless of the future recipient. This is because empty strings can also be specified as placeholders for potential recipients as `owner` of RelationshipAttributes.

Or am I missing something?

packages/consumption/src/modules/attributes/AttributesController.ts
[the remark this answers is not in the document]
Magnus-Kuhn: Would it in your example be more appropriate to have the new Attribute succeed the toBeDeleted one rather than them coexisting? I also just saw that we have the option to answer the request with an attribute succession only for the Read- and Propose-, but not the CreateAttributeRequests. Is there a reason for that? I think this could allow requests with the same key if it can be accepted as an attribute succession.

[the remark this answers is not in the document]
britsta: Probably, even if the succession can currently only be performed by the `owner` of the RelationshipAttribute. Requesting the succession of an existing RelationshipAttribute from its `owner` is a feature that we currently don't have.

Answering the Request with an Attribute Succession is only possible with IdentityAttributes and ThirdPartyRelationshipAttributes. For RelationshipAttributes, with all three RequestItems Read/Create/Propose, always a new RelationshipAttribute is created that cannot be part of an already existing succession chain. That was therefore not an option.

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
Magnus-Kuhn: No, I missed that you are checking the request items in isolation and don't need to know the relationship for that

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -245,6 +272,150 @@ export class IncomingRequestsController extends ConsumptionBaseController {
...
+
+                const fragmentOfAcceptedRequestItem = this.extractRelationshipAttributeFragmentFromAcceptedRequestItem(item, decideItemParams as DecideRequestItemParametersJSON);
+                if (fragmentOfAcceptedRequestItem) {
+                    fragmentsOfAcceptedItemsOfRequest.push(fragmentOfAcceptedRequestItem);
+                }
+            }
+        }
+
+        if (IncomingRequestsController.containsDuplicateRelationshipAttributeFragments(fragmentsOfMustBeAcceptedItemsOfRequest)) {
+            throw ConsumptionCoreErrors.requests.violatedKeyUniquenessOfRelationshipAttributes(
```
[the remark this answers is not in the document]
britsta: Do you mean that the entire Request should be automatically rejected if the Request is faulty in the sense that it contains several `mustBeAccepted` RequestItems, which would lead to the creation of RelationshipAttributes with the same key, owner and value type? Perhaps this discussion goes beyond the scope of the PR for the moment.

As you have already worked a lot with automatically processed Requests @Milena-Czierlinski, what do you think about it in principle?

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -245,6 +272,150 @@ export class IncomingRequestsController extends ConsumptionBaseController {
         }
     }
 
+    private validateKeyUniquenessOfRelationshipAttributesWithinRequest(
+        items: (RequestItem | RequestItemGroup)[],
+        params: (DecideRequestItemParametersJSON | DecideRequestItemGroupParametersJSON)[]
+    ) {
```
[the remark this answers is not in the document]
britsta: I can add the return types, but I've noticed that we don't specify return types for many private methods. Do you know why that is?

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -27,27 +27,49 @@ import validateQuery from "../utility/validateQuery";
...
+                    requestItem.query.key,
+                    ownerIsEmptyString ? recipient : requestItem.query.owner,
+                    requestItem.query.attributeCreationHints.valueType,
+                    recipient
+                );
+
+                if (relationshipAttributesWithSameKey.length !== 0) {
+                    return ValidationResult.error(
+                        ConsumptionCoreErrors.requests.invalidRequestItem(
+                            `The queried RelationshipAttribute could not be created because there is already a RelationshipAttribute in the context of this Relationship with the same key '${requestItem.query.key}', owner and value type.`
```
Milena-Czierlinski, line 68: See above.

packages/consumption/src/modules/requests/outgoing/OutgoingRequestsController.ts
```diff
@@ -77,9 +78,13 @@ export class OutgoingRequestsController extends ConsumptionBaseController {
         }
 
         const innerResults = await this.canCreateItems(parsedParams.content, parsedParams.peer);
-
         const result = ValidationResult.fromItems(innerResults);
 
+        const keyUniquenessValidationResult = validateKeyUniquenessOfRelationshipAttributesWithinOutgoingRequest(parsedParams.content.items, parsedParams.peer);
+        if (keyUniquenessValidationResult.isError() && result.isSuccess()) {
+            return keyUniquenessValidationResult;
+        }
```
Milena-Czierlinski, line 86:
```suggestion
        if (result.isError()) return result;
        
        const keyUniquenessValidationResult = validateKeyUniquenessOfRelationshipAttributesWithinOutgoingRequest(parsedParams.content.items, parsedParams.peer);
        if (keyUniquenessValidationResult.isError()) return keyUniquenessValidationResult;
```

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
```diff
@@ -0,0 +1,206 @@
...
+    key: string;
+    value: { "@type": string };
+}
+
+type ContainsDuplicateRelationshipAttributeFragmentsResponse =
+    | { containsDuplicates: false }
+    | {
+          containsDuplicates: true;
+          duplicateFragment: RelationshipAttributeFragment;
+      };
```
Milena-Czierlinski, line 27:
```suggestion
type ContainsDuplicateRelationshipAttributeFragmentsResponse =
    { containsDuplicates: false }
    | {
          containsDuplicates: true;
          duplicateFragment: RelationshipAttributeFragment;
      };
```

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
```diff
@@ -0,0 +1,206 @@
...
+    const fragmentsOfMustBeAcceptedItemsOfRequest: RelationshipAttributeFragment[] = [];
+    for (const item of items) {
+        if (item instanceof RequestItemGroup) {
+            const fragmentsOfMustBeAcceptedItemsOfGroup = extractRelationshipAttributeFragmentsFromMustBeAcceptedItemsOfGroup(item, recipient);
+            if (fragmentsOfMustBeAcceptedItemsOfGroup) fragmentsOfMustBeAcceptedItemsOfRequest.push(...fragmentsOfMustBeAcceptedItemsOfGroup);
+        } else {
+            const fragmentOfMustBeAcceptedRequestItem = extractRelationshipAttributeFragmentFromMustBeAcceptedRequestItem(item, recipient);
+            if (fragmentOfMustBeAcceptedRequestItem) fragmentsOfMustBeAcceptedItemsOfRequest.push(fragmentOfMustBeAcceptedRequestItem);
+        }
+    }
```
Milena-Czierlinski, line 39: I think you can simply call `extractRelationshipAttributeFragmentsFromMustBeAcceptedItemsOfGroup` here, since you already implemented the recursion there. Probably some names need to be adjusted then.

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
```diff
@@ -0,0 +1,206 @@
...
+
+function extractRelationshipAttributeFragmentFromRequestItem(requestItem: RequestItem, recipient?: CoreAddress): RelationshipAttributeFragment | void {
+    if (requestItem instanceof CreateAttributeRequestItem && requestItem.attribute instanceof RelationshipAttribute) {
+        const ownerIsEmptyString = requestItem.attribute.owner.toString() === "";
+        return {
+            owner: ownerIsEmptyString && recipient ? recipient.toString() : requestItem.attribute.owner.toString(),
+            key: requestItem.attribute.key,
+            value: { "@type": requestItem.attribute.value.toJSON()["@type"] }
+        };
+    } else if ((requestItem instanceof ReadAttributeRequestItem || requestItem instanceof ProposeAttributeRequestItem) && requestItem.query instanceof RelationshipAttributeQuery) {
```
Milena-Czierlinski, line 180: I think the `else` is not required, since there is a `return` in the former `if` block.

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
```diff
@@ -0,0 +1,206 @@
...
+
+export function validateKeyUniquenessOfRelationshipAttributesWithinIncomingRequest(
+    items: (RequestItem | RequestItemGroup)[],
+    params: (DecideRequestItemParametersJSON | DecideRequestItemGroupParametersJSON)[],
+    recipient: CoreAddress
+): ValidationResult {
+    const fragmentsOfMustBeAcceptedItemsOfRequest: RelationshipAttributeFragment[] = [];
+
+    for (let i = 0; i < params.length; i++) {
+        const item = items[i];
```
Milena-Czierlinski, line 61:
```suggestion
    for (const item of items) {
```

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
```diff
@@ -0,0 +1,206 @@
...
+            const fragmentsOfAcceptedItemsOfGroup = extractRelationshipAttributeFragmentsFromAcceptedItemsOfGroup(item, decideItemParams as DecideRequestItemGroupParametersJSON);
+            if (fragmentsOfAcceptedItemsOfGroup) {
+                fragmentsOfAcceptedItemsOfRequest.push(...fragmentsOfAcceptedItemsOfGroup);
+            }
+        } else {
+            const fragmentOfAcceptedRequestItem = extractRelationshipAttributeFragmentFromAcceptedRequestItem(item, decideItemParams as DecideRequestItemParametersJSON);
+            if (fragmentOfAcceptedRequestItem) {
+                fragmentsOfAcceptedItemsOfRequest.push(fragmentOfAcceptedRequestItem);
+            }
+        }
```
Milena-Czierlinski, line 99: In case you can simplify this for the mustBeAcceptedItems, please check if this is also possible here.

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    return;
+}
+
+function extractRelationshipAttributeFragmentFromRequestItem(requestItem: RequestItem, recipient?: CoreAddress): RelationshipAttributeFragment | void {
+    if (requestItem instanceof CreateAttributeRequestItem && requestItem.attribute instanceof RelationshipAttribute) {
+        const ownerIsEmptyString = requestItem.attribute.owner.toString() === "";
+        return {
+            owner: ownerIsEmptyString && recipient ? recipient.toString() : requestItem.attribute.owner.toString(),
```
Magnus-Kuhn, line 176: I think there's a scenario where the recipient remains unnamed and is named as attributeOwner in one request item but not named in the other request item. In both cases the recipient would be the owner but the strings would be different. What about having the owner be a boolean - request sender vs. recipient?

packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -27,27 +27,49 @@ import validateQuery from "../utility/validateQuery";
...
+
+            if (!(senderIsAttributeOwner || ownerIsEmptyString)) {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.invalidRequestItem(
+                        "The owner of the given `query` can only be an empty string or yourself. This is because you can only request RelationshipAttributes using a ReadAttributeRequestitem with a RelationshipAttributeQuery where the Recipient of the Request or yourself is the owner. And in order to avoid mistakes, the Recipient automatically will become the owner of the RelationshipAttribute later on if the owner of the `query` is an empty string."
+                    )
+                );
+            }
+
+            if (typeof recipient !== "undefined") {
```
Magnus-Kuhn, line 57: Could you decide for this and the `ProposeAttributeProcessor` if it's in `validateQuery` or not?

Also similarly, if the query owner is not the request sender, there's enough information to run the check

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
```diff
@@ -0,0 +1,206 @@
...
+            ConsumptionCoreErrors.requests.invalidAcceptParameters(
+                `The Request cannot be accepted with this parameters because it would lead to the creation of more than one RelationshipAttribute in the context of this Relationship with the same key '${containsAcceptedDuplicatesResult.duplicateFragment.key}', owner and value type.`
+            )
+        );
+    }
+
+    return ValidationResult.success();
+}
+
+function extractRelationshipAttributeFragmentsFromMustBeAcceptedItemsOfGroup(requestItemGroup: RequestItemGroup, recipient?: CoreAddress): RelationshipAttributeFragment[] | void {
```
Magnus-Kuhn, line 114: `void` is a bit more special, there also other instances here

```suggestion
function extractRelationshipAttributeFragmentsFromMustBeAcceptedItemsOfGroup(requestItemGroup: RequestItemGroup, recipient?: CoreAddress): RelationshipAttributeFragment[] | undefined {
```

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
Magnus-Kuhn: Also you are not using forEach in instances where it could be used

Magnus-Kuhn (commented):
I also still need to review the tests

Magnus-Kuhn (commented):
The other RequestItemProcessor tests should move to the standard of the CreateAttributeProcessor test. Also in the controller tests there coud be more options tested, although I can't tell which extent is sensible - I'd definitely want some positive tests though

Magnus-Kuhn:
@britsta Could you add the reason for why a failed validation is sometimes an error and sometimes just a negative ValidationResult to the description or somewhere? I can't find it 😃

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
```diff
@@ -46,6 +46,23 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
...
+                requestItem.query.key,
+                recipient,
+                requestItem.query.attributeCreationHints.valueType,
+                recipient
+            );
+
+            if (relationshipAttributesWithSameKey.length !== 0) {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.invalidRequestItem(
+                        `The queried RelationshipAttribute could not be created because there is already a RelationshipAttribute in the context of this Relationship with the same key '${requestItem.query.key}', owner and value type.`
```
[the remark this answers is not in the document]
britsta: I don't want to use a past tense here, but a subjunctive form.

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
```diff
@@ -30,30 +34,66 @@ export class CreateAttributeRequestItemProcessor extends GenericRequestItemProce
...
+                requestItem.attribute.value.toJSON()["@type"],
+                requestInfo.peer
             );
-        }
 
-        if (recipientIsAttributeOwner || senderIsAttributeOwner || ownerIsEmptyString) {
-            return ValidationResult.success();
+            if (relationshipAttributesWithSameKey.length !== 0 && requestItem.mustBeAccepted) {
+                throw ConsumptionCoreErrors.requests.violatedKeyUniquenessOfRelationshipAttributes(
+                    `The provided RelationshipAttribute cannot be created because there is already a RelationshipAttribute in the context of this Relationship with the same key '${requestItem.attribute.key}', owner and value type.`
```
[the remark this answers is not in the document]
britsta: It depends. Theoretically, it would also be possible for the other RelationshipAttribute with this `key` to be deleted. Then the Request would be acceptable again. The concept for unacceptable Requests would be used if the Request will **never** be acceptable (for example, if "contradictory" RequestItems are contained within a Request), regardless of what changes in the future.

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
britsta: I can't remove it, when I save, it is automatically added again. :D

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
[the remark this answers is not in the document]
britsta: That is not definitely clear. Theoretically, it would also be possible for the other RelationshipAttribute with this key to be deleted. Then the Request would be acceptable again. The concept for unacceptable Requests would be used if the Request will never be acceptable (for example, if "contradictory" RequestItems are contained within a Request), regardless of what changes in the future. In only such cases, rejecting the Request is necessary.

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
```diff
@@ -0,0 +1,206 @@
...
+    requestItemGroup: RequestItemGroup,
+    decideGroupParams: DecideRequestItemGroupParametersJSON
+): RelationshipAttributeFragment[] | void {
+    const fragmentsOfAcceptedItemsOfGroup: RelationshipAttributeFragment[] = [];
+
+    for (let i = 0; i < decideGroupParams.items.length; i++) {
+        const decideItemParams = decideGroupParams.items[i];
+        const item = requestItemGroup.items[i];
+
+        if (item instanceof RequestItem) {
```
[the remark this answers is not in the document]
Magnus-Kuhn: I wasn't aware that the `DecideParameters` didn't support the nesting. Then I see no need to change

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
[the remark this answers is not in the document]
Magnus-Kuhn: Got that, but do you want the error to suggest what fixes the problem (delete that attribute)? The runtime is inconsistent as to whether errors contain suggestions on how to fix the error, so I don't know

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
britsta: > I think you can simply call `extractRelationshipAttributeFragmentsFromMustBeAcceptedItemsOfGroup` here, since you already implemented the recursion there. Probably some names need to be adjusted then.

At first glance, this approach does not work with meaningful names or with the appropriate typing of the input parameters, but I'll see if I can get rid of the redundant recursion.

> Also you are not using forEach in instances where it could be used

We use `items.forEach((item) => { ... } )` compared to `for (const item of items) { ... }` less often in our codebase, which is why I would stick with the second variant when refactoring.

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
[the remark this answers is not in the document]
britsta: Whether the error message should provide problem fixes is a good question. At first I just want an error to be thrown at all to prevent the creation of duplicate RelationshipAttributes. :D Since a problem fix would be relatively complicated/extensive to describe here, I would perhaps consider adding a problem fix when generally refactoring the error messages with regard to problem fixes.

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
```diff
@@ -30,30 +34,66 @@ export class CreateAttributeRequestItemProcessor extends GenericRequestItemProce
...
+
+        if (!(recipientIsAttributeOwner || senderIsAttributeOwner || ownerIsEmptyString)) {
+            return ValidationResult.error(
+                ConsumptionCoreErrors.requests.invalidRequestItem(
+                    "The owner of the provided RelationshipAttribute for the `attribute` property can only be the address of the sender, the address of the recipient or an empty string. The latter will default to the address of the recipient."
+                )
+            );
+        }
+
+        if (typeof recipient !== "undefined") {
```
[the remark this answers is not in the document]
britsta: The `getRelationshipAttributesOfValueTypeToPeerWithGivenKeyAndOwner` method needs the `recipient` for its required `peer` input parameter. This method needs the `peer` because it checks the existing RelationshipAttributes. Other methods I have implemented do not need to know the `recipient` because they only ensure key uniqueness within the Request and do not take existing RelationshipAttributes into account.

If I have not understood you correctly, please let me know.

britsta:
> The other RequestItemProcessor tests should move to the standard of the CreateAttributeProcessor test. Also in the controller tests there coud be more options tested, although I can't tell which extent is sensible - I'd definitely want some positive tests though

I wanted to avoid too much redundancy, but will add a few more tests. :)

britsta:
> @britsta Could you add the reason for why a failed validation is sometimes an error and sometimes just a negative ValidationResult to the description or somewhere? I can't find it 😃

Hahah sure, I wanted to at least subtly differentiate whether a Request is messed up and therefore cannot be accepted regardless of the user input (this is where the concept of unacceptable Requests comes into play), or whether the user has provided inappropriate input and can accept (or create) the Request by correcting their input. 

With messed up Requests I throw an error, with inappropriate user input I return a negative ValidationResult.

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
britsta: Can you explain why this scenario should not be covered? Either the recipient is clearly named as the owner, or the empty string is used as a placeholder. If this is the case, the address of the recipient is used instead, so that you still have comparability.

Magnus-Kuhn: Ah, I wasn't aware of how the validation for the attributeOwner worked - I thought you could put the address of the recipient as the owner without giving that address as recipient. So it works the way it is.

In my scenario I meant that the recipient is undefined, this might have been unclear.

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
britsta: > At first glance, this approach does not work with meaningful names or with the appropriate typing of the input parameters, but I'll see if I can get rid of the redundant recursion.

Again, the not yet well thought-out features of the nested RequestItemGroups play a role, which is why the type of the `items` property of the Request and the `items` property of the RequestItemGroup do not match, which is why I could not implement your suggestion. Again, the not yet well thought-out features of the nested RequestItemGroups play a role, which is why the type of the `items` property of the request and the `items` property of the RequestItemGroup do not match, which is why I could not implement your suggestion.

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
```diff
@@ -84,12 +84,14 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
...
-                )
-            );
+        if (requestItem.query instanceof RelationshipAttributeQuery) {
+            if (requestItem.query.owner.toString() !== "") {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.invalidRequestItem(
+                        "The owner of the given `query` can only be an empty string. This is because you can only propose Attributes where the Recipient of the Request is the owner anyway. And in order to avoid mistakes, the owner will be automatically filled for you."
+                    )
+                );
+            }
```
Milena-Czierlinski, line 94: Why did you change this?

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
```diff
@@ -30,30 +34,66 @@ export class CreateAttributeRequestItemProcessor extends GenericRequestItemProce
...
+                throw ConsumptionCoreErrors.requests.violatedKeyUniquenessOfRelationshipAttributes(
+                    `The provided RelationshipAttribute cannot be created because there is already a RelationshipAttribute in the context of this Relationship with the same key '${requestItem.attribute.key}', owner and value type.`
+                );
+            } else if (relationshipAttributesWithSameKey.length !== 0) {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.invalidAcceptParameters(
+                        `This CreateAttributeRequestItem cannot be accepted as the provided RelationshipAttribute cannot be created because there is already a RelationshipAttribute in the context of this Relationship with the same key '${requestItem.attribute.key}', owner and value type.`
+                    )
+                );
+            }
```
Milena-Czierlinski, line 93:
```suggestion
            if (relationshipAttributesWithSameKey.length !== 0) {
                if (requestItem.mustBeAccepted) {
                    throw ConsumptionCoreErrors.requests.violatedKeyUniquenessOfRelationshipAttributes(
                        `The provided RelationshipAttribute cannot be created because there is already a RelationshipAttribute in the context of this Relationship with the same key '${requestItem.attribute.key}', owner and value type.`
                    );
                }

                return ValidationResult.error(
                    ConsumptionCoreErrors.requests.invalidAcceptParameters(
                        `This CreateAttributeRequestItem cannot be accepted as the provided RelationshipAttribute cannot be created because there is already a RelationshipAttribute in the context of this Relationship with the same key '${requestItem.attribute.key}', owner and value type.`
                    )
                );
            }
```

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
```diff
@@ -157,6 +176,27 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
...
+                throw ConsumptionCoreErrors.requests.violatedKeyUniquenessOfRelationshipAttributes(
+                    `The queried RelationshipAttribute cannot be created because there is already a RelationshipAttribute in the context of this Relationship with the same key '${requestItem.query.key}', owner and value type.`
+                );
+            } else if (relationshipAttributesWithSameKey.length !== 0) {
+                return ValidationResult.error(
+                    ConsumptionCoreErrors.requests.invalidAcceptParameters(
+                        `This ProposeAttributeRequestItem cannot be accepted as the queried RelationshipAttribute cannot be created because there is already a RelationshipAttribute in the context of this Relationship with the same key '${requestItem.query.key}', owner and value type.`
+                    )
+                );
+            }
```
Milena-Czierlinski, line 197: See other comment.

Milena-Czierlinski (commented):
I'll continue tomorrow. :)

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
britsta: That was not intentional. I had done a bit of refactoring based on another review comment, which I then undid. I must have forgotten to undo the change you are referring to. :D

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
britsta: I also like it much better this way 😄

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
Milena-Czierlinski: I'll take another look and get back to you. :)

packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
```diff
@@ -224,6 +230,56 @@ describe("OutgoingRequestsController", function () {
...
+        test("returns a validation result that contains an error for requests that would lead to the creation of more than one RelationshipAttribute with the same key", async function () {
+            const validationResult = await When.iCallCanCreateForAnOutgoingRequest({
+                content: {
+                    items: [
+                        CreateAttributeRequestItem.from({
+                            mustBeAccepted: true,
+                            attribute: RelationshipAttribute.from({
+                                "@type": "RelationshipAttribute",
+                                owner: "did:e:a-domain:dids:anidentity",
+                                key: "uniqueKey",
```
Milena-Czierlinski, line 243: Would `"duplicateKey"` be a better naming? Same in other places.

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
```diff
@@ -0,0 +1,162 @@
...
+        if (item instanceof RequestItemGroup) {
+            const fragmentsOfAcceptedItemsOfGroup = extractRelationshipAttributeFragmentsFromAcceptedItems(
+                item.items,
+                (decideItemParams as DecideRequestItemGroupParametersJSON).items
+            );
+            if (fragmentsOfAcceptedItemsOfGroup) {
+                fragmentsOfAcceptedItemsOfRequest.push(...fragmentsOfAcceptedItemsOfGroup);
+            }
+        } else {
+            if (!(decideItemParams as DecideRequestItemParametersJSON).accept) return;
```
britsta, line 112: If the condition is true, a return would be made and the `for` loop would be ended so that the remaining `items` are no longer processed. The condition should therefore be added to the one below.

packages/consumption/src/modules/requests/utility/validateRelationshipAttributesWithinRequest.ts
```diff
@@ -0,0 +1,162 @@
...
+    recipient?: CoreAddress
+): RelationshipAttributeFragment[] | undefined {
+    const fragmentsOfMustBeAcceptedItemsOfGroup: RelationshipAttributeFragment[] = [];
+
+    for (const item of items) {
+        if (item instanceof RequestItemGroup) {
+            const fragments = extractRelationshipAttributeFragmentsFromMustBeAcceptedItems(item.items, recipient);
+            if (fragments) fragmentsOfMustBeAcceptedItemsOfGroup.push(...fragments);
+        } else {
+            if (!item.mustBeAccepted) return;
```
britsta, line 83: If the condition is true, a return would be made and the `for` loop would be ended so that the remaining `items` are no longer processed. The condition should therefore be added to the one below.

packages/consumption/test/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.test.ts
```diff
@@ -535,6 +624,75 @@ describe("ProposeAttributeRequestItemProcessor", function () {
             });
         });
 
+        test("throws an error when another RelationshipAttribute with same key was queried", async function () {
```
Milena-Czierlinski, line 627: You could add a test trying to accept a RelationshipAttribute with the same key without mustBeAccepted. Same for ReadAttributeRequestItemProcessor.

packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
britsta: I think `"uniqueKey"` in connection with the name `error.consumption.requests.violatedKeyUniquenessOfRelationshipAttributes` of the error code is better. 😄 But I can also change it if it's important to you. 🎁

Milena-Czierlinski: Na, it's not too important to me. :)

Milena-Czierlinski (dismissed):

PR #336, opened 2024-11-19
Add UseCase for fetching the list of public relationshoip template references from the backbone

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.


Milena-Czierlinski:
@sebbi08 Is this ready for review or not? It's still a draft and has the WIP label.

packages/runtime/src/useCases/transport/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferenceMapper.ts
```diff
@@ -0,0 +1,8 @@
+import { PublicRelationshipTemplateReference } from "@nmshd/transport";
+import { PublicRelationshipTemplateReferenceDTO } from "../../../types/transport/PublicRelationshipTemplateReferenceDTO";
+
+export class PublicRelationshipTemplateReferenceMapper {
+    public static toPublicRelationshipTemplateReferencDTO(tagList: PublicRelationshipTemplateReference): PublicRelationshipTemplateReferenceDTO {
```
britsta, line 5: Typo, the last "e" of reference is missing: Write "toPublicRelationshipTemplateReferenceDTO" instead of "toPublicRelationshipTemplateReferencDTO".

britsta (commented):
Typo in the PR title (relationshoip template). Also, write RelationshipTemplate CamelCase and Backbone capitalized.

packages/runtime/src/Runtime.ts
```diff
@@ -258,6 +259,10 @@ export abstract class Runtime<TConfig extends RuntimeConfig = RuntimeConfig> {
             .factory(() => this.getAccountController().tokens)
             .scope(Scope.Request);
 
+        Container.bind(PublicRelationshipTemplateReferencesController)
+            .factory(() => this.getAccountController().publicRelationshipTemplateReferences)
+            .scope(Scope.Singleton);
```
britsta, line 264: Just out of interest: Many Controller have `.scope(Scope.Request)`, some have `.scope(Scope.Singleton)`. What's the difference here?

packages/runtime/src/useCases/transport/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferenceMapper.ts
britsta: The input parameter `tagList` doesn't seem to be named appropriately in this context.

packages/runtime/test/transport/publicRelationshipTemplateReferences.test.ts
```diff
@@ -0,0 +1,17 @@
+import { RuntimeServiceProvider, TestRuntimeServices } from "../lib";
+
+const serviceProvider = new RuntimeServiceProvider();
+let runtimeServices1: TestRuntimeServices;
+
+beforeAll(async () => {
+    const runtimeServices = await serviceProvider.launch(2);
```
britsta, line 7: It is enough to launch one instead of two runtimeServices instances. :D

I'd prefer to write `runtimeServices = (await serviceProvider.launch(1))[0];` if we only use one runtimeServices instance here and to omit the next line (use variable name `runtimeServices` instead of `runtimeServices1` and declare it above).

packages/transport/src/core/TransportController.ts
```diff
@@ -29,7 +29,8 @@ export enum ControllerName {
     RelationshipTemplator = "RelationshipTemplator",
     Secret = "Secret",
     Sync = "Sync",
-    Token = "Token"
+    Token = "Token",
+    PublicRelationshipTemplateController = "PublicRelationshipTemplateController"
```
britsta, line 33: This list is in alphabetical order, please insert correctly after `Message` and before `Relationship`.

packages/transport/src/modules/accounts/AccountController.ts
```diff
@@ -63,6 +64,7 @@ export class AccountController {
     public relationshipTemplates: RelationshipTemplateController;
     private synchronization: SyncController;
     public tokens: TokenController;
+    public publicRelationshipTemplateReferences: PublicRelationshipTemplateReferencesController;
```
britsta, line 67: This list is mainly in alphabetical order, please insert correctly after `messages` and before `relationships`.

packages/transport/src/modules/publicRelationshipTemplatesReference/PublicRelationshipTemplateController.ts
britsta, line 1: The folder needs to be renamed from `publicRelationshipTemplatesReference` to `publicRelationshipTemplateReferences`.

packages/transport/src/modules/index.ts
```diff
@@ -65,6 +65,9 @@ export * from "./messages/transmission/MessageEnvelope";
 export * from "./messages/transmission/MessageEnvelopeRecipient";
 export * from "./messages/transmission/MessageSignature";
 export * from "./messages/transmission/MessageSigned";
+export * from "./publicRelationshipTemplatesReference/backbone/PublicRelationshipTemplateReferenceClient";
```
britsta, line 68: A plural "s" is in the wrong position (we need to write "References" instead of "Templates"), please change the according folder name.

packages/transport/src/modules/publicRelationshipTemplatesReference/PublicRelationshipTemplateController.ts
britsta, line 1: The file needs to be renamed to `PublicRelationshipTemplateReferencesController.ts` (reference is missing in file name).

packages/transport/src/modules/publicRelationshipTemplatesReference/PublicRelationshipTemplateController.ts
```diff
@@ -0,0 +1,37 @@
+import { RequestError } from "../../core/backbone/RequestError";
+import { ControllerName, TransportController } from "../../core/TransportController";
+import { AccountController } from "../accounts/AccountController";
+import { PublicRelationshipTemplateReferenceClient } from "./backbone/PublicRelationshipTemplateReferenceClient";
+import { PublicRelationshipTemplateReference } from "./data/PublicRelationshipTemplateReference";
+
+export class PublicRelationshipTemplateReferencesController extends TransportController {
+    public constructor(parent: AccountController) {
+        super(ControllerName.PublicRelationshipTemplateController, parent);
```
britsta, line 9: Do you want to name the Controller `PublicRelationshipTemplateReferencesController` or `PublicRelationshipTemplateController`? There is an inconsistency currently. Since `PublicRelationshipTemplateReferencesController` occurs more frequently and the facade is also called like this, I would assume that it should be called `PublicRelationshipTemplateReferencesController`. 

Please use search functions to check that the other variant (`PublicRelationshipTemplateController`) is changed to the correct variant.

britsta (dismissed):

packages/runtime/src/Runtime.ts
jkoenig134: Sigleton is only for anonymous Controllers. So this is wrong and will explode when more than one account exists.

packages/runtime/src/extensibility/facades/transport/PublicRelationshipTemplateReferencesFacade.ts
```diff
@@ -0,0 +1,12 @@
+import { Result } from "@js-soft/ts-utils";
+import { Inject } from "@nmshd/typescript-ioc";
+import { PublicRelationshipTemplateReferenceDTO } from "../../../types/transport/PublicRelationshipTemplateReferenceDTO";
+import { GetPublicRelationshipTemplateReferencesUseCase } from "../../../useCases/transport/publicRelationshipTemplateReferences/GetPublicRelationshipTemplateReferences";
```
jkoenig134, line 4: missing barrels

sebbi08: This would result in circular deps that why I use the direct imports here

jkoenig134: I don't know why it should result in circular dependencies here but nowhere else.

sebbi08: Ok, we can use `type` and `use cases` but that actually is not lower the import count

jkoenig134: that's not the point at all. We want to make it easy to add new usecases, you just made it easy for yourself.

sebbi08: Ah, ok i mostly only use auto import, so I was not aware

.dev/appsettings.override.json
```diff
@@ -86,6 +86,25 @@
...
+                        "title": "b",
+                        "description": "Description",
+                        "truncatedReference": "Reference"
+                    },
+                    {
+                        "title": "c",
+                        "description": "Description",
+                        "truncatedReference": "Reference"
+                    }
+                ]
```
jkoenig134, line 107: hmm, shouldn't we mock this call instead? I'd be a bit careful about defining something here and expecting it to work, b/c when you bump up the backbone soon this will fail. Also now you loose the ability to test the 404 behaviour.

packages/runtime/src/useCases/transport/publicRelationshipTemplateReferences/GetPublicRelationshipTemplateReferences.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export class GetPublicRelationshipTemplateReferencesUseCase extends UseCase<void, PublicRelationshipTemplateReferenceDTO[]> {
+    public constructor(@Inject private readonly publicRelationshipTemplateReferencesController: PublicRelationshipTemplateReferencesController) {
+        super();
+    }
+
+    protected async executeInternal(): Promise<Result<PublicRelationshipTemplateReferenceDTO[]>> {
+        const templateReferences = (await this.publicRelationshipTemplateReferencesController.getPublicRelationshipTemplateReferences()).map((reference) => {
+            return PublicRelationshipTemplateReferenceMapper.toPublicRelationshipTemplateReferenceDTO(reference);
+        });
```
jkoenig134, line 16: please move this to 2 calls, or do the mapping in Result ok - AND the mapper should have a function to transform a list of PublicRelationshipTemplateReferences

packages/runtime/src/useCases/transport/publicRelationshipTemplateReferences/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./GetPublicRelationshipTemplateReferences";
+export * from "./PublicRelationshipTemplateReferenceMapper";
```
jkoenig134, line 2: why do you export the mapper?

packages/runtime/test/transport/publicRelationshipTemplateReferences.test.ts
```diff
@@ -0,0 +1,18 @@
...
+}, 30000);
+afterAll(() => serviceProvider.stop());
+
+describe("PublicRelationshipTemplateReferences", () => {
+    test("get PublicRelationshipTemplateReferences", async () => {
+        const publicRelationshipTemplates = await runtimeServices.transport.publicRelationshipTemplateReferences.getPublicRelationshipTemplateReferences();
+
+        expect(publicRelationshipTemplates.value).toBeDefined();
+        expect(publicRelationshipTemplates.value.length).toBeGreaterThan(0);
+    });
```
jkoenig134, line 17: this could fail soon

packages/transport/src/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.ts
```diff
@@ -0,0 +1,38 @@
...
+    public async getPublicRelationshipTemplateReferences(): Promise<PublicRelationshipTemplateReference[]> {
+        let publicRelationshipTemplateReferences: PublicRelationshipTemplateReference[];
+        try {
+            const backbonePublicRelationshipTemplateReferencesResult = await this.client.getPublicRelationshipTemplateReferences();
+            publicRelationshipTemplateReferences = backbonePublicRelationshipTemplateReferencesResult.value.map((reference) =>
+                PublicRelationshipTemplateReference.fromAny(reference)
+            );
+        } catch (e) {
+            if (e instanceof RequestError && e.status === 404) {
+                publicRelationshipTemplateReferences = [];
```
jkoenig134, line 31:
```suggestion
                return [];
```

packages/transport/src/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.ts
```diff
@@ -0,0 +1,38 @@
...
+        this.client = new PublicRelationshipTemplateReferenceClient(this.config, this.parent.authenticator, this.transport.correlator);
+
+        return this;
+    }
+
+    public async getPublicRelationshipTemplateReferences(): Promise<PublicRelationshipTemplateReference[]> {
+        let publicRelationshipTemplateReferences: PublicRelationshipTemplateReference[];
+        try {
+            const backbonePublicRelationshipTemplateReferencesResult = await this.client.getPublicRelationshipTemplateReferences();
+            publicRelationshipTemplateReferences = backbonePublicRelationshipTemplateReferencesResult.value.map((reference) =>
```
jkoenig134, line 26:
```suggestion
            return backbonePublicRelationshipTemplateReferencesResult.value.map((reference) =>
```

jkoenig134: I would love to have this in two lines.

packages/transport/src/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.ts
```diff
@@ -0,0 +1,38 @@
...
+                PublicRelationshipTemplateReference.fromAny(reference)
+            );
+        } catch (e) {
+            if (e instanceof RequestError && e.status === 404) {
+                publicRelationshipTemplateReferences = [];
+            } else {
```
jkoenig134, line 32: else is not needed. early return please.

packages/transport/test/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.test.ts
jkoenig134, line 1: you could make the whole test way cleaner by makeing it possible to inject a client to the Controller and just instanciate the controller here instead of testing this end to end.

jkoenig134 (changes requested):

packages/transport/test/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.test.ts
```diff
@@ -0,0 +1,74 @@
...
+            {
+                title: "b",
+                description: "Description",
+                truncatedReference: "Reference"
+            },
+            {
+                title: "c",
+                description: "Description",
+                truncatedReference: "Reference"
+            }
```
Milena-Czierlinski, line 53: You could use more descriptive values. :)

packages/runtime/test/transport/publicRelationshipTemplateReferences.test.ts
```diff
@@ -0,0 +1,50 @@
...
+            {
+                title: "b",
+                description: "Description",
+                truncatedReference: "Reference"
+            },
+            {
+                title: "c",
+                description: "Description",
+                truncatedReference: "Reference"
+            }
```
Milena-Czierlinski, line 39: Also here you could use more descriptive values.

packages/runtime/test/transport/publicRelationshipTemplateReferences.test.ts
```diff
@@ -0,0 +1,50 @@
...
+        ];
+
+        clientMocker.mockMethod("getPublicRelationshipTemplateReferences", () => {
+            return ClientResult.ok(mockResponse);
+        });
+
+        const publicRelationshipTemplateReferences = await runtimeServices.transport.publicRelationshipTemplateReferences.getPublicRelationshipTemplateReferences();
+
+        expect(publicRelationshipTemplateReferences.value).toStrictEqual(mockResponse);
```
Milena-Czierlinski, line 48:
```suggestion
        const publicRelationshipTemplateReferences = await runtimeServices.transport.publicRelationshipTemplateReferences.getPublicRelationshipTemplateReferences();
        expect(publicRelationshipTemplateReferences.isSuccess).toBe(true);
        expect(publicRelationshipTemplateReferences.value).toStrictEqual(mockResponse);
```

packages/transport/test/testHelpers/RestClientMocker.ts
```diff
@@ -0,0 +1,20 @@
...
+    public mockMethod(fnName: keyof T, callback: () => ClientResult<any>): void {
+        this.mockedMethods.push({ fnName, originalProperty: this.client[fnName] });
+
+        const mockFn = jest.fn().mockImplementation(() => {
+            return Promise.resolve(callback());
+        });
+        this.client[fnName] = mockFn.bind(this.client);
+    }
+
+    public restore(): void {
```
Milena-Czierlinski, line 15:
```suggestion
    public reset(): void {
```

packages/runtime/test/lib/RestClientMocker.ts
```diff
@@ -0,0 +1,20 @@
...
+    public mockMethod(fnName: keyof T, callback: () => ClientResult<any>): void {
+        this.mockedMethods.push({ fnName, originalProperty: this.client[fnName] });
+
+        const mockFn = jest.fn().mockImplementation(() => {
+            return Promise.resolve(callback());
+        });
+        this.client[fnName] = mockFn.bind(this.client);
+    }
+
+    public restore(): void {
```
Milena-Czierlinski, line 15:
```suggestion
    public reset(): void {
```

packages/runtime/test/transport/publicRelationshipTemplateReferences.test.ts
sebbi08: I like the one without the success more because when there is an error we only see that success should be true. But when directly checking value we will see the actual error that occurred

sebbi08: I actually like the one without the success check more. As for the success check in an error case, we will only get the message that isSuccess should be true. Without that check, the error that occurred is displayed.

Milena-Czierlinski: Okay, I think often we do have the isSuccess check, but it's no big issue for me.

.dev/appsettings.override.json
Milena-Czierlinski, line 1: Do we need tags in this PR? Or is this required due to the bumped Backbone version?

sebbi08: Yes the new backbone needs that config

packages/runtime/src/useCases/transport/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferenceMapper.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export class PublicRelationshipTemplateReferenceMapper {
+    public static toPublicRelationshipTemplateReferenceDTO(publicRelationshipTemplateReference: PublicRelationshipTemplateReference): PublicRelationshipTemplateReferenceDTO {
+        return {
+            title: publicRelationshipTemplateReference.title,
+            description: publicRelationshipTemplateReference.description,
+            truncatedReference: publicRelationshipTemplateReference.truncatedReference
+        };
+    }
+    public static toPublicRelationshipTemplateReferenceDTOList(
```
jkoenig134, line 12:
```suggestion
    }

    public static toPublicRelationshipTemplateReferenceDTOList(
```

packages/runtime/test/transport/publicRelationshipTemplateReferences.test.ts
```diff
@@ -0,0 +1,50 @@
...
+import { RestClientMocker } from "../lib/RestClientMocker";
+
+const serviceProvider = new RuntimeServiceProvider();
+let runtimeServices: TestRuntimeServices;
+let clientMocker: RestClientMocker<PublicRelationshipTemplateReferenceClient>;
+
+beforeAll(async () => {
+    runtimeServices = (await serviceProvider.launch(1))[0];
+    const client = (runtimeServices.transport.publicRelationshipTemplateReferences as any).getPublicRelationshipTemplateReferencesUseCase
+        .publicRelationshipTemplateReferencesController.client as PublicRelationshipTemplateReferenceClient;
```
jkoenig134, line 12: we don't do any casts to access private variables. instead you can access them by using `aController["privateVariable"]`

packages/transport/test/testHelpers/RestClientMocker.ts
jkoenig134, line 1: I'm not sure if I like the mocking pattern. There is another mocker like this which is OLD, and we try to mock everything with mockito now.

Also you could simplify your test by only testing the PublicRelationshipTemplateReferencesController without having to do a whole transport init and so on.

packages/transport/test/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.test.ts
```diff
@@ -0,0 +1,74 @@
...
+        ];
+
+        clientMocker.mockMethod("getPublicRelationshipTemplateReferences", () => {
+            return ClientResult.ok(mockResponse);
+        });
+
+        const publicRelationshipTemplates = await account.publicRelationshipTemplateReferences.getPublicRelationshipTemplateReferences();
+
+        expect(publicRelationshipTemplates.map((references) => references.toJSON())).toStrictEqual(mockResponse);
```
jkoenig134, line 62:
```suggestion
        expect(publicRelationshipTemplates.map((reference) => reference.toJSON())).toStrictEqual(mockResponse);
```

naming. you are mapping over a list of references (that you just call publicRelationshipTemplates here which I also don't like) -> in this case you have to name it reference as it will be only one reference and not multiple.

packages/transport/src/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public async getPublicRelationshipTemplateReferences(): Promise<PublicRelationshipTemplateReference[]> {
+        try {
+            const backbonePublicRelationshipTemplateReferencesResult = await this.client.getPublicRelationshipTemplateReferences();
+            return backbonePublicRelationshipTemplateReferencesResult.value.map((reference) => PublicRelationshipTemplateReference.fromAny(reference));
+        } catch (e) {
+            if (e instanceof RequestError && e.status === 404) {
+                return [];
+            }
```
jkoenig134, line 29: can be inline

packages/transport/src/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.ts
```diff
@@ -0,0 +1,33 @@
...
+
+        this.client = new PublicRelationshipTemplateReferenceClient(this.config, this.parent.authenticator, this.transport.correlator);
+
+        return this;
+    }
+
+    public async getPublicRelationshipTemplateReferences(): Promise<PublicRelationshipTemplateReference[]> {
+        try {
+            const backbonePublicRelationshipTemplateReferencesResult = await this.client.getPublicRelationshipTemplateReferences();
+            return backbonePublicRelationshipTemplateReferencesResult.value.map((reference) => PublicRelationshipTemplateReference.fromAny(reference));
```
jkoenig134, line 25: for debugging it would be great to store the mapped references in a variable

packages/runtime/test/transport/publicRelationshipTemplateReferences.test.ts
```diff
@@ -0,0 +1,50 @@
...
+            {
+                title: "aTitle",
+                description: "aDescription",
+                truncatedReference: "aReference"
+            },
+            {
+                title: "aTitle",
+                description: "aDescription",
+                truncatedReference: "aReference"
+            }
```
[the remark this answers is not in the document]
jkoenig134: yeah, very important.

packages/runtime/test/transport/publicRelationshipTemplateReferences.test.ts
jkoenig134: same for `packages/transport/test/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.test.ts` btw.

packages/runtime/test/transport/publicRelationshipTemplateReferences.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    const client = (runtimeServices.transport.publicRelationshipTemplateReferences as any).getPublicRelationshipTemplateReferencesUseCase
+        .publicRelationshipTemplateReferencesController.client as PublicRelationshipTemplateReferenceClient;
+    clientMocker = new RestClientMocker(client);
+}, 30000);
+
+afterAll(() => serviceProvider.stop());
+
+afterEach(() => {
+    clientMocker.reset();
+});
```
jkoenig134, line 20: why isn't this a one-liner?

packages/transport/test/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+afterAll(async () => {
+    await account.close();
+
+    await connection.close();
+});
+
+afterEach(() => {
+    clientMocker.reset();
+});
```
jkoenig134, line 34: why isn't this a one-liner?

packages/transport/test/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.test.ts
```diff
@@ -0,0 +1,74 @@
...
+            {
+                title: "aTitle",
+                description: "aDescription",
+                truncatedReference: "aReference"
+            },
+            {
+                title: "aTitle",
+                description: "aDescription",
+                truncatedReference: "aReference"
+            }
```
jkoenig134, line 53:
```suggestion
            {
                title: "aTitle",
                description: "aDescription",
                truncatedReference: "aReference"
            }
```
should be fine if there's only one. We only want to see if it's same in as out.

packages/transport/test/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    await connection.close();
+});
+
+afterEach(() => {
+    clientMocker.reset();
+});
+
+describe("PublicRelationshipTemplateReferencesController", () => {
+    test("should return the backbone defined PublicRelationshipTemplateReferences", async () => {
```
jkoenig134, line 37: I'd like to see a third test where there is an empty list, that this is correctly handled.

jkoenig134 (changes requested):

packages/transport/test/testHelpers/RestClientMocker.ts
sebbi08: Ah I was not aware of mockito changed that is actually doing the same i made but already there so i changed to it.

Regarding the controller it is true, but then we would have no test that is testing the initialization of the tags controller.

packages/transport/test/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.test.ts
jkoenig134: yeah, we should stick to codestyle from anywhere else and use the shorthand for oneliners

packages/transport/test/modules/publicRelationshipTemplateReferences/PublicRelationshipTemplateReferencesController.test.ts
```diff
@@ -0,0 +1,68 @@
...
+let account: AccountController;
+let mockedClient: PublicRelationshipTemplateReferenceClient;
+
+beforeAll(async function () {
+    connection = await TestUtil.createDatabaseConnection();
+    transport = TestUtil.createTransport(connection);
+
+    await transport.init();
+
+    const accounts = await TestUtil.provideAccounts(transport, 3);
```
Milena-Czierlinski, line 18:
```suggestion
    const accounts = await TestUtil.provideAccounts(transport, 1);
```

Milena-Czierlinski (commented):
The rest LGTM.

Milena-Czierlinski (approved):

jkoenig134 (approved):

Milena-Czierlinski (approved):

PR #337, opened 2024-11-20
Attributes can be succeeded without changing their content

Milena-Czierlinski:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/packages/consumption/src/modules/attributes/AttributesController.ts
+++ b/packages/consumption/src/modules/attributes/AttributesController.ts
@@ -979,10 +979,6 @@ export class AttributesController extends ConsumptionBaseController {
             return ValidationResult.error(ConsumptionCoreErrors.attributes.cannotSucceedAttributesWithASuccessor(predecessor.succeededBy.toString()));
         }
 
-        if (_.isEqual(parsedSuccessorParams.content, predecessor.content)) {
-            return ValidationResult.error(ConsumptionCoreErrors.attributes.successionMustChangeContent());
-        }
-
         if (predecessor.parentId) {
             return ValidationResult.error(ConsumptionCoreErrors.attributes.cannotSucceedChildOfComplexAttribute(predecessorId.toString()));
         }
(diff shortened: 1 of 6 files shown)


packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -979,6 +979,10 @@ export class AttributesController extends ConsumptionBaseController {
             return ValidationResult.error(ConsumptionCoreErrors.attributes.cannotSucceedAttributesWithASuccessor(predecessor.succeededBy.toString()));
         }
 
+        if (_.isEqual(parsedSuccessorParams.content, predecessor.content)) {
+            return ValidationResult.error(ConsumptionCoreErrors.attributes.successionMustChangeContent());
+        }
+
```
britsta, line 985: Why'd you decided to put this error here?

One could also argue to throw the check after error `cannotSucceedAttributesWithDeletionInfo`. (If possible, errors should be thrown in such a way that correcting the cause of the error cannot lead directly to the next error being thrown.)

Milena-Czierlinski:
I did some rearranging of the errors, that is partly independent of this PR, as a result of our discussion. I hope you don't mind it mixing up a bit, since the PR is still quite small.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -995,8 +999,8 @@ export class AttributesController extends ConsumptionBaseController {
             return ValidationResult.error(ConsumptionCoreErrors.attributes.successionMustNotChangeValueType());
         }
 
-        if (predecessor.hasDeletionInfo() && predecessor.deletionInfo.deletionStatus !== LocalAttributeDeletionStatus.DeletionRequestRejected) {
-            return ValidationResult.error(ConsumptionCoreErrors.attributes.cannotSucceedAttributesWithDeletionInfo());
+        if (successor.succeeds && !predecessorId.equals(successor.succeeds.toString())) {
+            return ValidationResult.error(ConsumptionCoreErrors.attributes.setPredecessorIdDoesNotMatchActualPredecessorId());
```
Magnus-Kuhn, line 1003: Why did you put this at the bottom? I'm not sure if it's possible but could there be a way that the content check fails and then the user tries to tweak with the content only to run into this error?

jkoenig134 (approved):

packages/consumption/src/modules/attributes/AttributesController.ts
Milena-Czierlinski: @Magnus-Kuhn That could in fact happen. But if there is an issue with the content and the manually set predecessorId, the second error will always be thrown fixing the first, no matter which way I put them. And since I think that fixing the content is more important, while simply omitting the manually set predecessorId is super simple, I decided to place them this way around.

PR #338, opened 2024-11-21
Improve CreateRepositoryAttribute error messages

Magnus-Kuhn:
# Readiness checklist

-   [x] I labeled the PR.

# Description

Validates attribute contents in the CreateRepositoryAttribute use case with serval instead of the ajv schema validator. Includes added tests with the resulting error messages when validations fail. It's not perfect (you can have additional properties and the errors of complex attributes are also unclear).
It's not clear that this approach is more promising than attempting to improve the ajv usage.


packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
```diff
@@ -9,7 +8,7 @@ import { AttributeMapper } from "./AttributeMapper";
 
 export interface CreateRepositoryAttributeRequest {
     content: {
-        value: AttributeValues.Identity.Json;
+        value: any;
```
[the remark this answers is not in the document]
Magnus-Kuhn: It looks like this:
1. If no `@type` is set, there is an error that `IdentityAttribute.value` is not of an allowed type
2. If `@type` is set and is not a correct type, a reflection class error is thrown
3. If `@type` is set and the value doesn't fit to that, an error for the set `@type` is thrown
4. If `@type` with fitting value is set that is not an IdentityAttribute, there is an error that `IdentityAttribute.value` is not of an allowed type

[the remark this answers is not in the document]
jkoenig134: Hm, sounds great. We should discuss this in the next sync.

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
```diff
@@ -16,9 +17,29 @@ export interface CreateRepositoryAttributeRequest {
...
+        const requestValidationResult = requestSchemaValidator.validate(value);
+        if (requestValidationResult.isInvalid()) return requestValidationResult;
+
+        const valueType = value.content.value["@type"];
+        if (!AttributeValues.Identity.TYPE_NAMES.includes(valueType)) {
+            const validationResult = new ValidationResult();
+
+            validationResult.addFailure(
+                new ValidationFailure(
+                    RuntimeErrors.general.invalidPropertyValue("content.value.@type must match one of the allowed attribute value types for IdentityAttributes"),
```
britsta, line 34: Please write "Attribute" capitalized.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -749,6 +749,100 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
         const attribute = result.value;
         expect(attribute.isDefault).toBeUndefined();
     });
+
+    describe("validation errors for the attribute content", () => {
```
britsta, line 753: Please write "Attribute" capitalized.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -749,6 +749,100 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
...
+                        "@type": "BirthMonth",
+                        value: 14
+                    },
+                    tags: ["tag1", "tag2"]
+                } as any
+            };
+            const result = await services1.consumption.attributes.createRepositoryAttribute(request);
+            expect(result.error.message).toBe("value must be equal to one of the allowed values");
+            expect(result.error.code).toBe("error.runtime.validation.invalidPropertyValue");
+        });
```
britsta, line 815: Does this test causes redundancy as we already have the test `"returns an error when trying to create an invalid BirthDate with violated validation criteria of a single property"` within the file `BirthDate.test.ts`? Within the mentioned test, the message `BirthMonth.value:Number :: must be an integer value between 1 and 12"` is used, which I prefer compared to the more generic `"value must be equal to one of the allowed values"` message. However, I am aware that I am comparing two tests from different libraries (Content and Runtime).

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -749,6 +749,100 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
...
+
+        test("should not accept an additional property", async () => {
+            const request: CreateRepositoryAttributeRequest = {
+                content: {
+                    value: {
+                        "@type": "GivenName",
+                        value: "Petra Pan",
```
britsta, line 822: Please use a more descriptive `value` here  than "Petra Pan". :D

packages/runtime/test/consumption/attributes.test.ts
Magnus-Kuhn: I'd rather leave the test namings to a different discussion/PR - capitalization is infrequent there :)

packages/runtime/test/consumption/attributes.test.ts
Magnus-Kuhn: The content error message comes from the serval, whereas the error here comes from ajv. I agree that the serval error is better but I'll have to see if there's a good way to improve the error here

packages/runtime/src/types/ajv/CreateRepositoryAttributeRequest.ts
jkoenig134, line 1: why is this in a new file?

jkoenig134: Maybe the both can also inherit

jkoenig134 (approved):

PR #339, opened 2024-11-25
Fix flaky test

sebbi08:
# Readiness checklist

-   [ ] I ensured that the PR title is good enough for the changelog.
-   [x] I labeled the PR.

### DIFF
--- a/packages/consumption/test/modules/requests/IncomingRequestsController.test.ts
+++ b/packages/consumption/test/modules/requests/IncomingRequestsController.test.ts
@@ -66,7 +66,7 @@ describe("IncomingRequestsController", function () {
         });
 
         test("takes the expiration date from the Template if the Request has no expiration date", async function () {
-            const timestamp = CoreDate.utc().subtract({ hours: 1 });
+            const timestamp = CoreDate.utc().subtract({ second: 1 });
             const incomingTemplate = TestObjectFactory.createIncomingRelationshipTemplate(timestamp);
             await When.iCreateAnIncomingRequestWith({ requestSourceObject: incomingTemplate });
             await Then.theRequestHasExpirationDate(timestamp);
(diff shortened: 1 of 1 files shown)


packages/consumption/test/modules/requests/IncomingRequestsController.test.ts
```diff
@@ -66,7 +66,7 @@ describe("IncomingRequestsController", function () {
         });
 
         test("takes the expiration date from the Template if the Request has no expiration date", async function () {
-            const timestamp = CoreDate.utc();
+            const timestamp = CoreDate.utc().subtract({ hours: 1 });
```
Magnus-Kuhn, line 69: Do you have an explanation why this test could be flaky?

Also would it be possible to fix this by subtracting only e. g. a couple of seconds? I think that this is a tolerable margin of error, so if this fixes the test, there might be no need to investigate further

sebbi08: Hi, it is because of tests like this:

https://github.com/nmshd/runtime/actions/runs/12005423256/job/33462001888

So in the test the request is still open and not expired and the only point where the status is set is here
https://github.com/nmshd/runtime/blob/b78593f7a8e8a0c24bc9d3383fc2cd7cf53e3b59/packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts#L64-L66

So one possibility would be that there is already `getExistingRelationshipToIdentity` that is mocked in the test and will always return `undefined`.

In the `updateStatusBasedOnExpiration` method, the date is compared with the current time with `>` so if both times are the same the request is considered to be not expired.

So it seems that the tests are too fast.

jkoenig134 (approved):
LGTM

PR #345, opened 2024-11-27
Add `deletionDate` to LocalAccount

Milena-Czierlinski:
# Readiness checklist

-   [x] I labeled the PR.


packages/app-runtime/src/modules/runtimeEvents/DatawalletSynchronizedModule.ts
```diff
@@ -33,7 +34,7 @@ export class DatawalletSynchronizedModule extends AppRuntimeModule<DatawalletSyn
 
         switch (mostRecentIdentityDeletionProcess.status) {
             case IdentityDeletionProcessStatus.Approved:
-                newDeletionDate = mostRecentIdentityDeletionProcess.gracePeriodEndsAt;
+                newDeletionDate = CoreDate.from(mostRecentIdentityDeletionProcess.gracePeriodEndsAt!);
```
jkoenig134, line 37: hmm, why do we need the null-check here? And is the null-check safe, b/c it's in status Approved?

jkoenig134 (changes requested):
some details, looks very good overall!

packages/app-runtime/src/multiAccount/data/LocalAccount.ts
```diff
@@ -1,14 +1,15 @@
...
-    id: CoreId;
-    address?: CoreAddress;
+    id: ICoreId;
+    address?: ICoreAddress;
     name: string;
     directory: string;
     order: number;
     lastAccessedAt?: ICoreDate;
     devicePushIdentifier?: string;
+    deletionDate?: ICoreDate;
```
Magnus-Kuhn, line 12: Why only the deletion date and not the status too?

jkoenig134: b/c we don't care. 

1. deletionDate set = in deletion
2. deletionDate not set = not in deletion

packages/app-runtime/src/modules/runtimeEvents/DatawalletSynchronizedModule.ts
Milena-Czierlinski: `gracePeriodEndsAt` is an optional property, that is only set if the status is Approved. We know that, but the compiler doesn't. I suppose we could add type guards for the IdentityDeletionProcess, but I'm not sure if it's worth the effort.

jkoenig134: Ok, that's all I wanted to know! Great!

packages/app-runtime/src/modules/runtimeEvents/IdentityDeletionProcessStatusChangedModule.ts
```diff
@@ -0,0 +1,44 @@
...
+                break;
+
+            case IdentityDeletionProcessStatus.Cancelled:
+                const account = await this.runtime.multiAccountController.getAccountByAddress(event.eventTargetAddress);
+                const previousDeletionDate = account.deletionDate;
+
+                if (!previousDeletionDate) break;
+
+                await this.runtime.multiAccountController.updateLocalAccountDeletionDate(event.eventTargetAddress, undefined);
+                break;
```
Magnus-Kuhn, line 34: no publishing of a `LocalAccountDeletionDateChangedEvent`?

packages/app-runtime/src/modules/runtimeEvents/DatawalletSynchronizedModule.ts
```diff
@@ -0,0 +1,56 @@
...
+        const services = await this.runtime.getServices(event.eventTargetAddress);
+        const identityDeletionProcessResult = await services.transportServices.identityDeletionProcesses.getIdentityDeletionProcesses();
+
+        if (identityDeletionProcessResult.isError) {
+            this.logger.error(identityDeletionProcessResult);
+            return;
+        }
+
+        let newDeletionDate;
+        const mostRecentIdentityDeletionProcess = identityDeletionProcessResult.value[identityDeletionProcessResult.value.length - 1];
```
Magnus-Kuhn, line 33:
```suggestion
        const mostRecentIdentityDeletionProcess = identityDeletionProcessResult.value.at(-1)
```

packages/app-runtime/src/multiAccount/MultiAccountController.ts
```diff
@@ -238,6 +251,22 @@ export class MultiAccountController {
...
+        if (!oldAccount) {
+            throw TransportCoreErrors.general.recordNotFound(LocalAccount, address).logWith(this._log);
+        }
+
+        const account = LocalAccount.from(oldAccount);
+
+        account.deletionDate = deletionDate ?? undefined;
+        await this._localAccounts.update(oldAccount, account);
+
+        const cachedAccount = this.sessionStorage.findSession(address)?.account;
```
Milena-Czierlinski, line 266: @jkoenig134 Should this always be defined? (I would like to return the updated cachedAccount, ideally knowing that it isn't undefined.)

jkoenig134: I think you could run "findOrCreateSession" IDK if it's called like this to make sure that it is defined. But I'd be careful about this cache and not rely on it too much.

packages/app-runtime/src/modules/runtimeEvents/DatawalletSynchronizedModule.ts
jkoenig134: I think we should use this at many more places (in another refactoring PR :) )

packages/app-runtime/src/modules/runtimeEvents/IdentityDeletionProcessStatusChangedModule.ts
jkoenig134: nah, not needed when the change was made by the current device

packages/transport/src/modules/accounts/AccountController.ts
```diff
@@ -230,11 +230,9 @@ export class AccountController {
...
 
     public async syncDatawallet(force = false): Promise<void> {
-        if (!force && !this.autoSync) {
-            return;
-        }
+        if (!force && !this.autoSync) return;
 
-        return await this.synchronization.sync("OnlyDatawallet");
+        await this.synchronization.sync("OnlyDatawallet");
     }
```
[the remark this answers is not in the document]
jkoenig134: that's a VERY bad advice and goes against everything we work for making the codebase cleaner.

packages/app-runtime/test/modules/DatawalletSynchronizedModule.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        await sessionDevice2.transportServices.account.syncDatawallet();
+        const event = await TestUtil.awaitEvent(runtimeDevice2, LocalAccountDeletionDateChangedEvent);
+        expect(event.data).toBe(initiateDeletionResult.value.gracePeriodEndsAt);
+
+        expect(sessionDevice2.account.deletionDate).toBe(initiateDeletionResult.value.gracePeriodEndsAt);
+
+        const account = await runtimeDevice2.multiAccountController.getAccount(CoreId.from(sessionDevice2.account.id));
+        expect(account.deletionDate!.toString()).toBe(initiateDeletionResult.value.gracePeriodEndsAt);
+    });
```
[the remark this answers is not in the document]
jkoenig134: It's a double check for persistence

packages/app-runtime/src/multiAccount/MultiAccountController.ts
```diff
@@ -238,6 +251,22 @@ export class MultiAccountController {
...
+        if (cachedAccount) cachedAccount.deletionDate = deletionDate?.toString();
```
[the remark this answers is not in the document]
jkoenig134: yeah, we actually do in our tests.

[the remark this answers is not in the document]
jkoenig134: That's exactly what your `What's the difference between this account and the account above?` question is about.

packages/app-runtime/src/modules/runtimeEvents/DatawalletSynchronizedModule.ts
jkoenig134: ^ done in #348

jkoenig134 (approved):
LGTM

Milena-Czierlinski:
@Siolto Please note that the DatawalletSynchronizedEvent is now published in transport instead of the app-runtime, which is why its namespace changes from `"app.datawalletSynchronized"` to `"transport.datawalletSynchronized"`.

jkoenig134:
@Magnus-Kuhn please re-check your comments and resolve if you are ok with the answers. If not we have to discuss the things.

PR #347, opened 2024-11-28
Simplify identity deletion process handling on second device by republishing events

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.


packages/transport/src/modules/accounts/AccountController.ts
```diff
@@ -235,11 +236,29 @@ export class AccountController {
...
+        const changedItems = await this.synchronization.sync("Everything");
+        await this.triggerEventsForChangedItems(changedItems);
+
+        return changedItems;
+    }
+
+    private async triggerEventsForChangedItems(changedItems: ChangedItems) {
+        const changedIdentityDeletionProcessIds = changedItems.changedObjectIdentifiersDuringDatawalletSync.filter((x) => x.startsWith("IDP"));
+        for (const id of changedIdentityDeletionProcessIds) {
+            const changedIdentityDeletionProcesses = await this.identityDeletionProcess.getIdentityDeletionProcess(id);
```
britsta, line 253: Switch the variable name to singular `changedIdentityDeletionProcess` please as at most one IdentityDeletionProcess can be returned by the method call `await this.identityDeletionProcess.getIdentityDeletionProcess(id)`.

packages/transport/src/modules/accounts/AccountController.ts
```diff
@@ -235,11 +236,29 @@ export class AccountController {
...
     public async syncEverything(): Promise<ChangedItems> {
-        return await this.synchronization.sync("Everything");
+        const changedItems = await this.synchronization.sync("Everything");
+        await this.triggerEventsForChangedItems(changedItems);
+
+        return changedItems;
+    }
+
+    private async triggerEventsForChangedItems(changedItems: ChangedItems) {
+        const changedIdentityDeletionProcessIds = changedItems.changedObjectIdentifiersDuringDatawalletSync.filter((x) => x.startsWith("IDP"));
```
britsta, line 251: I'd prefer `const idsOfChangedIdentityDeletionProcesses` here.

packages/transport/src/modules/accounts/AccountController.ts
```diff
@@ -235,11 +236,29 @@ export class AccountController {
...
+        for (const id of changedIdentityDeletionProcessIds) {
+            const changedIdentityDeletionProcesses = await this.identityDeletionProcess.getIdentityDeletionProcess(id);
+            if (!changedIdentityDeletionProcesses) {
+                this.log.error(`IdentityDeletionProcess with id ${id} not found for re-triggering event. Skipping.`);
```
britsta, line 255: Please consider formatting the ID string like this: `'${id}'`.

packages/app-runtime/test/modules/IdentityDeletionProcessStatusChangedModule.test.ts
```diff
@@ -22,23 +27,34 @@ describe("IdentityDeletionProcessStatusChanged", function () {
...
         }
     });
 
-    afterAll(async () => await runtime.stop());
+    afterAll(async function () {
+        await runtime.stop();
+        await runtimeDevice2?.stop();
 
-    test("should set the deletionDate on the LocalAccount initiating an IdentityDeletionProcess", async function () {
+        // this eventbus is not defenitely cleaned up with runtime2, so it needs to be closed manually
```
britsta, line 42: `runtime2` isn't declared as variable in this test file, please write `runtimeDevice2` instead or rename the variable.

packages/transport/src/modules/sync/ChangedItems.ts
```diff
@@ -12,7 +12,8 @@ export class ChangedItems implements IChangedItems {
     public constructor(
         public readonly relationships: Relationship[] = [],
         public readonly messages: Message[] = [],
-        public readonly identityDeletionProcesses: IdentityDeletionProcess[] = []
+        public readonly identityDeletionProcesses: IdentityDeletionProcess[] = [],
+        public readonly changedObjectIdentifiersDuringDatawalletSync: string[] = []
```
britsta, line 16: Do we really need this in this form and in this place? It looks a little out of place among the `relationships`, `messages` and `identityDeletionProcesses` arrays. Isn't it possible, for example, to obtain the `id` of a changed `identityDeletionProcess` via the `identityDeletionProcesses` array?

packages/app-runtime/test/modules/IdentityDeletionProcessStatusChangedModule.test.ts
```diff
@@ -48,4 +64,51 @@ describe("IdentityDeletionProcessStatusChanged", function () {
...
+            const createDeviceResult = await session.transportServices.devices.createDevice({ name: "test", isAdmin: true });
+            const onboardingInfoResult = await session.transportServices.devices.getDeviceOnboardingInfo({ id: createDeviceResult.value.id, profileName: "Test" });
+            const localAccountDevice2 = await runtimeDevice2.accountServices.onboardAccount(onboardingInfoResult.value);
+            sessionDevice2 = await runtimeDevice2.selectAccount(localAccountDevice2.id.toString());
+
+            await session.transportServices.account.syncDatawallet();
+            await sessionDevice2.transportServices.account.syncDatawallet();
+        });
+
+        test("should set the deletionDate on the LocalAccount on a second device when an IdentityDeletionProcess is initiated", async function () {
```
britsta, line 82: I would also adapt the test name to "...deletionDate of the LocalAccount..." instead of "...deletionDate on the LocalAccount...", as you did with the tests above.

packages/app-runtime/test/modules/IdentityDeletionProcessStatusChangedModule.test.ts
```diff
@@ -48,4 +64,51 @@ describe("IdentityDeletionProcessStatusChanged", function () {
...
+
+            await expect(eventBus).toHavePublished(LocalAccountDeletionDateChangedEvent, (e) => e.data.deletionDate! === initiateDeletionResult.value.gracePeriodEndsAt!);
+
+            expect(sessionDevice2!.account.deletionDate!.toString()).toStrictEqual(initiateDeletionResult.value.gracePeriodEndsAt);
+
+            const account = await runtimeDevice2!.multiAccountController.getAccount(CoreId.from(sessionDevice2!.account.id));
+            expect(account.deletionDate!.toString()).toBe(initiateDeletionResult.value.gracePeriodEndsAt!.toString());
+        });
+
+        test("should unset the deletionDate on the LocalAccount on a second device when an IdentityDeletionProcess is cancelled", async function () {
```
britsta, line 96: I would also adapt the test name to "...deletionDate of the LocalAccount..." instead of "...deletionDate on the LocalAccount...", as you did with the tests above.

packages/transport/src/modules/accounts/AccountController.ts
jkoenig134: hmm. me not.

packages/app-runtime/src/modules/runtimeEvents/IdentityDeletionProcessStatusChangedModule.ts
```diff
@@ -21,23 +23,29 @@ export class IdentityDeletionProcessStatusChangedModule extends AppRuntimeModule
...
                 break;
         }
     }
 
+    private async updateLocalAccountDeletionDate(eventTargetAddress: string, deletionDate: CoreDate | undefined) {
+        const account = await this.runtime.multiAccountController.getAccountByAddress(eventTargetAddress);
+        const previousDeletionDate = account.deletionDate;
+
+        if (!deletionDate && !previousDeletionDate) return;
+        if (deletionDate && previousDeletionDate && deletionDate.equals(previousDeletionDate)) return;
```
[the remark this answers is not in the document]
jkoenig134: Yep, it's intentional. We cannot compare CoreDates with `==` so this is the cleanest way to do that.

packages/transport/src/modules/sync/ChangedItems.ts
jkoenig134: the other ones are for the external event syncs, the new one tracks changes during datawallet sync.

jkoenig134: The goal of exactly this implementation is, that we can easily add new events for other thing that are changes - without touching the Datawallet / Sync itself again, only the AccountController.

packages/app-runtime/src/modules/runtimeEvents/IdentityDeletionProcessStatusChangedModule.ts
```diff
@@ -19,25 +21,44 @@ export class IdentityDeletionProcessStatusChangedModule extends AppRuntimeModule
...
-                await this.runtime.multiAccountController.updateLocalAccountDeletionDate(event.eventTargetAddress, undefined);
+                await this.updateLocalAccountDeletionDate(event.eventTargetAddress, undefined);
                 break;
 
             default:
                 break;
         }
     }
 
+    private async updateLocalAccountDeletionDate(eventTargetAddress: string, deletionDate: CoreDate | undefined, publishEvent = false) {
```
Milena-Czierlinski, line 49: Would it be cleaner to adjust the name of the function such that it doesn't exactly match the name of the MultiAccountController functions, since we do more stuff here?

packages/app-runtime/src/modules/runtimeEvents/IdentityDeletionProcessStatusChangedModule.ts
```diff
@@ -19,25 +21,44 @@ export class IdentityDeletionProcessStatusChangedModule extends AppRuntimeModule
...
+        if (!identityDeletionProcess) {
+            const services = await this.runtime.getServices(event.eventTargetAddress);
+            const identityDeletionProcesses = await services.transportServices.identityDeletionProcesses.getIdentityDeletionProcesses();
+
+            const approvedIdentityDeletionProcess = identityDeletionProcesses.value.filter((idp) => idp.status === IdentityDeletionProcessStatus.Approved).at(0);
+            const deletionDate = approvedIdentityDeletionProcess?.gracePeriodEndsAt ? CoreDate.from(approvedIdentityDeletionProcess.gracePeriodEndsAt) : undefined;
+
+            await this.updateLocalAccountDeletionDate(event.eventTargetAddress, deletionDate, true);
+            return;
+        }
```
Milena-Czierlinski, line 33:
```suggestion
        let publishEvent = false;
        let identityDeletionProcess = event.data;

        if (!identityDeletionProcess) {
            const services = await this.runtime.getServices(event.eventTargetAddress);
            const identityDeletionProcesses = await services.transportServices.identityDeletionProcesses.getIdentityDeletionProcesses();
            const identityDeletionProcess = identityDeletionProcesses.value.at(0)!;
            publishEvent = true;
        }
```

Milena-Czierlinski: What about this? To me it seems simpler.

packages/transport/src/modules/sync/ChangedItems.ts
Milena-Czierlinski, line 16: If you change it here, won't it need to be adjusted in the SyncEverythingResponse, too?

packages/app-runtime/src/modules/runtimeEvents/IdentityDeletionProcessStatusChangedModule.ts
Milena-Czierlinski: But we save the filtering step before... Maybe filtering is better than the switch logic perfomance-wise, but I find it the other way around a bit easier to comprehend.

jkoenig134: ACtually the filtering does more than the switch as the `default:` is empty.

packages/transport/src/modules/sync/ChangedItems.ts
jkoenig134: Nope, I defenitely don't want to expose this.

packages/app-runtime/src/modules/runtimeEvents/IdentityDeletionProcessStatusChangedModule.ts
jkoenig134: I don't see a reason why matching function names are a problem.

Milena-Czierlinski (approved):

britsta (approved):

PR #353, opened 2024-11-29
Query RelationshipTemplates for password protection

Magnus-Kuhn:
# Readiness checklist

-   [X] I labeled the PR.
-   [X] I self-reviewed the PR.


packages/runtime/test/transport/relationshipTemplates.test.ts
```diff
@@ -327,4 +327,68 @@ describe("RelationshipTemplates query", () => {
...
+                    "passwordProtection.passwordIsPin": "true"
+                }
+            })
+        ).value;
+        const templateIdsWithPin = templatesWithPin.map((t) => t.id);
+
+        expect(templateIdsWithPin).not.toContain(passwordProtectedTemplate.id);
+        expect(templateIdsWithPin).toContain(pinProtectedTemplate.id);
+        expect(templateIdsWithPin).not.toContain(unprotectedTemplate.id);
+    });
```
jkoenig134, line 393: please see the other tests that do querying, we have a fabulous tester for querystrings :)

jkoenig134 (changes requested):

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
```diff
@@ -51,6 +55,23 @@ export class GetRelationshipTemplatesUseCase extends UseCase<GetRelationshipTemp
...
+                    query[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.passwordType)}`] = {
+                        $regex: "^pin"
+                    };
+                }
+                if (input === "!true") {
+                    query[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.passwordType)}`] = {
+                        $in: ["pw", undefined]
+                    };
+                }
+            }
```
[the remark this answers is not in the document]
jkoenig134: argh, I see. Then we'll have to discuss on how to align these custom thingies to our query translator.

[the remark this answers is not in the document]
jkoenig134: Maybe it could be enough to do `!` instead of `!true`

packages/runtime/test/transport/relationshipTemplates.test.ts
```diff
@@ -327,4 +327,54 @@ describe("RelationshipTemplates query", () => {
 
         await conditions.executeTests((c, q) => c.relationshipTemplates.getRelationshipTemplates({ query: q, ownerRestriction: OwnerRestriction.Peer }));
     });
+
+    test("password query with password", async () => {
```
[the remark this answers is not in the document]
Magnus-Kuhn: I kept password query because `password` is the queried property

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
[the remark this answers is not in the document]
Magnus-Kuhn: I just stumbled across the BooleanFilter in the connector's `openapi` - the operators there are `?foo=true` and `?foo=!true`

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
```diff
@@ -51,6 +55,23 @@ export class GetRelationshipTemplatesUseCase extends UseCase<GetRelationshipTemp
                 (t) => t.maxNumberOfAllocations
             )}`,
             [nameof<RelationshipTemplateDTO>((r) => r.forIdentity)]: `${nameof<RelationshipTemplate>((r) => r.cache)}.${nameof<CachedRelationshipTemplate>((t) => t.forIdentity)}`
+        },
+        custom: {
+            [`${nameof<RelationshipTemplateDTO>((r) => r.passwordProtection)}.password`]: (query: any, input: string) => {
+                query[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.password)}`] = input;
+            },
```
jkoenig134, line 62: this should just be an alias

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
```diff
@@ -51,6 +55,23 @@ export class GetRelationshipTemplatesUseCase extends UseCase<GetRelationshipTemp
...
+            [`${nameof<RelationshipTemplateDTO>((t) => t.passwordProtection)}.passwordIsPin`]: (query: any, input: string) => {
+                if (input === "true") {
+                    query[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.passwordType)}`] = {
+                        $regex: "^pin"
+                    };
+                }
+                if (input === "!") {
+                    query[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.passwordType)}`] = {
+                        $in: ["pw", undefined]
```
jkoenig134, line 71:
```suggestion
                        $in: ["pw", undefined]
```

why do you want to say with `$in: undefined`? The field is never empty, right? So we should match `"pw"` exactly

packages/runtime/test/transport/relationshipTemplates.test.ts
```diff
@@ -327,4 +327,54 @@ describe("RelationshipTemplates query", () => {
...
+        const passwordProtectedTemplate = await createTemplate(runtimeServices1.transport, undefined, { password: "password" });
+
+        const conditions = new QueryParamConditions<GetRelationshipTemplatesQuery>(passwordProtectedTemplate, runtimeServices1.transport).addStringSet(
+            "passwordProtection.password"
+        );
+        await conditions.executeTests((c, q) => c.relationshipTemplates.getRelationshipTemplates({ query: q }));
+    });
+
+    test("password query for a template protected with a PIN", async () => {
+        const pinProtectedTemplate = await createTemplate(runtimeServices1.transport, undefined, { password: "1234", passwordIsPin: true });
```
jkoenig134, line 341: actually it should be possible to add this test (and the other one) to the querying test above, just add the password protection and the filters.

jkoenig134: You can also add some completely custom queries and define if it should be matched or not.

jkoenig134 (changes requested):

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
Magnus-Kuhn: an alias would turn a PIN "1234" into a number 1234, so this was my workaround

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
Magnus-Kuhn: Technically speaking, passwordIsPin is undefined if there is no password protection

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
jkoenig134: I see, I'm ok with that :)

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
jkoenig134: But shouldn't this be matched in the password protection itself?

Magnus-Kuhn: I don't quite understand this question - do you assume that users will think that querying `passwordIsPin` will automatically restrict to password protected objects?

jkoenig134: IMHO, someone should query `"passwordProtection": "!"` when they want to find unprotected templates. It's very unintuitive to query `"passwordProtection.passwordIsPin": "!"` when wanting unprotected PWs.

Magnus-Kuhn: They can use `"passwordProtection.password": "!"` for unprotected templates

jkoenig134: the query for pw exactly here please

jkoenig134: `$in: ["pw", undefined]` is not a correct db syntax for all of our dbs, you would have to write:

```
$or: [
  {[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => **t.passwordType)}`]: { $exists: false }},
  {[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => **t.passwordType)}`]: "pw"},
]
```

packages/runtime/test/transport/relationshipTemplates.test.ts
```diff
@@ -287,7 +291,23 @@ describe("RelationshipTemplates query", () => {
             .addStringSet("createdBy")
             .addStringSet("createdByDevice")
             .addNumberSet("maxNumberOfAllocations")
-            .addStringSet("forIdentity");
+            .addStringSet("forIdentity")
+            .addSingleCondition({
+                expectedResult: true,
+                key: "passwordProtection",
+                value: ""
+            })
```
jkoenig134, line 299: where's the negative matcher?

Magnus-Kuhn: I wanted to comment that this test fails - how is it supposed to look like?

jkoenig134: What's the error?

Magnus-Kuhn: me forgetting to build schemas again 🤦

PR #356, opened 2024-12-03
Delete obsolete LocalAccounts on App startup

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [ ] I self-reviewed the PR.


jkoenig134:
🔥

packages/app-runtime/src/AppRuntime.ts
```diff
@@ -302,6 +303,26 @@ export class AppRuntime extends Runtime<AppConfig> {
...
 
+    protected override async startAccounts(): Promise<void> {
+        const accounts = await this._multiAccountController.getAccounts();
+
+        for (const account of accounts) {
+            const session = await this.selectAccount(account.id.toString());
+            const syncResult = await session.transportServices.account.syncDatawallet();
+
+            if (syncResult.isSuccess) continue;
+            // TODO: can I check whether the error is a 400?
```
Milena-Czierlinski, line 314: @jkoenig134 What do you say?

packages/runtime/test/transport/account.test.ts
```diff
@@ -218,3 +219,25 @@ describe("Un-/RegisterPushNotificationToken", () => {
...
+
+        await sleep(10000);
+
+        const result = await sTransportServices.account.checkDeletionOfIdentity();
+        expect(result.isSuccess).toBe(true);
+        expect(result.value.isDeleted).toBe(true);
+        expect(result.value.deletionDate).toBe(identityDeletionProcess.gracePeriodEndsAt);
+    });
+
+    // TODO: test for deleted Identity
```
jkoenig134, line 242: when you do this test you can see if syncDatawallet does exactly what we think it should do ..

packages/transport/test/modules/account/IdentityController.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    test("check deletion of Identity that is not deleted", async function () {
+        const result = await account.identity.checkDeletionOfIdentity();
+        expect(result.value.isDeleted).toBe(false);
+        expect(result.value.deletionDate).toBeUndefined();
+    });
+
+    test("check deletion of Identity having IdentityDeletionProcess with expired grace period", async function () {
+        const identityDeletionProcess = await account.identityDeletionProcess.initiateIdentityDeletionProcess();
+
+        await sleep(10000);
```
jkoenig134, line 41: 10s is hard :S

Milena-Czierlinski: I can decrease the grace period. To be honest, I wanted to check the pipelines first if anything breaks. 😅

jkoenig134: WTB 0?

jkoenig134: Ah no, then we cannot cancel the idp

jkoenig134: I heard that there is some discussion to override the gracePeriod when creating an idp, would be awesome for this

packages/transport/src/modules/accounts/backbone/IdentityDeletionProcessClient.ts
```diff
@@ -1,12 +1,13 @@
...
 import { BackboneIdentityDeletionProcess } from "./BackboneIdentityDeletionProcess";
+import { BackbonePostIdentityDeletionProcessRequest } from "./BackbonePostIdentityDeletionProcess";
 
 export class IdentityDeletionProcessClient extends RESTClientAuthenticate {
     protected override _logDirective = RESTClientLogDirective.LogResponse;
 
-    public async initiateIdentityDeletionProcess(): Promise<ClientResult<BackboneIdentityDeletionProcess>> {
-        return await this.post<BackboneIdentityDeletionProcess>("/api/v1/Identities/Self/DeletionProcesses", {});
+    public async initiateIdentityDeletionProcess(value: BackbonePostIdentityDeletionProcessRequest): Promise<ClientResult<BackboneIdentityDeletionProcess>> {
+        return await this.post<BackboneIdentityDeletionProcess>("/api/v1/Identities/Self/DeletionProcesses", value, {});
```
britsta, line 10: @jkoenig134 Should the third input value (`{}`) be removed as you suggested in the other PR (https://github.com/nmshd/runtime/pull/358#discussion_r1875494942)?

jkoenig134: yes! We don't want to pass a config here.

Milena-Czierlinski: Already done, just not pushed yet. :)

packages/app-runtime/src/multiAccount/MultiAccountController.ts
Milena-Czierlinski, line 1: @jkoenig134 Can you please take a look if I can rename `deleteAccount` to `offboardDevice` and create (and expose) another function called `deleteAccount` or if that would lead to big complications in the App? I guess strictly speaking that would be a breaking change, but I`d hope we could handle it swiftly since we are currently the only consumers. The easier way would probably to give the new function (here `deleteAccount` a different name, but I felt this naming is more accurate.

jkoenig134: go for it!

jkoenig134: I don't see a real reason why we ever should do a breaking change for only app-runtime changes as we can easily handle these changes in the bridge / apps!

packages/app-runtime/src/multiAccount/AccountServices.ts
```diff
@@ -38,6 +38,10 @@ export class AccountServices {
         return LocalAccountMapper.toLocalAccountDTO(localAccount);
     }
 
+    public async offboardAccount(id: string): Promise<void> {
```
jkoenig134, line 41: maybe we should just call this offboardDevice :)

packages/app-runtime/src/multiAccount/AccountServices.ts
```diff
@@ -38,6 +38,10 @@ export class AccountServices {
...
+        await this.multiAccountController.offboardDevice(CoreId.from(id));
```
jkoenig134, line 42: or maybe this offboardAccount. I think that would be a better naming.

packages/app-runtime/src/multiAccount/AccountServices.ts
Milena-Czierlinski: That's how I initially named it. Then, I saw that all methods have `Account` in their name, also `onboardAccount`. But I would be fine renaming it. Should `onboardAccount` also be renamed?

jkoenig134: As mentioned in my other comment, I would prefer to name the MultiAccountController method offboardAccount too, then everything matches.

Milena-Czierlinski: Should the MultiAccountController method `onboardDevice` be renamed to `onboardAccount`, too?

jkoenig134: argh, I hate this. I think we should leave the name as it triggers a deviceOnboarding which creates a new account..

packages/app-runtime/test/runtime/Startup.test.ts
```diff
@@ -64,3 +64,46 @@ describe("Runtime Startup", function () {
...
+    });
+
+    afterAll(async () => await runtime.stop());
+
+    test("should not delete Account running startAccounts for an active Identity", async function () {
+        await runtime["startAccounts"]();
+        await expect(runtime.selectAccount(sessionA.account.id)).resolves.not.toThrow();
+    });
+
+    // TODO: Jest did not exit one second after the test run has completed.
```
Milena-Czierlinski, line 90: I'm a bit lost here. The reason for this must be in `await runtime["startAccounts"]();`. However in the following test, where the deletion job is executed, this doesn't occur. This was tested setting only for each test. The only difference in startAccounts between these tests it that here syncDatawallet is successful. @jkoenig134 Any idea?

jkoenig134: Hm, do we do a proper cleanup when removing the old accounts? Removing all references to controllers?

Milena-Czierlinski: I wouldn't vow for that. I only took what was in the previous deleteAccount method. Admittedly, I'm not too familiar with the wiring of Accounts. :D

packages/app-runtime/src/AppRuntime.ts
```diff
@@ -300,6 +300,42 @@ export class AppRuntime extends Runtime<AppConfig> {
         await this.lokiConnection.close().catch(logError);
     }
 
+    protected override async startAccounts(): Promise<void> {
```
jkoenig134, line 303: startAccounts should not be defined and called in the runtime, instead please make this private (and remove the override) and then add this:

```ts
public override async start(): Promise<void> {
    await super.start();

    await this.startAccounts();
}
```

packages/app-runtime/test/runtime/Startup.test.ts
jkoenig134: I think I did my best to clean up, but it's kind of hard to really get rid of everything..

packages/runtime/src/useCases/transport/account/CheckDeletionOfIdentity.ts
britsta, line 1: I'm not quite a fan of the name yet. The name `CheckDeletionOfIdentity` sounds a bit like monitoring whether the process of Identity deletion is being carried out correctly. I would find `CheckDeletionStatusOfIdentity` or `CheckIfIdentityIsDeleted` clearer.

Milena-Czierlinski: I purposely decided against something with deletionStatus or deletionInfo, since we have properties with this name that are not involved here. But I am fine with `CheckIfIdentityIsDeleted`.

jkoenig134 (dismissed):

jkoenig134 (approved):

PR #329, opened 2024-11-14
Add possibility to manually run deletion job on backbone

jkoenig134:
# Readiness checklist

-   [ ] I added/updated tests.
-   [x] I ensured that the PR title is good enough for the changelog.
-   [x] I labeled the PR.

PR #358, opened 2024-12-04
Deletion of Relationship is proposed after peer deletion

britsta:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

After PR https://github.com/nmshd/runtime/pull/329 has been merged, it is possible to write a test that the `status` of a Relationship changes to `"DeletionProposed"` after the peer to which the Relationship exists is deleted.


packages/transport/src/modules/accounts/backbone/IdentityDeletionProcessClient.ts
```diff
@@ -1,12 +1,13 @@
...
 import { BackboneIdentityDeletionProcess } from "./BackboneIdentityDeletionProcess";
+import { BackbonePostIdentityDeletionProcessRequest } from "./BackbonePostIdentityDeletionProcess";
 
 export class IdentityDeletionProcessClient extends RESTClientAuthenticate {
     protected override _logDirective = RESTClientLogDirective.LogResponse;
 
-    public async initiateIdentityDeletionProcess(): Promise<ClientResult<BackboneIdentityDeletionProcess>> {
-        return await this.post<BackboneIdentityDeletionProcess>("/api/v1/Identities/Self/DeletionProcesses", {});
+    public async initiateIdentityDeletionProcess(value: BackbonePostIdentityDeletionProcessRequest): Promise<ClientResult<BackboneIdentityDeletionProcess>> {
+        return await this.post<BackboneIdentityDeletionProcess>("/api/v1/Identities/Self/DeletionProcesses", value, {});
```
jkoenig134, line 10:
```suggestion
        return await this.post<BackboneIdentityDeletionProcess>("/api/v1/Identities/Self/DeletionProcesses", value);
```

jkoenig134 (changes requested):
Only one small thing - rest LGTM.

packages/transport/test/modules/relationships/decomposition.test.ts
```diff
@@ -118,3 +118,42 @@ describe("Data cleanup after relationship decomposition", function () {
...
+        const activeRelationship = await sender.relationships.getActiveRelationshipToIdentity(recipient.identity.address);
+        expect(activeRelationship?.status).toBe(RelationshipStatus.Active);
+
+        await recipient.identityDeletionProcess.initiateIdentityDeletionProcess(0);
+        await TestUtil.runDeletionJob();
+
+        await sender.syncEverything();
+        const deletionProposedRelationship = await sender.relationships.getRelationshipToIdentity(recipient.identity.address);
+        expect(deletionProposedRelationship?.status).toBe(RelationshipStatus.DeletionProposed);
+    });
```
Milena-Czierlinski, line 158:
```suggestion
    test("RelationshipStatus of a previously active Relationship should be set to DeletionProposed after the peer is deleted", async function () {
        const activeRelationship = await sender.relationships.getActiveRelationshipToIdentity(recipient.identity.address);
        expect(activeRelationship!.status).toBe(RelationshipStatus.Active);

        await recipient.identityDeletionProcess.initiateIdentityDeletionProcess(0);
        await TestUtil.runDeletionJob();

        await sender.syncEverything();
        const deletionProposedRelationship = await sender.relationships.getRelationshipToIdentity(recipient.identity.address);
        expect(deletionProposedRelationship!.status).toBe(RelationshipStatus.DeletionProposed);
    });
```

From my point of view the test name is a bit unambiguous about whether something is actually proposed to the Identity when in truth only the status is changed.

Milena-Czierlinski (commented):
Also just one small comment.

jkoenig134 (dismissed):

packages/transport/test/modules/relationships/decomposition.test.ts
Milena-Czierlinski: @britsta Please mind the (hidden) `!`.

britsta: Oops, I didn't see them at first. :D

Why is it necessary to insert two `!` at these locations?

Milena-Czierlinski: Technically it isn't required, the `?` serve the same purpose. However, in tests we do expect the objects to be defined and if not we want an error to be thrown immediately. So `!` are favored. In productive code, on the other hand, `?` should be used instead, as we don't want it to break on an error.

Milena-Czierlinski (approved):

jkoenig134 (approved):

packages/transport/test/modules/relationships/decomposition.test.ts
jkoenig134: actually in that case it doesn't matter if ! or ? is used as toBe would fail when it's undefined.

britsta: As I understand it, it's more of a high level thought behind it than actually having an impact on the failure or success of the test.

PR #363, opened 2024-12-10
Identity Recovery Kits

jkoenig134:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

## Added UseCases

- check for an existing Recovery Kit
- create a Recovery Kit. 
  
  This deletes an existing “backup device” and the corresponding Token, makes sure that the Datawallet is enabled, creates a device flagged as “backup device” and creates a password protected long-lived Token for the created Device.

## Testing

Testing the feature had to be split into two tests as the runtime is not able to do an onboarding. 

For this reason

- the Runtime tests the actual UseCases
- the AppRuntime tests 
  - that the recovery kit is actually usable for recovery
  - that the Device that is taken for onboarding correctly if flagged from `isBackupDevice: false` to `isBackupDevice: true` to make it possible to create another recovery kit


packages/transport/src/modules/devices/DevicesController.ts
```diff
@@ -40,15 +40,30 @@ export class DevicesController extends TransportController {
...
+        return device;
+    }
+
+    private async createDevice(params: SendDeviceParameters): Promise<Device> {
         const [signedChallenge, devicePwdDn] = await Promise.all([this.parent.challenges.createChallenge(ChallengeType.Identity), PasswordGenerator.createStrongPassword(45, 50)]);
 
         this.log.trace("Device Creation Challenge signed. Creating device on backbone...");
 
         const response = (
             await this.client.createDevice({
```
Magnus-Kuhn, line 63: Is the backbone notified once the device is onboarded and loses its backup device status?

packages/runtime/src/useCases/transport/identityRecoveryKits/CreateIdentityRecoveryKit.ts
```diff
@@ -0,0 +1,76 @@
...
+        const token = await this.tokenController.sendToken({
+            content: TokenContentDeviceSharedSecret.from({ sharedSecret }),
+            expiresAt: CoreDate.from("9999-12-31"),
+            ephemeral: false,
+            passwordProtection: PasswordProtectionCreationParameters.create(request.passwordProtection)
+        });
+
+        await this.accountController.syncDatawallet();
+
+        return Result.ok(TokenMapper.toTokenDTO(token, true));
```
Magnus-Kuhn, line 59: Here the token is given as ephemeral and above it's not - which is it?

packages/transport/src/modules/devices/local/Device.ts
```diff
@@ -32,6 +32,7 @@ export interface IDevice extends ICoreSynchronizable {
     initialPassword?: string;
     datawalletVersion?: number;
     isOffboarded?: boolean;
+    isBackupDevice: boolean;
```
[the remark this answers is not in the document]
jkoenig134: ehm, no? this should never be undefined..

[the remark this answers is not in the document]
Magnus-Kuhn: What's the difference between this and isDefault for attributes, passwordIsPin where we went for true/undefined instead of true/false?

packages/runtime/test/transport/identityRecoveryKits.test.ts
```diff
@@ -0,0 +1,121 @@
...
+const serviceProvider = new RuntimeServiceProvider();
+
+let services: TestRuntimeServices;
+let servicesWithDisabledDatawallet: TestRuntimeServices;
+
+beforeAll(async () => {
+    const runtimeServices = await serviceProvider.launch(1, { enableDatawallet: true });
+    services = runtimeServices[0];
+
+    const datawalletDisbledServices = await serviceProvider.launch(1, { enableDatawallet: false });
```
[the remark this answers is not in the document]
jkoenig134: nah.

packages/transport/src/modules/devices/DevicesController.ts
jkoenig134: indeed

packages/runtime/src/useCases/transport/identityRecoveryKits/CreateIdentityRecoveryKit.ts
jkoenig134: good catch! not ephemeral is it.

packages/transport/src/modules/devices/local/Device.ts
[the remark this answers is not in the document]
jkoenig134: a set value that the backbone decides on vs a value we decide how to work with

packages/runtime/src/extensibility/facades/transport/IdentityRecoveryKitFacade.ts
britsta, line 1: Most facades are named in their plural form. If nothing speaks against it, please use plural form `IdentityRecoveryKitsFacade` for the file name and class name for consistency. Please note that the test file is named `identityRecoveryKits.test.ts` already, for example.

packages/runtime/test/transport/identityRecoveryKits.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        expect(tokens[0].content.sharedSecret.id).toBe(backupDevice.id.toString());
+    });
+
+    test("should delete a recovery kit and its token when creating a consecutive recovery kit", async () => {
+        const firstToken = (await services.transport.identityRecoveryKits.createIdentityRecoveryKit({ profileName: "profileName", passwordProtection: { password: "aPassword" } }))
+            .value;
+        const firstBackupDevice = (await services.transport.devices.getDevices()).value.find((device) => device.isBackupDevice)!;
+
+        const response = await services.transport.identityRecoveryKits.createIdentityRecoveryKit({ profileName: "profileName", passwordProtection: { password: "aPassword" } });
+        expect(response).toBeSuccessful();
```
britsta, line 67: Variable name `secondToken` instead of `response` would help to understand the test even better so that it is clear that it is referred to the second Identity recovery kit.

packages/runtime/test/transport/identityRecoveryKits.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        const backupDevice = backupDevices[0];
+
+        const tokens = (await services.transport.tokens.getTokens({ ownerRestriction: OwnerRestriction.Own, query: { expiresAt: "^9999-12-31" } })).value;
+        expect(tokens).toHaveLength(1);
+
+        expect(tokens[0].content["@type"]).toBe("TokenContentDeviceSharedSecret");
+        expect(tokens[0].content.sharedSecret.id).toBe(backupDevice.id.toString());
+    });
+
+    test("should delete a recovery kit and its token when creating a consecutive recovery kit", async () => {
```
britsta, line 61: A test like this with a second device would be great in addition.

packages/runtime/src/useCases/transport/identityRecoveryKits/CreateIdentityRecoveryKit.ts
```diff
@@ -0,0 +1,72 @@
...
+        const devices = await this.devicesController.list();
+
+        const backupDevices = devices.filter((device) => device.isBackupDevice);
+        if (backupDevices.length > 0) await this.removeBackupDevices(backupDevices);
+
+        const newBackupDevice = await this.devicesController.sendDevice({ isAdmin: true, isBackupDevice: true, name: "Backup Device" });
+        const sharedSecret = await this.devicesController.getSharedSecret(newBackupDevice.id, request.profileName);
+        const token = await this.tokenController.sendToken({
+            content: TokenContentDeviceSharedSecret.from({ sharedSecret }),
+            expiresAt: CoreDate.from("9999-12-31"),
```
britsta, line 48: Is this supposed to be configurable in the future?

Also, will there be a use case for deleting an Identity recovery kit some day?

packages/runtime/src/useCases/transport/identityRecoveryKits/CreateIdentityRecoveryKit.ts
```diff
@@ -0,0 +1,72 @@
+import { Result } from "@js-soft/ts-utils";
+import { CoreDate } from "@nmshd/core-types";
+import { AccountController, Device, DevicesController, PasswordProtectionCreationParameters, TokenContentDeviceSharedSecret, TokenController } from "@nmshd/transport";
+import { Inject } from "@nmshd/typescript-ioc";
+import { TokenDTO } from "../../../types";
+import { RuntimeErrors, SchemaRepository, TokenAndTemplateCreationValidator, UseCase } from "../../common";
+import { TokenMapper } from "../tokens/TokenMapper";
+
+export interface CreateIdentityRecoveryKitRequest {
+    profileName: string;
```
britsta, line 10: Why is `profileName` not optional?

packages/app-runtime/test/runtime/Onboarding.test.ts
```diff
@@ -24,4 +31,30 @@ describe("Onboarding", function () {
...
+        await services1.transportServices.account.syncDatawallet();
+        const devices = (await services1.transportServices.devices.getDevices()).value;
+        const backupDevices = devices.filter((device) => device.isBackupDevice);
+        expect(backupDevices).toHaveLength(0);
+
+        const recoveryKitResponse2 = await services2.transportServices.identityRecoveryKits.createIdentityRecoveryKit({
+            profileName: "profileName",
+            passwordProtection: { password: "aPassword" }
+        });
+        expect(recoveryKitResponse2).toBeSuccessful();
```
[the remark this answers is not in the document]
jkoenig134: I had the choice to do this here - or duplicate EVERYTHING regarding this test, so I would really prefer to leave it as is.

packages/runtime/src/extensibility/facades/transport/IdentityRecoveryKitFacade.ts
jkoenig134: yeah, I started to migrate to s form but missed that, good catch

packages/runtime/test/transport/identityRecoveryKits.test.ts
jkoenig134: as mentioned in the PR description - this is out of scope in the runtime, also we don't have to test the datawallet sync again and again.

packages/runtime/src/useCases/transport/identityRecoveryKits/CreateIdentityRecoveryKit.ts
jkoenig134: nope

yes

packages/runtime/src/useCases/transport/identityRecoveryKits/CreateIdentityRecoveryKit.ts
jkoenig134: B/C the app heavily relies on that.

britsta: Is this a name that you can freely assign, or do you have to refer to an existing profile name? If this is the case, you would need somewhere to validate that the profile name exists. Would this then take place in the App because we don't know the existing profile names?

jkoenig134: profile names are an app runtime thing so we cannot do that

packages/runtime/test/transport/identityRecoveryKits.test.ts
britsta: `secondTokenResponse` would help as well, like you did with `getTokenResponse` and `getDeviceResponse` the lines below.

jkoenig134: Hm, I disagree. getTokenResponse and getDeviceResponse are assertions, response is ALWAYS the action itself in this test file.

britsta: Fine for me, just one detail I stumbled across.

packages/app-runtime/test/runtime/Onboarding.test.ts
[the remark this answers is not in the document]
jkoenig134: will do!

packages/app-runtime/test/runtime/Onboarding.test.ts
```diff
@@ -24,4 +31,30 @@ describe("Onboarding", function () {
...
+        const token = await runtime2.anonymousServices.tokens.loadPeerToken({ reference: recoveryKitResponse.value.truncatedReference, password: "aPassword" });
+        const deviceOnboardingDTO = DeviceMapper.toDeviceOnboardingInfoDTO(DeviceSharedSecret.from(token.value.content.sharedSecret));
+
+        const result = await runtime2.accountServices.onboardAccount(deviceOnboardingDTO);
+        expect(result.address!).toBe((await services1.transportServices.account.getIdentityInfo()).value.address);
+
+        const services2 = await runtime2.getServices(result.id);
+
+        await services1.transportServices.account.syncDatawallet();
+        const devices = (await services1.transportServices.devices.getDevices()).value;
```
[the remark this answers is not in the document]
britsta: IMO it would be interesting here because the test is about correct onboarding using a recovery kit but it's just a detail and can stay that way :)

britsta (dismissed):

Magnus-Kuhn (dismissed):
I also didn't have anything left before @britsta took off

packages/runtime/src/extensibility/facades/transport/IdentityRecoveryKitsFacade.ts
```diff
@@ -0,0 +1,19 @@
+import { Result } from "@js-soft/ts-utils";
+import { Inject } from "@nmshd/typescript-ioc";
+import { TokenDTO } from "../../../types";
+import { CreateIdentityRecoveryKitRequest, CreateIdentityRecoveryKitUseCase, DoesIdentityRecoveryKitExistResponse, DoesIdentityRecoveryKitExistUseCase } from "../../../useCases";
+
+export class IdentityRecoveryKitsFacade {
+    public constructor(
+        @Inject private readonly createIdentityRecoveryKitUseCase: CreateIdentityRecoveryKitUseCase,
+        @Inject private readonly existsIdentityRecoveryKitUseCase: DoesIdentityRecoveryKitExistUseCase
```
britsta, line 9: Missing change: `doesIdentityRecoveryKitExistUseCase` instead of `existsIdentityRecoveryKitUseCase`.

Milena-Czierlinski, line 9: Inconsistency.

Also, I'm still not completely happy with the new naming. I'd prefer the name of the use case to **describe** what the use case does instead of **asking a question**. That's why I suggested something starting with `check`, e.g. `checkForIdentityRecoveryKit`, `checkForExistingIdentityRecoveryKit`, `checkIdentityRecoveryKitExistence`, `checkIfIdentityRecoveryKitExists` ...

packages/runtime/test/transport/identityRecoveryKits.test.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    const datawalletDisabledServices = await serviceProvider.launch(1, { enableDatawallet: false });
+    servicesWithDisabledDatawallet = datawalletDisabledServices[0];
+}, 30000);
+
+afterAll(async () => await serviceProvider.stop());
+
+afterEach(async () => {
+    const devicesController = services.transport.identityRecoveryKits["createIdentityRecoveryKitUseCase"]["devicesController"];
+    const tokenController = services.transport.identityRecoveryKits["createIdentityRecoveryKitUseCase"]["tokenController"];
```
Milena-Czierlinski, line 21: Do the controllers change? Otherwise, this doesn't need to be done again after each test.

packages/runtime/test/transport/identityRecoveryKits.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        });
+        expect(response).toBeSuccessful();
+
+        const devices = (await services.transport.devices.getDevices()).value;
+        const backupDevices = devices.filter((device) => device.isBackupDevice);
+        expect(backupDevices).toHaveLength(1);
+
+        const backupDevice = backupDevices[0];
+
+        const tokens = (await services.transport.tokens.getTokens({ ownerRestriction: OwnerRestriction.Own, query: { expiresAt: "^9999-12-31" } })).value;
```
Milena-Czierlinski, line 54: Is the `^` in the expiry date there on purpose?

packages/transport/src/modules/devices/backbone/BackbonePostDevices.ts
```diff
@@ -3,11 +3,13 @@ import { IChallengeSignedSerialized } from "../../challenges/data/ChallengeSigne
 export interface BackbonePostDevicesRequest {
     devicePassword: string;
     signedChallenge: IChallengeSignedSerialized;
+    isBackupDevice?: boolean;
```
Milena-Czierlinski, line 6: At first glance I was confused why here `isBackupDevice` may be undefined. So I took a deeper look and saw that `sendDevice` is called by `CreateIdentityRecoveryKit` and `CreateDevice`. The former sets `isBackupDevice` to true while the latter doesn't set it. We could set it to false in the use case or have `sendDevice` set it to false if it is undefined, instead of letting the Backbone do that. Is there a coding guideline that suggests if such things should be done at a higher or lower level? If we do this at a higher level, it might make our code less complicated in some places.

packages/runtime/test/transport/identityRecoveryKits.test.ts
jkoenig134: idk

packages/runtime/test/transport/identityRecoveryKits.test.ts
jkoenig134: yes, regex matcher

packages/transport/src/modules/devices/backbone/BackbonePostDevices.ts
jkoenig134: Is this so important? The backbone has a default, we take that default (false) when not sending anything.

Milena-Czierlinski: I don't consider it indispensable for the implementation here.

> Is there a coding guideline that suggests if such things should be done at a higher or lower level?

 This was meant as a general question. If there is an unambiguous suggestion, your question about the importance should probably be answered by the significance of code style we have in our repos in general.

jkoenig134: IMO `Backbone{Post|GET|...}...Request`s should be **as close** to what the backbone expects as possible.

Best example are the responses - we work with undefined for not set fields, the backbone sets null. If we only look on our codestyle everything blows up.

Theses Requests and Responses are defined by the backbone, so we do not really decide - also I don't see a really good reason to have this required.

Milena-Czierlinski: I'm not completely satisfied by "outcome" of this discussion, but I get the feeling that you do not wish to discuss this further. And as already stated, since I don't consider it indispensable to change something here, I'm okay with putting it to rest for now.

Milena-Czierlinski (approved):

PR #367, opened 2024-12-12
Prevent the creation of duplicated RepositoryAttributes

sebbi08:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [ ] I self-reviewed the PR.

### DIFF
--- a/packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
+++ b/packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
@@ -4,7 +4,7 @@ import { AttributeValues } from "@nmshd/content";
 import { AccountController } from "@nmshd/transport";
 import { Inject } from "@nmshd/typescript-ioc";
 import { LocalAttributeDTO } from "../../../types";
-import { ISO8601DateTimeString, SchemaRepository, SchemaValidator, UseCase } from "../../common";
+import { flattenObject, ISO8601DateTimeString, RuntimeErrors, SchemaRepository, SchemaValidator, UseCase } from "../../common";
 import { AttributeMapper } from "./AttributeMapper";
 
 export interface CreateRepositoryAttributeRequest {
@@ -39,6 +39,19 @@ export class CreateRepositoryAttributeUseCase extends UseCase<CreateRepositoryAt
                 ...request.content
             }
         });
+        // ignore (tags,validFrom,validTo) when checking for duplicates
+        const query = flattenObject({
+            content: {
+                "@type": "IdentityAttribute",
+                owner: this.accountController.identity.address.toString(),
+                value: request.content.value
+            }
+        });
+        const existingAttribute = await this.attributeController.getLocalAttributes(query);
+        if (existingAttribute.length > 0) {
+            return Result.fail(RuntimeErrors.attributes.cannotCreateDuplicateRepositoryAttribute(existingAttribute[0].id));
+        }
+
         const createdLocalAttribute = await this.attributeController.createRepositoryAttribute(params);
 
         await this.accountController.syncDatawallet();
--- a/packages/runtime/test/consumption/attributes.test.ts
+++ b/packages/runtime/test/consumption/attributes.test.ts
@@ -14,6 +14,7 @@ import {
 } from "@nmshd/content";
 import { CoreDate, CoreId } from "@nmshd/core-types";
 import { CoreIdHelper } from "@nmshd/transport";
+import { randomUUID } from "crypto";
 import {
     AttributeCreatedEvent,
     ChangeDefaultRepositoryAttributeUseCase,
@@ -52,6 +53,7 @@ import {
     RuntimeServiceProvider,
     TestRuntimeServices,
     acceptIncomingShareAttributeRequest,
+    cleanupAttributes,
     establishRelationship,
     exchangeAndAcceptRequestByMessage,
     executeFullCreateAndShareRelationshipAttributeFlow,
@@ -96,30 +98,18 @@ beforeAll(async () => {
 }, 30000);
 afterAll(async () => await runtimeServiceProvider.stop());
 
-beforeEach(() => {
+beforeEach(async () => {
     services1.eventBus.reset();
     services2.eventBus.reset();
     services3.eventBus.reset();
+    await cleanupAttributes(services1, services2, services3, appService);
 });
-
-async function cleanupAttributes() {
-    await Promise.all(
-        [services1, services2, services3, appService].map(async (services) => {
-            const servicesAttributeController = services.consumption.attributes["getAttributeUseCase"]["attributeController"];
-
-            const servicesAttributesResult = await services.consumption.attributes.getAttributes({});
-            for (const attribute of servicesAttributesResult.value) {
-                await servicesAttributeController.deleteAttributeUnsafe(CoreId.from(attribute.id));
-            }
-        })
-    );
-}
 
 describe("get attribute(s)", () => {
     let relationshipAttributeId: string;
     let identityAttributeIds: string[];
     let appAttributeIds: string[];
-    beforeAll(async function () {
+    beforeEach(async function () {
         const senderRequests: CreateRepositoryAttributeRequest[] = [
             {
                 content: {
@@ -172,10 +162,6 @@ describe("get attribute(s)", () => {
             const appAttribute = (await appService.consumption.attributes.createRepositoryAttribute(request)).value;
             appAttributeIds.push(appAttribute.id);
         }
-    });
-
-    afterAll(async function () {
-        await cleanupAttributes();
     });
 
     describe(GetAttributeUseCase.name, () => {
@@ -278,7 +264,7 @@ describe("attribute queries", () => {
     let repositoryAttribute: LocalAttributeDTO;
     let ownSharedRelationshipAttribute: LocalAttributeDTO;
 
-    beforeAll(async function () {
+    beforeEach(async function () {
         const createRepositoryAttributeRequest: CreateRepositoryAttributeRequest = {
             content: {
                 value: {
@@ -300,10 +286,6 @@ describe("attribute queries", () => {
                 confidentiality: RelationshipAttributeConfidentiality.Protected
             }
         });
-    });
-
-    afterAll(async function () {
-        await cleanupAttributes();
     });
 
     describe(ExecuteIdentityAttributeQueryUseCase.name, () => {
@@ -365,7 +347,7 @@ describe("get repository, own shared and peer shared attributes", () => {
 
     let services1SharedTechnicalRelationshipAttribute: LocalAttributeDTO;
 
-    beforeAll(async function () {
+    beforeEach(async function () {
         // unshared succeeded repository attribute
         services1RepoSurnameV0 = (
             await services1.consumption.attributes.createRepositoryAttribute({
@@ -493,10 +475,6 @@ describe("get repository, own shared and peer shared attributes", () => {
         });
     });
 
-    afterAll(async function () {
-        await cleanupAttributes();
-    });
-
     describe(GetRepositoryAttributesUseCase.name, () => {
         test("get only latest version of repository attributes", async () => {
             const result = await services1.consumption.attributes.getRepositoryAttributes({});
@@ -600,7 +578,7 @@ describe("get repository, own shared and peer shared attributes", () => {
         let allReceivedAttributes: LocalAttributeDTO[];
         let onlyLatestReceivedAttributes: LocalAttributeDTO[];
         let notTechnicalReceivedAttributes: LocalAttributeDTO[];
-        beforeAll(async function () {
+        beforeEach(async function () {
             const services1SharedAttributeIds = [
                 services1SharedGivenNameV0,
                 services1SharedGivenNameV1,
@@ -661,8 +639,9 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
                 tags: ["tag1", "tag2"]
             }
         };
+
         const result = await services1.consumption.attributes.createRepositoryAttribute(request);
-        expect(result.isError).toBe(false);
+        expect(result.value).toBeDefined();
         const attribute = result.value;
         expect(attribute.content).toMatchObject(request.content);
         await services1.eventBus.waitForEvent(AttributeCreatedEvent, (e) => e.data.id === attribute.id);
@@ -740,14 +719,40 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
             content: {
                 value: {
                     "@type": "JobTitle",
-                    value: "A job title"
+                    value: "A unique job title"
+                }
+            }
+        };
+        const request2: CreateRepositoryAttributeRequest = {
+            content: {
+                value: {
+                    "@type": "JobTitle",
+                    value: "A more unique job title"
                 }
             }
         };
         await appService.consumption.attributes.createRepositoryAttribute(request);
-        const result = await appService.consumption.attributes.createRepositoryAttribute(request);
+        const result = await appService.consumption.attributes.createRepositoryAttribute(request2);
         const attribute = result.value;
         expect(attribute.isDefault).toBeUndefined();
+    });
+
+    test("should not create a duplicate RepositoryAttribute", async () => {
+        const request: CreateRepositoryAttributeRequest = {
+            content: {
+                value: {
+                    "@type": "GivenName",
+                    value: `Petrus Pan${randomUUID()}`
+                },
+                tags: ["tag1", "tag2"]
+            }
+        };
+
+        const result = await services1.consumption.attributes.createRepositoryAttribute(request);
+        expect(result.value).toBeDefined();
+
+        const result2 = await services1.consumption.attributes.createRepositoryAttribute(request);
+        expect(result2).toBeAnError(`The Attribute is a copy of the attribute ${result.value.id.toString()}.`, "error.runtime.attributes.cannotCreateDuplicateRepositoryAttribute");
     });
 });
 
@@ -1227,7 +1232,7 @@ describe(NotifyPeerAboutRepositoryAttributeSuccessionUseCase.name, () => {
                 content: {
                     value: {
                         "@type": "GivenName",
-                        value: "Petra Pan"
+                        value: "Petrus Pan"
                     }
                 }
             })
@@ -1397,10 +1402,6 @@ describe(SucceedRelationshipAttributeAndNotifyPeerUseCase.name, () => {
 });
 
 describe(ChangeDefaultRepositoryAttributeUseCase.name, () => {
-    beforeAll(async () => {
-        await cleanupAttributes();
-    });
-
     test("should change default RepositoryAttribute", async () => {
         const defaultAttribute = (
             await appService.consumption.attributes.createRepositoryAttribute({
@@ -1760,7 +1761,7 @@ describe("Get (shared) versions of attribute", () => {
     });
 
     describe(GetSharedVersionsOfAttributeUseCase.name, () => {
-        beforeAll(async () => {
+        beforeEach(async () => {
             await setUpIdentityAttributeVersions();
         });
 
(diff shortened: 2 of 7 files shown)


Magnus-Kuhn (changes requested):
I ran a test that you can still create duplicate repository attributes as children of distinct complex attributes. (e. g. Birthdate 3.5. vs. BirthDate 4.5.)

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
```diff
@@ -39,6 +39,18 @@ export class CreateRepositoryAttributeUseCase extends UseCase<CreateRepositoryAt
                 ...request.content
             }
         });
+        const queryWithoutOptionalProperties = flattenObject({
```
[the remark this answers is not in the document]
Magnus-Kuhn: A more serious issue: If an attribute exists with an optional property set - and you try to create an attribute without that optional property (e. g. leaving out the state of an address), then it is blocked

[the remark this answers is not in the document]
sebbi08: Hi I rename this to queryWithoutOptionalRequestProperties this only refers to optional properties of the Request : tags, validFrom, validTo for the attribute itself all combinations are still possible only direct duplicates are blocked

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -1016,7 +1017,7 @@ describe("RelationshipDecomposition", () => {
             content: {
                 value: {
                     "@type": "GivenName",
-                    value: "Own name"
+                    value: `Own name${randomUUID()}`
```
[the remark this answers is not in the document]
sebbi08: Hi yea but this is inside a function so adding a number here would not allow getting a conclusion where the n attribute came from.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -1227,7 +1234,7 @@ describe(NotifyPeerAboutRepositoryAttributeSuccessionUseCase.name, () => {
                 content: {
                     value: {
                         "@type": "GivenName",
-                        value: "Petra Pan"
+                        value: "aGivenName2"
```
Magnus-Kuhn, line 1237:
```suggestion
                        value: "aGivenName"
```

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -1892,7 +1895,7 @@ describe("DeleteAttributeUseCases", () => {
             content: {
                 value: {
                     "@type": "GivenName",
-                    value: "Petra Pan"
+                    value: "Petro Pan"
```
Magnus-Kuhn, line 1898:
```suggestion
                    value: "aGivenName"
```

Magnus-Kuhn (commented):
I'm trying the vscode "convert changes to PR suggestions" feature

packages/runtime/test/transport/relationships.test.ts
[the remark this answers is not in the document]
Magnus-Kuhn: This can also be turned into `aGivenName`

packages/runtime/test/consumption/attributes.test.ts
jkoenig134: please use givenName and anotherGivenName - we don't use ANY of those fakish names.

Magnus-Kuhn (commented):
I'm not happy with that vscode feature, it created multiple reviews and left some things pending

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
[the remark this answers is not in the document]
Magnus-Kuhn: Still not fond of that because it requires understanding that there are optional properties - also the query content is named in the code, there's no further need to explain what it is. What the query is for is more useful IMO.

britsta (changes requested):
The way I see it, the creation of RepositoryAttribute duplicates must be prevented in other places as well, because new RepositoryAttributes are sometimes also created when responding to Requests.

However, we must be careful if we add validation not only at use case level but also at a deeper level: For example, when creating the two birth dates "1.1.2000" and "2.2.2000", two child RepositoryAttributes are created which have “2000” as `content.value`. We can therefore not assume the non-existence of RepositoryAttribute duplicates even after merging this PR. (However, please keep in mind that the concept of complex Attributes and their child Attributes is currently being revised so that we should not spend too much time on the handling of the child Attributes.)

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -996,6 +996,8 @@ describe("RelationshipDecomposition", () => {
         );
     });
 
+    let nameCounter = 0;
```
britsta, line 999: Instead of the `nameCounter` I would prefer an approach in which two random character strings are generated for use as GivenName below.

Magnus-Kuhn: I suggested something like this instead of random characters such that if something fails, you have an idea which one failed

sebbi08:
> The way I see it, the creation of RepositoryAttribute duplicates must be prevented in other places as well, because new RepositoryAttributes are sometimes also created when responding to Requests. (**Edit**: I understand if it's not part of the scope of the PR if it's not seen as manually creating RepositoryAttributes, but just wanted to mention it so it's not forgotten.)
> 
> However, we must be careful if we add validation not only at use case level but also at a deeper level: For example, when creating the two birth dates "1.1.2000" and "2.2.2000", two child RepositoryAttributes are created which have “2000” as `content.value`. We can therefore not assume the non-existence of RepositoryAttribute duplicates even after merging this PR. (However, please keep in mind that the concept of complex Attributes and their child Attributes is currently being revised so that we should not spend too much time on the handling of the child Attributes.)

Hi yea totally with you. For the attributes created by requests, we already have another ticket. And we should create one for checking this again after the Complex attribute handling is done

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -1016,7 +1017,7 @@ describe("RelationshipDecomposition", () => {
             content: {
                 value: {
                     "@type": "GivenName",
-                    value: "Own name"
+                    value: randomUUID()
```
jkoenig134, line 1020: that's not how we create random strings. Also please put this to a variable called ownName to not break the wording here completely.

Also the value below was the same before, now you use a different one for both?

sebbi08: Yea the main problem was that they were the same because that's what was disabled and as this function was called multiple times so we needed random values here

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -1012,20 +1012,22 @@ describe("RelationshipDecomposition", () => {
 
         await sendAndReceiveNotification(services1.transport, services2.transport, services2.consumption);
 
+        const randomName1 = Math.random().toString(36).substring(7);
```
jkoenig134, line 1015: this is also not how we generate random strings :D please use transport's Random

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
```diff
@@ -39,6 +39,20 @@ export class CreateRepositoryAttributeUseCase extends UseCase<CreateRepositoryAt
...
+
+        const queryForRepositoryAttributeDuplicates = flattenObject({
+            content: {
+                "@type": "IdentityAttribute",
+                owner: this.accountController.identity.address.toString(),
+                value: request.content.value
+            }
+        });
+
+        const existingRepositoryAttributes = await this.attributeController.getLocalAttributes(queryForRepositoryAttributeDuplicates);
```
jkoenig134, line 51: I think this introduces a major bug because you don't filter out attributes that have been succeeded. @Milena-Czierlinski & @britsta do you also see a problem with that?

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
```diff
@@ -39,6 +39,20 @@ export class CreateRepositoryAttributeUseCase extends UseCase<CreateRepositoryAt
...
             }
         });
+
+        const queryForRepositoryAttributeDuplicates = flattenObject({
+            content: {
+                "@type": "IdentityAttribute",
+                owner: this.accountController.identity.address.toString(),
+                value: request.content.value
+            }
+        });
```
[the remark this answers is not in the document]
sebbi08: Because the value that comes in is not flat and therefor we anyway need to call "flattenObject" and I find this easier readable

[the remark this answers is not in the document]
jkoenig134: Oh, I forgot about the value that is not flat - makes sense!

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -740,15 +718,44 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
...
         const attribute = result.value;
         expect(attribute.isDefault).toBeUndefined();
     });
+
+    test("should not create a duplicate RepositoryAttribute", async () => {
+        const request: CreateRepositoryAttributeRequest = {
+            content: {
+                value: {
+                    "@type": "GivenName",
+                    value: `aGivenName3`
```
Magnus-Kuhn, line 744:
```suggestion
                    value: "aGivenName3"
```

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
sebbi08: Yea, I think you are right, I added a query to only find attributes without a successor.

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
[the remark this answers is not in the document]
Magnus-Kuhn: > A more serious issue: If an attribute exists with an optional property set - and you try to create an attribute without that optional property (e. g. leaving out the state of an address), then it is blocked

Should I push a test for that?

packages/runtime/test/consumption/attributes.test.ts
Magnus-Kuhn: Sorry, initially forgot to remove the `3` in the suggestion

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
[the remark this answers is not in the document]
jkoenig134: yes!

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
britsta: Please also add a test that it is also possible to create a new RepositoryAttribute with a certain `content.value` if the previous RepositoryAttribute with this `content.value` has already been succeeded and is therefore no longer up-to-date.

packages/runtime/test/consumption/attributes.test.ts
britsta, line 1: > Hi I rename this to queryWithoutOptionalRequestProperties this only refers to optional properties of the Request : tags, validFrom, validTo for the attribute itself all combinations are still possible only direct duplicates are blocked

This means that it should be possible to create another RepositoryAttribute with a certain `content.value` if there is already a RepositoryAttribute with the same `content.value` but different `tags` (or different validity periods), for example? If so, it would be great if you could add a test for this! 🎄

sebbi08: I ment the exact opposite so tags and so don't count as duplicate but I will add a test for that.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -740,15 +718,205 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
...
+        };
+
+        const result4 = await services1.consumption.attributes.createRepositoryAttribute(request4);
+        expect(result4).toBeAnError(
+            `The RepositoryAttribute cannot be created because it has the same content.value as the already existing RepositoryAttribute with id '${result.value.id.toString()}'.`,
+            "error.runtime.attributes.cannotCreateDuplicateRepositoryAttribute"
+        );
+    });
+
+    test("should not a duplicate RepositoryAttribute even if the Tags/validFrom/validTo are the same", async () => {
```
Magnus-Kuhn, line 888: Something went wrong with the name

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
```diff
@@ -39,6 +40,25 @@ export class CreateRepositoryAttributeUseCase extends UseCase<CreateRepositoryAt
...
+
+        queryForRepositoryAttributeDuplicates["succeededBy"] = { $exists: false };
+
+        const existingRepositoryAttributes = await this.attributeController.getLocalAttributes(queryForRepositoryAttributeDuplicates);
+
+        const filterForExactValueContent = existingRepositoryAttributes.some((duplicate) => _.isEqual(duplicate.content.value.toJSON(), request.content.value));
+
+        if (filterForExactValueContent && existingRepositoryAttributes.length > 0) {
```
Magnus-Kuhn, line 58: I think the length check is not needed anymore

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
```diff
@@ -39,6 +40,25 @@ export class CreateRepositoryAttributeUseCase extends UseCase<CreateRepositoryAt
...
+                owner: this.accountController.identity.address.toString(),
+                value: request.content.value
+            }
+        });
+
+        queryForRepositoryAttributeDuplicates["succeededBy"] = { $exists: false };
+
+        const existingRepositoryAttributes = await this.attributeController.getLocalAttributes(queryForRepositoryAttributeDuplicates);
+
+        const filterForExactValueContent = existingRepositoryAttributes.some((duplicate) => _.isEqual(duplicate.content.value.toJSON(), request.content.value));
```
Magnus-Kuhn, line 56: I'm not too fond of the name which I would consider to be more appropriate for an actual filtering function rather than its result.
What about `exactMatchExists` ?

sebbi08, line 56: @jkoenig134 do you know if this can be achieved with and DB query instead of this Manuel filtering?

packages/runtime/test/consumption/attributes.test.ts
britsta: Mhh I wonder if that makes sense. We can probably argue for it as long as it's about manual RepositoryAttribute creation via the use case. When it comes to responding to Requests, we would have to allow the existence of RepositoryAttribute duplicates if they have different `tags`, for example, because it is not within our control which Requests the different providers will send. But that is, as you rightly said, in the scope of another PR.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -816,6 +816,107 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
...
+                validTo
+            }
+        };
+
+        const result3 = await services1.consumption.attributes.createRepositoryAttribute(request3);
+        expect(result3).toBeAnError(
+            `The RepositoryAttribute cannot be created because it has the same content.value as the already existing RepositoryAttribute with id '${result.value.id.toString()}'.`,
+            "error.runtime.attributes.cannotCreateDuplicateRepositoryAttribute"
+        );
+        const request4: CreateRepositoryAttributeRequest = {
```
britsta, line 870: Please insert empty line before defining `request4`.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -816,6 +816,107 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
...
+                validFrom
+            }
+        };
+
+        const result2 = await services1.consumption.attributes.createRepositoryAttribute(request2);
+        expect(result2).toBeAnError(
+            `The RepositoryAttribute cannot be created because it has the same content.value as the already existing RepositoryAttribute with id '${result.value.id.toString()}'.`,
+            "error.runtime.attributes.cannotCreateDuplicateRepositoryAttribute"
+        );
+        const request3: CreateRepositoryAttributeRequest = {
```
britsta, line 854: Please insert empty line before defining `request3`.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -740,15 +718,205 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
...
+        };
+
+        const result = await services1.consumption.attributes.createRepositoryAttribute(request);
+        expect(result).toBeSuccessful();
+
+        const result2 = await services1.consumption.attributes.createRepositoryAttribute(request2);
+        expect(result2).toBeSuccessful();
+    });
+
+    test("should not create a duplicate RepositoryAttribute even if the Tags/validFrom/validTo are different", async () => {
```
britsta, line 820: Please write `tags` lowercase.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -740,15 +718,205 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
...
+        expect(result).toBeSuccessful();
+
+        const result2 = await services1.consumption.attributes.createRepositoryAttribute(request);
+        expect(result2).toBeAnError(
+            `The RepositoryAttribute cannot be created because it has the same content.value as the already existing RepositoryAttribute with id '${result.value.id.toString()}'.`,
+            "error.runtime.attributes.cannotCreateDuplicateRepositoryAttribute"
+        );
+    });
+
+    test("should not prevent the creation when the duplicate got succeeded", async () => {
```
britsta, line 760: Please write more detailed "RepositoryAttribute duplicate" instead of "duplicate".

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -740,15 +718,207 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
...
+        };
+
+        const result4 = await services1.consumption.attributes.createRepositoryAttribute(request4);
+        expect(result4).toBeAnError(
+            `The RepositoryAttribute cannot be created because it has the same content.value as the already existing RepositoryAttribute with id '${result.value.id.toString()}'.`,
+            "error.runtime.attributes.cannotCreateDuplicateRepositoryAttribute"
+        );
+    });
+
+    test("should create a RepositoryAttribute even if the Tags/validFrom/validTo are duplicates", async () => {
```
Magnus-Kuhn, line 890: another `tags` lowercase instance

Magnus-Kuhn (commented):
I resolved some other comments, the remainder basically looks like cosmetics only

Magnus-Kuhn (dismissed):

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -740,15 +718,207 @@ describe(CreateRepositoryAttributeUseCase.name, () => {
...
+                value: {
+                    "@type": "GivenName",
+                    value: "aGivenName2"
+                },
+                tags: ["tag1", "tag2"],
+                validFrom,
+                validTo
+            }
+        };
+        const result2 = await services1.consumption.attributes.createRepositoryAttribute(request2);
```
britsta, line 919: Add empty line before defining `result2`.

packages/runtime/test/consumption/attributes.test.ts
britsta: @sebbi08 What do you say? 🙂

britsta (approved):

packages/runtime/test/consumption/attributes.test.ts
sebbi08: We definitely need to think about that in the next PR, i would even say that we should prevent the creation from external event if the tags are different, but let's take this to another discussion.

PR #369, opened 2024-12-16
Request can be created and sent via a message with an expiration date that has passed

sebbi08:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
+++ b/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
@@ -324,6 +324,10 @@ class Requests {
         return new CoreError("error.consumption.requests.peerIsInDeletion", message);
     }
 
+    public violatedKeyUniquenessOfRelationshipAttributes(message: string) {
+        return new CoreError("error.consumption.requests.violatedKeyUniquenessOfRelationshipAttributes", message);
+    }
+
     public inheritedFromItem(message: string) {
         return new ApplicationError("error.consumption.requests.validation.inheritedFromItem", message);
     }
--- a/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
+++ b/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
@@ -331,6 +331,7 @@ class Requests {
     public cannotShareRequestWithYourself() {
         return new CoreError("error.consumption.requests.cannotShareRequestWithYourself", "You cannot share a Request with yourself.");
     }
+
     public cannotCreateRequestWithExpirationDateInPast() {
         return new CoreError("error.consumption.requests.cannotCreateRequestWithExpirationDateInPast", "You cannot create a Request with an expiration date in the past.");
     }
--- a/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
+++ b/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
@@ -333,7 +333,7 @@ class Requests {
     }
 
     public cannotCreateRequestWithExpirationDateInPast() {
-        return new CoreError("error.consumption.requests.cannotCreateRequestWithExpirationDateInPast", "You cannot create a Request with an expiration date in the past.");
+        return new CoreError("error.consumption.requests.cannotCreateRequestWithExpirationDateInPast", "You cannot create a Request with an expiration date that is in the past.");
     }
 
     private static readonly _decideValidation = class {
--- a/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
+++ b/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
@@ -331,6 +331,9 @@ class Requests {
     public cannotShareRequestWithYourself() {
         return new CoreError("error.consumption.requests.cannotShareRequestWithYourself", "You cannot share a Request with yourself.");
     }
+    public cannotCreateRequestWithExpirationDateInPast() {
+        return new CoreError("error.consumption.requests.cannotCreateRequestWithExpirationDateInPast", "You cannot create a Request with an expiration date in the past.");
+    }
 
     private static readonly _decideValidation = class {
         public invalidNumberOfItems(message: string) {
--- a/packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
+++ b/packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
@@ -711,7 +711,7 @@ describe("OutgoingRequestsController", function () {
                     items: [TestRequestItem.from({ mustBeAccepted: false })]
                 }
             });
-            await sleep(100);
+            await sleep(150);
 
             await When.iCompleteTheOutgoingRequestWith({ responseSourceObject: incomingMessage });
             await Then.theRequestMovesToStatus(LocalRequestStatus.Completed);
@@ -725,7 +725,7 @@ describe("OutgoingRequestsController", function () {
                     items: [TestRequestItem.from({ mustBeAccepted: false })]
                 }
             });
-            await sleep(100);
+            await sleep(150);
 
             const incomingMessage = TestObjectFactory.createIncomingIMessage(context.currentIdentity);
             await When.iTryToCompleteTheOutgoingRequestWith({ responseSourceObject: incomingMessage });
@@ -752,7 +752,7 @@ describe("OutgoingRequestsController", function () {
                     items: [TestRequestItem.from({ mustBeAccepted: false })]
                 }
             });
-            await sleep(100);
+            await sleep(150);
 
             await When.iGetTheOutgoingRequest();
             await Then.theRequestIsInStatus(LocalRequestStatus.Expired);
@@ -832,7 +832,7 @@ describe("OutgoingRequestsController", function () {
                     items: [TestRequestItem.from({ mustBeAccepted: false })]
                 }
             });
-            await sleep(100);
+            await sleep(150);
 
             await When.iGetOutgoingRequestsWithTheQuery({ id: outgoingRequest.id.toString() });
             await Then.theOnlyReturnedRequestIsInStatus(LocalRequestStatus.Expired);
--- a/packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
+++ b/packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
@@ -224,6 +224,24 @@ describe("OutgoingRequestsController", function () {
                 message: "You cannot share a Request with yourself."
             });
         });
+
+        test("returns a validation result that contains an error for requests th at are expired", async function () {
+            const validationResult = await When.iCallCanCreateForAnOutgoingRequest({
+                content: {
+                    expiresAt: CoreDate.utc().subtract({ days: 1 }),
+                    items: [
+                        TestRequestItem.from({
+                            mustBeAccepted: false
+                        })
+                    ]
+                }
+            });
+
+            expect(validationResult).errorValidationResult({
+                code: "error.consumption.requests.cannotCreateRequestWithExpirationDateInPast",
+                message: "You cannot create a Request with an expiration date in the past."
+            });
+        });
     });
 
     describe("CanCreate (on terminated relationship)", function () {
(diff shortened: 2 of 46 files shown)


packages/consumption/src/consumption/ConsumptionCoreErrors.ts
```diff
@@ -331,6 +331,9 @@ class Requests {
     public cannotShareRequestWithYourself() {
         return new CoreError("error.consumption.requests.cannotShareRequestWithYourself", "You cannot share a Request with yourself.");
     }
+    public cannotCreateRequestWithExpirationDateInPast() {
```
jkoenig134, line 334: newline

packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
```diff
@@ -224,6 +224,24 @@ describe("OutgoingRequestsController", function () {
                 message: "You cannot share a Request with yourself."
             });
         });
+
+        test("returns a validation result that contains an error for requests th at are expired", async function () {
```
jkoenig134, line 228:
```suggestion
        test("returns a validation result that contains an error for requests that are expired", async function () {
```

packages/consumption/src/consumption/ConsumptionCoreErrors.ts
sebbi08: woops already did that but forgot to push

jkoenig134 (changes requested):
one small thing

jkoenig134 (approved):
LGTM

packages/consumption/src/consumption/ConsumptionCoreErrors.ts
```diff
@@ -332,6 +332,10 @@ class Requests {
         return new CoreError("error.consumption.requests.cannotShareRequestWithYourself", "You cannot share a Request with yourself.");
     }
 
+    public cannotCreateRequestWithExpirationDateInPast() {
+        return new CoreError("error.consumption.requests.cannotCreateRequestWithExpirationDateInPast", "You cannot create a Request with an expiration date in the past.");
```
Milena-Czierlinski, line 336:
```suggestion
        return new CoreError("error.consumption.requests.cannotCreateRequestWithExpirationDateInPast", "You cannot create a Request with an expiration date that is in the past.");
```

packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -128,6 +128,13 @@ class Messages {
         );
     }
 
+    public cannotSendMessageWithExpiredRequest() {
+        return new ApplicationError(
+            "error.runtime.messages.cannotSendMessageWithExpiredRequest",
+            "The Message cannot be sent as the Request has already expired. Please create a new Request and try again."
```
Milena-Czierlinski, line 134:
```suggestion
            "The Message cannot be sent as the containing Request is already expired. Please create a new Request and try again."
```

jkoenig134: contained would be better wording, right?

jkoenig134:
ehm, I removed this from draft b/c I thought it's a mistake, but it looks like you set this manually

packages/consumption/test/modules/requests/RequestsIntegrationTest.ts
```diff
@@ -318,6 +318,17 @@ export class RequestsGiven {
...
+            const dateString = params.content.expiresAt.dateTime.toISO();
+            if (dateString) {
+                const date = CoreDate.from(dateString);
+                while (CoreDate.utc().isBefore(date)) {
+                    await sleep(1000);
+                }
+                await sleep(1000);
+            }
+        }
+
```
Milena-Czierlinski, line 331: In my opinion it is much easier to remove this here and simply add a sleep to those tests that require an expiration time to pass. Looking only at the tests, I didn't understand why we don't have a sleep there, since the true that was passed to `anOutgoingRequestWith` doesn't make it obvious, what's going on. Also, you don't have to adjust `ICoreDate` then. Exposing a DateTime there doesn't feel like something we want to do.

packages/runtime/test/transport/messages.test.ts
```diff
@@ -273,6 +275,70 @@ describe("Message errors", () => {
         expect(result).toBeAnError("The recipient does not match the Request's peer.", "error.runtime.validation.invalidPropertyValue");
     });
 
+    test("should throw correct error message for trying to send a Message with a Request that expired", async () => {
+        const requestItem = {
+            "@type": "ConsentRequestItem",
+            consent: "I consent to this RequestItem",
+            mustBeAccepted: true
+        };
+        const expiresAt = CoreDate.utc().add({ seconds: 5 }).toString();
```
Milena-Czierlinski, line 284: Have you tried if a shorter time also works? That would be great.

packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
```diff
@@ -726,12 +746,13 @@ describe("OutgoingRequestsController", function () {
 
         test("moves the Request to status 'Expired' when expiredAt is reached", async function () {
             await Given.anOutgoingRequestWith({
-                status: LocalRequestStatus.Draft,
+                status: LocalRequestStatus.Expired,
```
Milena-Czierlinski, line 749:
```suggestion
                status: LocalRequestStatus.Draft,
```
If the Request was already expired, there'd be no reason for this test.

packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
```diff
@@ -689,25 +707,27 @@ describe("OutgoingRequestsController", function () {
             await Given.anOutgoingRequestWith({
                 status: LocalRequestStatus.Expired,
                 content: {
-                    expiresAt: CoreDate.utc().subtract({ days: 1 }),
+                    expiresAt: CoreDate.utc().add({ second: 1 }),
```
Milena-Czierlinski, line 710: Can this be even shorter? If so, please also adjust the sleeps accordingly.

Same for other tests.

packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
jkoenig134, line 1: this test now runs 8 seconds longer with all those sleeps, this sums up quickly - we have to find an alternative.

sebbi08: @Milena-Czierlinski @jkoenig134 I now changed it to 100ms and also added a sleep for 100ms because with lower timing it started to be flaky even locally and also the wait was needed because sometimes the tests were faster.

sebbi08: Ok, not working in the CI, so I might need to think about another solution

Milena-Czierlinski: Okay, this is already much better. Please apply the same times to the `messages.test`.

Milena-Czierlinski: > Ok, not working in the CI, so I might need to think about another solution

That only refers to one test. Maybe it is already sufficient to wait for 200ms there.

jkoenig134: > 100ms and also added a sleep for 100ms

wtb 100 and 120?

packages/runtime/test/transport/messages.test.ts
```diff
@@ -273,6 +276,68 @@ describe("Message errors", () => {
...
+        const createRequestResult = (
+            await client1.consumption.outgoingRequests.create({
+                content: {
+                    expiresAt,
+                    items: [requestItem]
+                },
+                peer: client2.address
+            })
+        ).value;
+        await sleep(150);
```
Milena-Czierlinski, line 296: Please add empty lines around this line.

packages/runtime/test/transport/messages.test.ts
```diff
@@ -273,6 +276,68 @@ describe("Message errors", () => {
...
+            "error.runtime.messages.cannotSendMessageWithExpiredRequest"
+        );
+    });
+
+    test("should mark a Request as expired when synced after the expiration date", async () => {
+        const requestItem = {
+            "@type": "ConsentRequestItem",
+            consent: "I consent to this RequestItem",
+            mustBeAccepted: true
+        };
```
Milena-Czierlinski, line 312: Empty line below as in the previous test.

packages/runtime/test/transport/messages.test.ts
```diff
@@ -273,6 +276,68 @@ describe("Message errors", () => {
...
+        const expiresAt = CoreDate.utc().add({ milliseconds: 100 }).toString();
+        const createRequestResult = (
+            await client1.consumption.outgoingRequests.create({
+                content: {
+                    expiresAt,
+                    items: [requestItem]
+                },
+                peer: client2.address
+            })
+        ).value;
```
Milena-Czierlinski, line 322: Empty line below please.

Milena-Czierlinski (commented):
Apart from this I'm fine with the PR.

Milena-Czierlinski (dismissed):

jkoenig134 (dismissed):
LGTM, please wait for another review by @Milena-Czierlinski or @britsta

packages/consumption/test/modules/requests/OutgoingRequestsController.test.ts
```diff
@@ -280,6 +280,24 @@ describe("OutgoingRequestsController", function () {
                     "The Request cannot be created because its acceptance would lead to the creation of more than one RelationshipAttribute in the context of this Relationship with the same key 'uniqueKey', owner and value type."
             });
         });
+
+        test("returns a validation result that contains an error for requests that are expired", async function () {
```
britsta, line 284: Please put this test before the `"returns a validation result that contains an error for requests that would lead to the creation of more than one RelationshipAttribute with the same key"` test right above as the Request expiry test is less specific and fits better with the above tests than the Relationship `key` uniqueness test. :D

packages/runtime/test/transport/messages.test.ts
```diff
@@ -272,6 +275,73 @@ describe("Message errors", () => {
         expect(result).toBeAnError("The recipient does not match the Request's peer.", "error.runtime.validation.invalidPropertyValue");
     });
 
+    test("should throw correct error message for trying to send a Message with a Request that expired", async () => {
```
britsta, line 278: Please adjust the test name to `"should throw correct error for trying to send a Message with an expired Request"`.

("should throw correct error" instead of "should throw correct error message" for consistency with test names above and "an expired Request" instead of "a Request that expired" as it is grammatically preferable.)

packages/runtime/test/transport/messages.test.ts
```diff
@@ -272,6 +275,73 @@ describe("Message errors", () => {
...
+        const requestItem = {
+            "@type": "ConsentRequestItem",
+            consent: "I consent to this RequestItem",
+            mustBeAccepted: true
+        };
```
britsta, line 283: There isn't a need to define this `requestItem` again as it is already defined within the `beforeAll` block of the tests.

packages/runtime/test/transport/messages.test.ts
```diff
@@ -272,6 +275,73 @@ describe("Message errors", () => {
...
+            "error.runtime.messages.cannotSendMessageWithExpiredRequest"
+        );
+    });
+
+    test("should mark a Request as expired when synced after the expiration date", async () => {
+        const requestItem = {
+            "@type": "ConsentRequestItem",
+            consent: "I consent to this RequestItem",
+            mustBeAccepted: true
+        };
```
britsta, line 314: This `requestItem` can be omitted as well as it is already defined within the `beforeAll` block.

britsta (commented):
Just a few small things, rest LAwesomeTM.

britsta (approved):

PR #370, opened 2024-12-16
[POC, WIP] Reverse ajv errors 

Magnus-Kuhn:
# Readiness checklist

- [ ] I added/updated tests.
- [ ] I labeled the PR.
- [ ] I self-reviewed the PR.

# Description

In case the JSON schema contains `anyOf` and every option throws an error, ajv returns an array of all the errors thrown by each case followed by an overall error `must match a schema in anyOf`. Currently the first error is returned, this investigates changing it to the last error.

`anyOf` is used in
- distinguishing request item types and groups (both when creating and deciding a request)
- distinguishing response item types
- distinguishing attribute types (also in queries)
- attribute owners in a third party relationship attribute query
- request response source is message or relationship/ request source is message or template
- get requests (string | string[])
- truncated reference is from token / file / template
- challenge is for relationship / identity / device

It could also come up if messages or templates were validated with ajv with their individual ContentDerivations, currently they (and requests) are validated using serval.

Especially for requests, messages and similar with a high property depth, the error message `must match a schema` could be thrown too early to be helpful, e. g. if a contained attribute is wrong. A possible approach would be to use `@type` annotations to then figure out what the correct content should be and then run another validation against that schema.
TestRequestItems and TestNotificationItems would also potentially need to be changed.

### DIFF
--- a/packages/runtime/scripts/buildSchemas.mjs
+++ b/packages/runtime/scripts/buildSchemas.mjs
@@ -17,7 +17,7 @@ const requestTypes = schemaGenerator
 const schemaDeclarations = requestTypes
     .map((type) => {
         try {
-            const schema = schemaGenerator.createSchema(type);
+            const schema = reverseJsonSchemaProperties(schemaGenerator.createSchema(type));
             return `export const ${type}: any = ${JSON.stringify(schema, undefined, 4)}`;
         } catch (e) {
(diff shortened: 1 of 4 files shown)


Magnus-Kuhn:
> @Magnus-Kuhn this can be closed & deleted now, right?

I think so, it would still be an improvement but it's not one that's asked for by the customer. We can revisit it later if it's needed. I'm closing it.

PR #374, opened 2024-12-19
Allow uploading empty files

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [ ] I labeled the PR.
- [ ] I self-reviewed the PR.

# Description

As discussed with @stnmtz empty files are valid files so we shouldn't block uploading them.

### DIFF
--- a/packages/runtime/src/useCases/transport/files/UploadOwnFile.ts
+++ b/packages/runtime/src/useCases/transport/files/UploadOwnFile.ts
@@ -44,15 +44,6 @@ class Validator extends SchemaValidator<UploadOwnFileValidatableRequest> {
             );
         }
 
-        if (input.content.length === 0) {
-            validationResult.addFailure(
-                new ValidationFailure(
-                    RuntimeErrors.general.invalidPropertyValue(`'${nameof<UploadOwnFileValidatableRequest>((r) => r.content)}' is empty`),
-                    nameof<UploadOwnFileValidatableRequest>((r) => r.content)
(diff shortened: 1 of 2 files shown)


jkoenig134:
This also fixes the issues that `'content' is empty` was a bad error message for integrators as they didn't work with the content itself and instead with a multipart file containing a buffer.

Milena-Czierlinski:
> As discussed with @stnmtz empty files are valid files so we shouldn't block uploading them.

As I was not part of this discussion, can you please explain why we should allow empty files? Is there a use case for this that lead to this discussion?

jkoenig134:
> > As discussed with @stnmtz empty files are valid files so we shouldn't block uploading them.
> 
> As I was not part of this discussion, can you please explain why we should allow empty files? Is there a use case for this that lead to this discussion?

The main reason is the bad error message. And why should we block it?

Milena-Czierlinski:
> And why should we block it?

Why should we allow it? I fear this will only lead to problems later in the process.

jkoenig134:
> > And why should we block it?
> 
> Why should we allow it? I fear this will only lead to problems later in the process.

TBH - I don't see any problems.

Milena-Czierlinski:
Ok, we discussed to remove the validation for now, as other platforms don't have such a validation either.

britsta (approved):

Milena-Czierlinski (approved):

PR #376, opened 2024-12-20
Add debug log for SynchronizedCollection in CI runs for database writes with outdated data

sebbi08:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/transport/src/modules/accounts/AccountController.ts
+++ b/packages/transport/src/modules/accounts/AccountController.ts
@@ -438,10 +438,10 @@ export class AccountController {
     public async getSynchronizedCollection(collectionName: string): Promise<SynchronizedCollection> {
         const collection = await this.db.getCollection(collectionName);
         if (!this.config.datawalletEnabled) {
-            return new SynchronizedCollection(collection, this.config.supportedDatawalletVersion);
+            return new SynchronizedCollection(collection, this.config.supportedDatawalletVersion, undefined, this.config.debug);
         }
 
-        return new SynchronizedCollection(collection, this.config.supportedDatawalletVersion, this.unpushedDatawalletModifications);
+        return new SynchronizedCollection(collection, this.config.supportedDatawalletVersion, this.unpushedDatawalletModifications, this.config.debug);
     }
 
     public async cleanupDataOfDecomposedRelationship(relationship: Relationship): Promise<void> {
--- a/packages/transport/src/modules/accounts/AccountController.ts
+++ b/packages/transport/src/modules/accounts/AccountController.ts
@@ -438,10 +438,10 @@ export class AccountController {
     public async getSynchronizedCollection(collectionName: string): Promise<SynchronizedCollection> {
         const collection = await this.db.getCollection(collectionName);
         if (!this.config.datawalletEnabled) {
-            return new SynchronizedCollection(collection, this.config.supportedDatawalletVersion, undefined, this.config.debug);
+            return new SynchronizedCollection(collection, this.config.supportedDatawalletVersion);
         }
 
-        return new SynchronizedCollection(collection, this.config.supportedDatawalletVersion, this.unpushedDatawalletModifications, this.config.debug);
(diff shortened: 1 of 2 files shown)


packages/transport/src/modules/sync/SynchronizedCollection.ts
```diff
@@ -87,6 +88,16 @@ export class SynchronizedCollection implements IDatabaseCollection {
 
         const newObjectJson = newObject.toJSON();
 
+        if (this.debugMode) {
+            const oldDocUpdated = Serializable.fromUnknown(await this.parent.read(newObject.id.toString()));
+
+            const readDiff = jsonpatch.compare(oldDocUpdated.toJSON(), oldObject.toJSON());
+            if (readDiff.length > 0) {
+                // eslint-disable-next-line no-console
+                console.error(`Object has been changed since last read.\n${new Error().stack}\n${JSON.stringify(readDiff, null, 2)}`);
```
jkoenig134, line 97: I'd like to give **way more** context on what's wrong here.

Also maybe we can find out that we are in a test any only log then? B/c I think the debugMode flag is also set in the connector, we then don't want to have this kind of log there.

jkoenig134: from my research one possibility would be to use `typeof globalThis.it === 'function';` to see if you are in a testing environment.

jkoenig134: great idea. but we cannot access process.env in prod code to be browser compatible ..

sebbi08: I just tested this `if (typeof process === "object" && process.env.CI === "true")` and it works in node and browser.

I just need to test if the bundling in the app has any problem with it

packages/transport/src/modules/sync/SynchronizedCollection.ts
```diff
@@ -15,7 +15,8 @@ export class SynchronizedCollection implements IDatabaseCollection {
     public constructor(
         private readonly parent: IDatabaseCollection,
         private readonly datawalletVersion: number,
-        private readonly datawalletModifications?: IDatabaseCollection
+        private readonly datawalletModifications?: IDatabaseCollection,
+        private readonly debugMode?: boolean
```
jkoenig134, line 19: this is unused now

packages/transport/src/modules/sync/SynchronizedCollection.ts
```diff
@@ -87,6 +88,23 @@ export class SynchronizedCollection implements IDatabaseCollection {
 
         const newObjectJson = newObject.toJSON();
 
+        if (typeof process === "object" && process.env.CI === "true") {
```
jkoenig134, line 91:
```suggestion
        if (typeof globalThis.process === "object" && globalThis.process.env.CI) {
```

two things:
- used globalThis to be clear about where process should be defined or not
- don't exactly pin CI (we don't care what's inside this var)

packages/transport/src/modules/accounts/AccountController.ts
```diff
@@ -438,10 +438,10 @@ export class AccountController {
     public async getSynchronizedCollection(collectionName: string): Promise<SynchronizedCollection> {
         const collection = await this.db.getCollection(collectionName);
         if (!this.config.datawalletEnabled) {
-            return new SynchronizedCollection(collection, this.config.supportedDatawalletVersion);
+            return new SynchronizedCollection(collection, this.config.supportedDatawalletVersion, undefined, this.config.debug);
         }
 
-        return new SynchronizedCollection(collection, this.config.supportedDatawalletVersion, this.unpushedDatawalletModifications);
+        return new SynchronizedCollection(collection, this.config.supportedDatawalletVersion, this.unpushedDatawalletModifications, this.config.debug);
```
jkoenig134, line 444: you should also undo that.

jkoenig134 (changes requested):

jkoenig134 (approved):

PR #382, opened 2025-01-07
Improve pipeline debug logging for SynchronizedCollection data inconsistencies

sebbi08:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/transport/src/modules/sync/SynchronizedCollection.ts
+++ b/packages/transport/src/modules/sync/SynchronizedCollection.ts
@@ -88,9 +88,9 @@ export class SynchronizedCollection implements IDatabaseCollection {
         const newObjectJson = newObject.toJSON();
 
         if (typeof globalThis.process === "object" && globalThis.process.env.CI) {
-            const oldDocUpdated = Serializable.fromUnknown(await this.parent.read(newObject.id.toString()));
+            const databaseObject = Serializable.fromUnknown(await this.parent.read(newObject.id.toString()));
 
-            const readDiff = jsonpatch.compare(oldDocUpdated.toJSON(), oldObject.toJSON());
+            const readDiff = jsonpatch.compare(databaseObject.toJSON(), oldObject.toJSON());
             if (readDiff.length > 0) {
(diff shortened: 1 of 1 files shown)


jkoenig134:
@sebbi08 please re-check the PR title, you checked the `I ensured that the PR title is good enough for the changelog.` checkbox, but it isn't.

Please treat these checkboxes as actual task that need to be done before requesting any reviews.

packages/transport/src/modules/sync/SynchronizedCollection.ts
```diff
@@ -88,16 +88,21 @@ export class SynchronizedCollection implements IDatabaseCollection {
...
             if (readDiff.length > 0) {
                 // eslint-disable-next-line no-console
                 console.error(`
 The data that is currently updated got modified between it initial reading and this update.
 This will lead to an data loss and inconsistency.
 Here is the diff of the data:
 ${JSON.stringify(readDiff, null, 2)}
+Old Object from Database:
+${JSON.stringify(oldDoc, null, 2)}
+Object in Database:
```
jkoenig134, line 103: `Object in Database` is a very strange wording. Maybe `Object to be stored to the Database` is more accurate?

jkoenig134: or `Object to write`

sebbi08: This is the Object from the DB that got written in the meantime since the last read, the `Object to write` would the `newObject`

jkoenig134: ehm, what? :D and what's oldDoc?

sebbi08: The `oldDoc` is the document that was initial read by the calling functions.
And the `databaseObject` is the current document in the database.

jkoenig134 (approved):
please make sure to have a correct commit message (as you only have one commit this will not be automatic)

PR #383, opened 2025-01-08
Add Runtime errors for IdentityMetadata use cases

britsta:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

It should not be possible to create IdentityMetadata for Identities whose `address` is not known. An `address` is known if it is your own `address` or the `address` of a `peer` of a Relationship.

In addition, a special error for records that could not be found was introduced for IdentityMetadata. This was necessary because IdentityMetadata is usually accessed by an Integrator of a Connector via its combination of `reference` and `ke`y and not via its `id`.


packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -276,6 +276,19 @@ class IdentityDeletionProcess {
     }
 }
 
+class IdentityMetadata {
+    public recordNotFound() {
```
jkoenig134, line 280: the naming record from general errors is chosen to be generic, IMO you can write identityMetadata instead of record here.

packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -276,6 +276,19 @@ class IdentityDeletionProcess {
...
+        return new ApplicationError("error.runtime.identityMetadata.recordNotFound", "There is no stored IdentityMetadata for the specified combination of reference and key.");
+    }
+
+    public unknownReferencedIdentity() {
+        return new ApplicationError(
+            "error.runtime.identityMetadata.unknownReferencedIdentity",
+            "The reference of the IdentityMetadata does not resolve to the address of a known Identity. IdentityMetadata can only be upserted for the own Identity or for the peers of the Relationships."
```
jkoenig134, line 287: Nothing about the code `unknownReferencedIdentity` makes clear that this can only be used for the upsert usecase. Maybe we can either make the code clearer or make the message a bit more generic.

=> imagine we check for an existing identity in get / delete and throw an error there, you would just pick this error based on the method name

packages/runtime/src/useCases/common/RuntimeErrors.ts
britsta: I chose it like this at first. However, I thought "error.runtime.identityMetadata.identityMetadataNotFound" could be too redundant. Do you think it's better despite the redundancy? Then I have no problem changing it.

jkoenig134: IMO it's redundancy vs. "understandability"

jkoenig134: @Milena-Czierlinski wdym?

Milena-Czierlinski: `error.runtime.identityMetadata.notFound` is fine for me.

jkoenig134 (approved):
LGTM

PR #387, opened 2025-01-13
Extend RequestConfig of DeciderModule depending on existence of Relationship

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/runtime/src/modules/DeciderModule.ts
+++ b/packages/runtime/src/modules/DeciderModule.ts
@@ -1,5 +1,6 @@
+import { ApplicationError } from "@js-soft/ts-utils";
 import { LocalRequestStatus } from "@nmshd/consumption";
-import { RequestItemGroupJSON, RequestItemJSONDerivations } from "@nmshd/content";
+import { isRequestItemDerivation, RequestItemGroupJSON, RequestItemJSONDerivations } from "@nmshd/content";
 import { CoreDate } from "@nmshd/core-types";
 import {
     IncomingRequestStatusChangedEvent,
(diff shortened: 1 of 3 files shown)


packages/runtime/src/modules/DeciderModule.ts
```diff
@@ -287,15 +295,35 @@ function filterConfigElementByPrefix(requestItemConfigElement: RequestItemDeriva
     return filteredRequestItemConfigElement;
 }
 
-function checkCompatibility(requestConfigElement: RequestConfig, requestOrRequestItem: LocalRequestDTO | RequestItemJSONDerivations): boolean {
+async function checkCompatibility(requestConfigElement: RequestConfig, requestOrRequestItem: LocalRequestDTO, services: RuntimeServices): Promise<boolean | Error>;
+async function checkCompatibility(requestConfigElement: RequestConfig, requestOrRequestItem: RequestItemJSONDerivations): Promise<boolean>;
+async function checkCompatibility(
```
jkoenig134, line 300: as this is not exported

```suggestion
async function checkCompatibility(
```

should be enough

jkoenig134 (changes requested):
only one small detail

packages/runtime/src/modules/DeciderModule.ts
Milena-Czierlinski: Given you suggestion, in line 442 of DeciderModule.ts the compiler doesn't know that checkCompatibility won't return an error.

Milena-Czierlinski: Also, it makes it clearer why we can use `services!` in line 319 in my opinion.

jkoenig134: Hm, I have to look at this in  the Editor.

jkoenig134 (approved):

packages/runtime/src/modules/DeciderModule.ts
jkoenig134: I still think it's over-engeneered for a helper method, but if it works - go :)

PR #390, opened 2025-01-16
Update usable attribute valuetypes

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [ ] I labeled the PR.
- [ ] I self-reviewed the PR.


packages/content/test/attributes/IdentityAttribute.test.ts
```diff
@@ -119,37 +119,38 @@ describe("IdentityAttribute", function () {
...
+    //     expect(() =>
+    //         IdentityAttribute.from({
+    //             value: {
+    //                 "@type": "BirthMonth",
+    //                 value: "10"
+    //             },
+    //             owner: CoreAddress.from("address")
+    //         })
+    //     ).toThrow("BirthMonth.value :: Value is not a number");
+    // });
```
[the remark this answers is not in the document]
jkoenig134: I think I commented it out b/c I wanted to get back to it and then forgot about it. Very strange that eslint didn't detected it.

packages/content/test/attributes/IdentityAttribute.test.ts
```diff
@@ -201,14 +207,24 @@ describe("IdentityAttribute", function () {
     test("should validate attribute values from objects", function () {
```
Milena-Czierlinski, line 207: I see similarities to the test "should allow to validate integer values" above. Why do we have two different tests for that?

packages/runtime/test/dataViews/DraftIdentityAttributeDVO.test.ts
```diff
@@ -231,43 +135,4 @@ describe("DraftIdentityAttributeDVO", () => {
         expect(dvo.valueHints.values).toHaveLength(183);
         expect(dvo.valueHints.values![31]).toStrictEqual({ key: "de", displayName: "i18n://attributes.values.languages.de" });
     });
-
-    test("check a complex PersonName", async () => {
```
Milena-Czierlinski, line 235: Maybe we should add a test for another complex Attribute anyways.

packages/runtime/test/dataViews/requestItems/ComplexReadAttributeRequestItemDVO.test.ts
```diff
@@ -130,11 +122,9 @@ describe("ComplexReadAttributeRequestItemDVO with IdentityAttributeQuery", () =>
         expect(requestItemDVO.query).toBeDefined();
         expect(requestItemDVO.query.type).toBe("IdentityAttributeQueryDVO");
         const identityAttributeQueryDVO = requestItemDVO.query as IdentityAttributeQueryDVO;
-        expect(identityAttributeQueryDVO.renderHints.technicalType).toBe("Object");
-        expect(identityAttributeQueryDVO.renderHints.editType).toBe("Complex");
```
Milena-Czierlinski, line 134: I doubt that this whole test file ("Complex...") makes sense if we use Surname instead of PersonName. Actually, I've never noticed this renderHints.editType "Complex" before. But if we are to keep these tests, we should use e.g. StreetAddress instead of Surname.

Milena-Czierlinski (changes requested):

packages/content/test/attributes/IdentityAttribute.test.ts
jkoenig134: you should ask the original author of those tests :D

packages/runtime/test/dataViews/requestItems/ComplexReadAttributeRequestItemDVO.test.ts
jkoenig134: :D yeah, my solution is dogshit here.. I'll switch to streetaddress or so

packages/content/test/attributes/IdentityAttribute.test.ts
jkoenig134: great.

PR #394, opened 2025-01-21
Move check for RepositoryAttribute duplicates to AttributesController

Milena-Czierlinski:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

I moved this to a lower level to be able to reuse it. This is advantageous for preventing the creation of duplicate RepositoryAttributes via a Request. Also, it allows to implement a CanCreateRepositoryAttribute use case more easily.

### DIFF
--- a/packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
+++ b/packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
@@ -5,7 +5,7 @@ import { AccountController } from "@nmshd/transport";
 import { Inject } from "@nmshd/typescript-ioc";
 import _ from "lodash";
 import { LocalAttributeDTO } from "../../../types";
-import { flattenObject, ISO8601DateTimeString, RuntimeErrors, SchemaRepository, SchemaValidator, UseCase, ValidationFailure, ValidationResult } from "../../common";
+import { ISO8601DateTimeString, RuntimeErrors, SchemaRepository, SchemaValidator, UseCase, ValidationFailure, ValidationResult } from "../../common";
 import { IValidator } from "../../common/validation/IValidator";
 import { AttributeMapper } from "./AttributeMapper";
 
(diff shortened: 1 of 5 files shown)


packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
```diff
@@ -64,41 +64,29 @@ class Validator implements IValidator<CreateRepositoryAttributeRequest> {
...
         @Inject private readonly accountController: AccountController,
         @Inject validator: Validator
     ) {
         super(validator);
     }
 
     protected async executeInternal(request: CreateRepositoryAttributeRequest): Promise<Result<LocalAttributeDTO>> {
+        const duplicateRepositoryAttributes = await this.attributesController.getRepositoryAttributesWithSameValue(request.content.value);
+
+        const exactMatchExists = duplicateRepositoryAttributes.some((duplicate) => _.isEqual(duplicate.content.value.toJSON(), request.content.value));
```
britsta, line 77: I'd expect the exact match logic within the `getRepositoryAttributesWithSameValue` method in the AttributesController as this is what I would understand by "same value". It should also increase reusability of the method.

Milena-Czierlinski: `exactMatchExists` is only a boolean, but I would like to be more flexible and to be able to explicitly state what the duplicate Attribute is. This is why I implemented `getRepositoryAttributesWithSameValue` similar to `getRelationshipAttributesOfValueTypeToPeerWithGivenKeyAndOwner`.

Magnus-Kuhn: `.find` instead of `.some` would get that job done

Magnus-Kuhn: `.find` instead of `.some` would get that job done

britsta: I did not mean to calculate the boolean in the method, but as @Magnus-Kuhn suggested, to restrict the found RepositoryAttributes so that it is an exact match. I would find it confusing from the method name `getRepositoryAttributesWithSameValue` if RepositoryAttributes are also returned by the method, although they are not an exact match.

Milena-Czierlinski: Gotcha!

Milena-Czierlinski: Thinking about this again, do you guys know why there is the exactMatchExists logic in the first place? Shouldn't this already be covered by the query anyway?

jkoenig134 (approved):

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
Milena-Czierlinski: The exactMatchExists logic checks for optional properties, too. So I moved it to the AttributesController like you suggested.

PR #396, opened 2025-01-22
Add CanCreateRepositoryAttribute use case

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


packages/runtime/src/useCases/consumption/attributes/CanCreateRepositoryAttribute.ts
```diff
@@ -0,0 +1,63 @@
...
+class Validator implements IValidator<CanCreateRepositoryAttributeRequest> {
+    public constructor(@Inject private readonly schemaRepository: SchemaRepository) {}
+
+    public validate(value: CanCreateRepositoryAttributeRequest): ValidationResult {
+        const requestSchemaValidator = new SchemaValidator(this.schemaRepository.getSchema("CanCreateRepositoryAttributeRequest"));
+        return requestSchemaValidator.validate(value);
+    }
+}
+
+export type CanCreateRepositoryAttributeResponse =
```
Milena-Czierlinski, line 31: The tests fail with `Property 'message'/'code' does not exist on type 'CanCreateRepositoryAttributeResponse'.` and the compiler only knows CanCreateRepositoryAttributeResponse as `{ isSuccess: true }`. Does someone have an idea why that is? @jkoenig134 @britsta @Magnus-Kuhn  I did the implementation according to CanCreateRelationshipResponse.

jkoenig134: I think you have to do a typeguard here?

jkoenig134: in this case a check on isSuccess = false

Milena-Czierlinski: Do you know why this is different to CanCreateRelationshipResponse where no typeguards are used?

jkoenig134: I'm assuming there is used a typeguard in any way :)

jkoenig134: `assert(!canCreateRelationshipResponse.isSuccess);`

Milena-Czierlinski: Ahh thank you!

packages/runtime/src/useCases/consumption/attributes/CreateRepositoryAttribute.ts
```diff
@@ -29,6 +29,15 @@ class Validator implements IValidator<CreateRepositoryAttributeRequest> {
         const requestValidationResult = requestSchemaValidator.validate(value);
         if (requestValidationResult.isInvalid()) return requestValidationResult;
 
+        const attributeValueTypeValidator = new IdentityAttributeValueTypeValidator(this.schemaRepository);
```
Magnus-Kuhn, line 32: Only one nitpick: With that name I would expect the validation to only validate the `@type` and not more of the attribute content.

Magnus-Kuhn (approved):

PR #397, opened 2025-01-24
Restrict character sets for attribute values

Magnus-Kuhn:
# Readiness checklist

- [x] I added/updated tests.
- [X] I labeled the PR.
- [x] I self-reviewed the PR.


jkoenig134:
@Magnus-Kuhn as this is a breaking change please

1. put it back to draft
2. re-add wip

until we are ready to merge.

packages/content/test/attributes/CharacterSets.test.ts
```diff
@@ -0,0 +1,307 @@
...
+});
+
+test.each(proprietaryAttributeTestParameters)("description of $type is considered as invalid", ({ type, value, valueType, confidentiality }) => {
+    const invalidCall = () => {
+        Serializable.fromUnknown({ "@type": type, value, valueType, confidentiality, title: "aTitle", description: "invalidValueΩ" });
+    };
+    expect(invalidCall).toThrow(new ParsingError(type, "description", errorMessageDatatypeC));
+});
+
+test("Key of RelationshipAttribute is considered as valid", () => {
```
britsta, line 190: Please write the `key` lowercase.

Same for the tests below.

packages/content/src/attributes/constants/CharacterSets.ts
```diff
@@ -0,0 +1,13 @@
+export const characterSets: Record<string, RegExp> = {
+    // DIN91379 character sets from https://xoev.de/schemata/din/91379/2022-08/din-norm-91379-datatypes.xsd
+    // See the sources in the browser's dev mode, they're not equal to what's initially displayed
+    // datatype A for names of natural persons
```
britsta, line 4: Usually "data type" is written instead of "datatype". However, if you want to keep the proper names "datatypeA", "datatypeB" and "datatypeC" from the mentioned file, you should remove the space between "datatype" and "A".

Same considerations for "datatype B" and "datatype C" below.

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -1012,7 +1012,7 @@ describe("RelationshipDecomposition", () => {
 
         await sendAndReceiveNotification(services1.transport, services2.transport, services2.consumption);
 
-        const randomName1 = await Random.string(7);
+        const randomName1 = await Random.string(7, "qwertzuiop");
```
britsta, line 1015: Please use `RandomCharacterRange.LowerCase` or similar instead of "qwertzuiop".

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -1022,7 +1022,7 @@ describe("RelationshipDecomposition", () => {
             }
         });
 
-        const randomName2 = await Random.string(7);
+        const randomName2 = await Random.string(7, "asdfghjkl");
```
britsta, line 1025: Please use `RandomCharacterRange.LowerCase` or similar instead of "asdfghjkl".

packages/content/src/attributes/types/proprietary/ProprietaryJSON.ts
```diff
@@ -56,6 +57,13 @@ function validateValue(value: any) {
...
         }
+
+        // JSON.stringify replaces some characters with their code points (e. g. "\\u0012" instead of the character), hence transforming it back for the regex validation
+        const stringContainingStringifiedJSON = `"${string.replace(/(?<!\\)"/g, '\\"')}"`;
+        const parsedString = JSON.parse(stringContainingStringifiedJSON);
+        if (!characterSets.din91379DatatypeC.test(parsedString)) {
+            return "Value does not match regular expression /^([\\u0009-\\u000A]|\\u000D|[ -~]|[ -¬]|[®-ž]|[Ƈ-ƈ]|Ə|Ɨ|[Ơ-ơ]|[Ư-ư]|Ʒ|[Ǎ-ǜ]|[Ǟ-ǟ]|[Ǣ-ǰ]|[Ǵ-ǵ]|[Ǹ-ǿ]|[Ȓ-ȓ]|[Ș-ț]|[Ȟ-ȟ]|[ȧ-ȳ]|ə|ɨ|ʒ|[ʹ-ʺ]|[ʾ-ʿ]|ˈ|ˌ|[Ḃ-ḃ]|[Ḇ-ḇ]|[Ḋ-ḑ]|ḗ|[Ḝ-ḫ]|[ḯ-ḷ]|[Ḻ-ḻ]|[Ṁ-ṉ]|[Ṓ-ṛ]|[Ṟ-ṣ]|[Ṫ-ṯ]|[Ẁ-ẇ]|[Ẍ-ẗ]|ẞ|[Ạ-ỹ]|’|‡|€|A̋|C(̀|̄|̆|̈|̕|̣|̦|̨̆)|D̂|F(̀|̄)|G̀|H(̄|̦|̱)|J(́|̌)|K(̀|̂|̄|̇|̕|̛|̦|͟H|͟h)|L(̂|̥|̥̄|̦)|M(̀|̂|̆|̐)|N(̂|̄|̆|̦)|P(̀|̄|̕|̣)|R(̆|̥|̥̄)|S(̀|̄|̛̄|̱)|T(̀|̄|̈|̕|̛)|U̇|Z(̀|̄|̆|̈|̧)|a̋|c(̀|̄|̆|̈|̕|̣|̦|̨̆)|d̂|f(̀|̄)|g̀|h(̄|̦)|j́|k(̀|̂|̄|̇|̕|̛|̦|͟h)|l(̂|̥|̥̄|̦)|m(̀|̂|̆|̐)|n(̂|̄|̆|̦)|p(̀|̄|̕|̣)|r(̆|̥|̥̄)|s(̀|̄|̛̄|̱)|t(̀|̄|̕|̛)|u̇|z(̀|̄|̆|̈|̧)|Ç̆|Û̄|ç̆|û̄|ÿ́|Č(̕|̣)|č(̕|̣)|ē̍|Ī́|ī́|ō̍|Ž(̦|̧)|ž(̦|̧)|Ḳ̄|ḳ̄|Ṣ̄|ṣ̄|Ṭ̄|ṭ̄|Ạ̈|ạ̈|Ọ̈|ọ̈|Ụ(̄|̈)|ụ(̄|̈))*$/";
```
Magnus-Kuhn, line 65: Do we want to validate ProprietaryJSON / ProprietaryXML?

Milena-Czierlinski: My first thought is to not validate them. I hardly believe an App user will create such Attributes and if an Integrator needs certain expressions, we should probably not restrict them.

Milena-Czierlinski: @Magnus-Kuhn If you remove the validation (and bring the branch up to date) I think this PR is good to go.

Magnus-Kuhn: Alright, I'll do it like that for now, but I'd still want @stnmtz to have a look sometime later - he has a better idea of what the integrators' systems can handle

Milena-Czierlinski: Gotcha! Then we'll urge @stnmtz to review this as soon as we want to merge our breaking changes. :D

Milena-Czierlinski (dismissed):
Requires additional approval of @stnmtz once we want to merge this PR.

packages/content/src/attributes/RelationshipAttribute.ts
```diff
@@ -28,7 +29,7 @@ export class RelationshipAttribute<TValueClass extends AttributeValues.Relations
     public value: TValueClass;
 
     @serialize()
-    @validate({ max: 100 })
+    @validate({ max: 100, regExp: characterSets.din91379DatatypeC })
```
jkoenig134, line 32: this key is only machine read, I don't see a reason why we should limit the charset here.

packages/content/src/attributes/RelationshipAttributeQuery.ts
```diff
@@ -30,7 +31,7 @@ export interface IRelationshipAttributeCreationHints extends ISerializable {
 @type("RelationshipAttributeCreationHints")
 export class RelationshipAttributeCreationHints extends Serializable implements IRelationshipAttributeCreationHints {
     @serialize()
-    @validate({ max: PROPRIETARY_ATTRIBUTE_MAX_TITLE_LENGTH })
+    @validate({ max: PROPRIETARY_ATTRIBUTE_MAX_TITLE_LENGTH, regExp: characterSets.din91379DatatypeC })
```
jkoenig134, line 34: why are title and description validated?

jkoenig134 (changes requested):
Seeing what you did with https://github.com/nmshd/runtime/pull/437 I think we should do all this in the controller and not on serval level. Additionally it wouldn't be a breaking change then.

packages/content/src/attributes/RelationshipAttribute.ts
Magnus-Kuhn: My reason for validating everything was that integrators can safely copy attributes into their systems, even if they can't handle all characters

packages/content/src/attributes/RelationshipAttributeQuery.ts
Magnus-Kuhn: If a query has forbidden characters, then there would be no attribute that fits to the query

packages/content/src/attributes/RelationshipAttribute.ts
jkoenig134: I don't see someone not being able to "handle" this.

Magnus-Kuhn:
> Seeing what you did with #437 I think we should do all this in the controller and not on serval level. Additionally it wouldn't be a breaking change then.

Doing it in serval validates requests and notifications as well

packages/content/src/attributes/RelationshipAttribute.ts
Magnus-Kuhn: The reason for creating the norm and making it mandatory was that there were non-conformant IT systems (some that weren't even using UTF8) - see https://www.it-planungsrat.de/fileadmin/beschluesse/2022/Beschluss2022-51_Umsetzung.pdf. I would not be as confident with this.

packages/content/src/attributes/RelationshipAttributeQuery.ts
jkoenig134: ehm, what? this would require that title and description are validated in the attribute itself..

jkoenig134:
> > Seeing what you did with #437 I think we should do all this in the controller and not on serval level. Additionally it wouldn't be a breaking change then.
> 
> Doing it in serval validates requests and notifications as well

But it requires you to duplicate the logic _**everywhere**_ ... I don't see a reason for doing that.

packages/content/src/attributes/RelationshipAttribute.ts
jkoenig134: Ehm, nope?! The main reason is that we don't want to have `✌🏼😜😜😜🤣😆🥹😅` as a display name or surname.

packages/content/src/attributes/RelationshipAttributeQuery.ts
Magnus-Kuhn: They are currently validated in the Proprietary... types

jkoenig134: True, but also there I'd heavily argue that this is necessary.

packages/content/src/attributes/RelationshipAttribute.ts
Magnus-Kuhn: Okay, then there has been some misunderstanding

jkoenig134:
> > > Seeing what you did with #437 I think we should do all this in the controller and not on serval level. Additionally it wouldn't be a breaking change then.
> > 
> > 
> > Doing it in serval validates requests and notifications as well
> 
> But it requires you to duplicate the logic _**everywhere**_ ... I don't see a reason for doing that.

Additionally the error message for regex sucks. When doing that in controllers we would have the possibility to return a WAY BETTER error message than `hey, your entered value does not conform vberuozfapmwjewhrubwihrieoawdkvjqnljfldwawer78f6r7w1d3wef4ads8615as, please deal with it`

jkoenig134:
> Also generally the regex checks for e.g. Email occur in content - which also has a big error message. Do you suggest moving them all (and maybe the string length checks as well) into the controller?

I don't know why this has to be a general thing. I see a thing that gets introduced in almost every content class, which is not ok and we should find a solution for that - and I proposed a solution.

Projecting this to "but then we should remove all regex validations" is very misplaced here.

Magnus-Kuhn:
https://github.com/nmshd/documentation/pull/309 does documentation, https://github.com/nmshd/runtime/pull/448 is an alternative approach

PR #399, opened 2025-01-27
Make `title` optional in FileDTO

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -1832,9 +1832,10 @@ export class DataViewExpander {
     public async expandFileDTO(file: FileDTO): Promise<FileDVO> {
         return {
             ...file,
+            title: file.title ?? file.filename,
```
britsta, line 1835: This is needed in addition to the `name` because the `...file` no longer necessarily contains a `title`, right?

jkoenig134: ehm, kind of. before it checked for emptiness, now for undefinedness.

britsta (approved):

PR #401, opened 2025-01-27
Relationship creation without accepting Request of RelationshipTemplateContent

britsta:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

It shouldn't be possible to create a Relationship based on a RelationshipTemplate with a RelationshipTemplateContent if the Request within the `onNewRelationship` property of the RelationshipTemplateContent wasn't accepted and the provided `creationContent` of the Relationship isn't a RelationshipCreationContent whose `response` is the Response of the accepted Request. Validation needs to be added for the CreateRelationshipUseCase as well as for the CanCreateRelationshipUseCase.


packages/runtime/src/useCases/transport/relationships/CanCreateRelationship.ts
```diff
@@ -54,11 +59,48 @@ export class CanCreateRelationshipUseCase extends UseCase<CanCreateRelationshipR
                 );
                 return Result.ok({ isSuccess: false, code: error.code, message: error.message });
             }
-        }
 
-        const canSendRelationship = await this.relationshipController.canSendRelationship({ creationContent: request.creationContent, template });
-        if (canSendRelationship.isError) return Result.ok({ isSuccess: false, code: canSendRelationship.error.code, message: canSendRelationship.error.reason });
+            const transformedTemplateContent = template.cache.content;
+            if (transformedTemplateContent instanceof ArbitraryRelationshipTemplateContent && !(transformedCreationContent instanceof ArbitraryRelationshipCreationContent)) {
```
Magnus-Kuhn, line 64: The documentation bans ArbitraryCreationContent and non-arbitrary TemplateContent - I guess to protect app users, but doesn't cover this case. I think we could be more permissive here, but I don't know who uses the arbitrary contents to be fair. I'm also  fine with banning this.

packages/runtime/test/lib/testUtils.ts
```diff
@@ -394,13 +395,39 @@ export async function establishRelationship(transportServices1: TransportService
 
 export async function establishRelationshipWithContents(
```
Magnus-Kuhn, line 396: You could use `establishPendingRelationshipWithRequestFlow`

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -71,39 +91,284 @@ afterAll(() => serviceProvider.stop());
...
+
+            const createRelationshipResponse = await services5.transport.relationships.createRelationship({
+                templateId: templateId,
+                creationContent: RelationshipCreationContent.from({ response: acceptedRequest.response!.content }).toJSON()
+            });
+            expect(createRelationshipResponse).toBeSuccessful();
+        });
+    });
+
+    test("load RelationshipTemplate in Connector 2", async () => {
```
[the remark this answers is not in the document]
britsta: Let it go, let it go! ❄️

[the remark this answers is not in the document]
jkoenig134: ehm ... connector?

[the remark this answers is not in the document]
britsta: Yeah, the test seems to be quite old, I remove it.

packages/runtime/src/useCases/transport/relationships/CanCreateRelationship.ts
```diff
@@ -54,11 +59,48 @@ export class CanCreateRelationshipUseCase extends UseCase<CanCreateRelationshipR
...
+            "source.reference": templateId,
+            "response.content.result": "Accepted"
+        });
+
+        if (acceptedIncomingRequests.length === 0) {
+            return RuntimeErrors.relationships.noAcceptedIncomingRequest();
+        }
+
+        if (JSON.stringify(acceptedIncomingRequests[0].response!.content) !== JSON.stringify(relationshipCreationContent.response)) {
+            return RuntimeErrors.relationships.wrongResponseProvidedAsCreationContent();
```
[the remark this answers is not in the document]
britsta: I'm also not sure which is more performant, so I would just use your suggestion for now.

packages/runtime/test/lib/testUtils.ts
britsta: The `establishPendingRelationshipWithRequestFlow` function assumes an enabled RequestModule and can therefore not be used here.

packages/runtime/src/useCases/transport/relationships/CanCreateRelationship.ts
britsta: I think if we have an extra ArbitraryRelationshipCreationContent, we should also expect it to be used in the case in which it may be used.

packages/runtime/src/useCases/transport/relationships/CreateRelationship.ts
```diff
@@ -36,6 +38,9 @@ export class CreateRelationshipUseCase extends UseCase<CreateRelationshipRequest
             return Result.fail(RuntimeErrors.general.recordNotFound(RelationshipTemplate));
         }
 
+        const canSendRelationship = await this.relationshipsController.canSendRelationship({ creationContent: request.creationContent, template });
```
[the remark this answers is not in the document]
britsta: That's right, the ungrateful failure in the CreateRelationship use case is usually triggered by a failure to call the `sendRelationship` method. I saw a need to explicitly add the `canSendRelationship` method now to control the order of errors thrown because I want to prevent someone from optimizing their `creationContent` if a Relationship cannot be created anyway. Do you have an idea how to prevent the implicit redundant call of the `canSendRelationship` method?

packages/runtime/src/useCases/transport/relationships/CanCreateRelationship.ts
```diff
@@ -54,11 +59,48 @@ export class CanCreateRelationshipUseCase extends UseCase<CanCreateRelationshipR
...
+                if (responseToRequestOfTemplateValidationError) {
+                    return Result.ok({ isSuccess: false, code: responseToRequestOfTemplateValidationError.code, message: responseToRequestOfTemplateValidationError.message });
+                }
+            }
+        }
 
         return Result.ok({ isSuccess: true });
     }
+
+    private async validateResponseToRequestOfTemplate(templateId: RelationshipTemplateIdString, relationshipCreationContent: RelationshipCreationContent) {
```
Milena-Czierlinski, line 89: Might be nice to add return type.

packages/runtime/test/lib/testUtils.ts
```diff
@@ -394,13 +395,39 @@ export async function establishRelationship(transportServices1: TransportService
...
     transportServices1: TransportServices,
-    transportServices2: TransportServices,
+    runtimeServices2: TestRuntimeServices,
```
Milena-Czierlinski, line 398: I'd keep the input symmetrical, i.e. if you need the complete TestRuntimeService2 then also require TestRuntimeService1 and access its transport property within the function.

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -71,39 +90,277 @@ afterAll(() => serviceProvider.stop());
 describe("Can Create / Create Relationship", () => {
     let relationshipId: string;
 
-    test("load relationship Template in connector 2", async () => {
-        const template = await createTemplate(services1.transport);
+    describe("tests on creationContent of Relationship", () => {
+        let relationshipCreationContentDummy: RelationshipCreationContentJSON;
```
Milena-Czierlinski, line 94: Why did you call this Dummy? If it refers to the fact that it isn't possible to create a Relationship from it, I'd be more explicit about that in the name.

packages/runtime/test/transport/relationships.test.ts
```diff
@@ -44,12 +58,17 @@ const serviceProvider = new RuntimeServiceProvider();
...
+let services5: TestRuntimeServices;
 
 beforeAll(async () => {
-    const runtimeServices = await serviceProvider.launch(3, { enableRequestModule: true, enableDeciderModule: true, enableNotificationModule: true });
+    const runtimeServices = await serviceProvider.launch(4, { enableRequestModule: true, enableDeciderModule: true, enableNotificationModule: true });
+    const runtimeServicesWithDisabledRequestModule = await serviceProvider.launch(1, { enableRequestModule: false, enableDeciderModule: true, enableNotificationModule: true });
     services1 = runtimeServices[0];
     services2 = runtimeServices[1];
     services3 = runtimeServices[2];
+    services4 = runtimeServices[3];
```
Milena-Czierlinski, line 70: You don't need an an runtimeService, but can e.g. use `services3` instead.

Milena-Czierlinski (commented):
Just some small comments.

packages/runtime/test/transport/relationships.test.ts
britsta: Any suggestions? I would like to keep `relationshipCreationContent` in the name and that the name is not too long.

packages/runtime/src/useCases/transport/relationships/CreateRelationship.ts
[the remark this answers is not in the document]
Magnus-Kuhn: The order of errors is a good point - I don't see a way to realize this without the `CanSend`

packages/runtime/test/transport/relationships.test.ts
Magnus-Kuhn: `invalidRelationshipCreationContent` should do

packages/runtime/src/useCases/transport/relationships/CreateRelationship.ts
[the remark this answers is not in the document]
britsta: Update on avoiding redundant code: I put the type check of the `creationContent` of a Relationship in a separate file. However, the `validateResponseToRequestOfTemplate` method is still implemented twice, as this method needs access to `this.incomingRequestsController.getIncomingRequests({ ... })` and at first glance I didn't see a way to move such logic to a separate file. Any ideas or should I resolve? :)

packages/runtime/test/transport/relationships.test.ts
Milena-Czierlinski: But it the RelationshipCreationContent in fact invalid? Or is it itself totally fine? Then we could simply call it `relationshipCreationContent`.

britsta: I think it's more like an object with placeholders rather than it is an "invalid" object.

Originally I had named the variable `relationshipCreationContent`. Then I added "Dummy" to the variable name to indicate that it couldn't be used (due to a comment) because it contains placeholders (for the Request `id` or LocalAttribute `id`).

Magnus-Kuhn: `aRequestId` is no longer a valid requestId with this PR. Also those tests are long and to me it makes it clearer in the code that a wrong creationContent is being tested.

britsta: Fine with me :)

packages/runtime/src/useCases/transport/relationships/utility/validateCreationContentOfRelationship.ts
```diff
@@ -0,0 +1,64 @@
+import { Serializable } from "@js-soft/ts-serval";
+import { ApplicationError } from "@js-soft/ts-utils";
+import { IncomingRequestsController, LocalRequestStatus } from "@nmshd/consumption";
+import { ArbitraryRelationshipCreationContent, ArbitraryRelationshipTemplateContent, RelationshipCreationContent, RelationshipTemplateContent } from "@nmshd/content";
+import { RelationshipTemplate } from "@nmshd/transport";
+import { RelationshipTemplateIdString, RuntimeErrors } from "../../../common";
+
+export async function validateCreationContentOfRelationship(
+    incomingRequestsController: IncomingRequestsController,
```
Milena-Czierlinski, line 9: Passing the controller was suggested to me by GitHub Copilot. But we don't do this anywhere else. @jkoenig134 Is this legal? :D

jkoenig134: I don't see a reason why it shouldn't be ok!

PR #419, opened 2025-02-12
Enrich metadata of File with tags

sebbi08:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/runtime/src/useCases/transport/files/FileMapper.ts
+++ b/packages/runtime/src/useCases/transport/files/FileMapper.ts
@@ -24,6 +24,7 @@ export class FileMapper {
         return {
             id: file.id.toString(),
             filename: file.cache.filename,
+            tags: file.cache.tags ?? [],
             filesize: file.cache.filesize,
             createdAt: file.cache.createdAt.toString(),
             createdBy: file.cache.createdBy.toString(),
--- a/packages/runtime/src/useCases/transport/files/FileMapper.ts
+++ b/packages/runtime/src/useCases/transport/files/FileMapper.ts
@@ -24,7 +24,7 @@ export class FileMapper {
         return {
             id: file.id.toString(),
             filename: file.cache.filename,
-            tags: file.cache.tags ?? [],
+            tags: file.cache.tags,
             filesize: file.cache.filesize,
             createdAt: file.cache.createdAt.toString(),
             createdBy: file.cache.createdBy.toString(),
--- a/packages/runtime/src/useCases/transport/files/GetFiles.ts
+++ b/packages/runtime/src/useCases/transport/files/GetFiles.ts
@@ -18,6 +18,7 @@ export interface GetFilesQuery {
     mimetype?: string | string[];
     title?: string | string[];
     isOwn?: string | string[];
+    tags?: string | string[];
 }
 
 export interface GetFilesRequest {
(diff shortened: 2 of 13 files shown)


packages/runtime/test/transport/files.test.ts
jkoenig134, line 1: please also test sending this file and see if the recipient also sees the tags

jkoenig134 (changes requested):
only one small details, rest LGTM

jkoenig134 (dismissed):
LGTM

Milena-Czierlinski (changes requested):
I think FileDVO is missing. Would be great if you wait to merge this until I can make a proper review. :)

jkoenig134 (dismissed):

Milena-Czierlinski:
> Please add tags to -GetFilesQuery (similar to GetAttributesQuery) -FileDVO

@sebbi08 What about GetFilesQuery?

packages/runtime/src/useCases/transport/files/GetFiles.ts
```diff
@@ -55,7 +57,23 @@ export class GetFilesUseCase extends UseCase<GetFilesRequest, FileDTO[]> {
...
+                if (typeof input === "string") {
+                    query[`${nameof<FileDTO>((x) => x.tags)}`] = { $contains: input };
+                    return;
+                }
+                const allowedTags = [];
+                for (const tag of input) {
+                    const tagQuery = { [`${nameof<FileDTO>((x) => x.tags)}`]: { $contains: tag } };
+                    allowedTags.push(tagQuery);
+                }
+                query["$or"] = allowedTags;
```
jkoenig134, line 75: I don't really get why you are writing query.$or here. A list of strings should be automatically translated using the alias.

Milena-Czierlinski: It's the same like we have it for Attributes. We decided that querying multiple tags, all entries that contain at least one of those tags should be returned.

packages/runtime/src/useCases/transport/files/FileMapper.ts
```diff
@@ -24,6 +24,7 @@ export class FileMapper {
         return {
             id: file.id.toString(),
             filename: file.cache.filename,
+            tags: file.cache.tags ?? [],
```
Milena-Czierlinski, line 27: Should we rather allow tags to be undefined? I fear adding tags to FileDTO and FileDVO as a required property is a breaking change.

jkoenig134: IMO the app can handle this well, BUT it's just not required to make this required.

packages/runtime/src/useCases/transport/files/GetFiles.ts
jkoenig134: Hm, that's strange. We have to investigate if writing $or to the query could have some side effects as it wouldn't be possible to have multiple $or's in one query with this approach.

But for now this should be ok.

packages/runtime/src/useCases/transport/files/FileMapper.ts
sebbi08: I would leave it as it is because DVO and DTO are not used as input and therefor no breaking change anyway

Milena-Czierlinski: It would be for the App that uses the DVO, but @jkoenig134 said we could handle this. Apart from that, I remember that we used to have it like this for Attributes and then had to change it for some reason, that I currently can't remember. That's why now we have tags optional on the corresponding LocalAttributeDVOs. I prefer to have it similar for Files for consistency (and to avoid to possibly run into the problem that caused us to change it for Attributes).

Milena-Czierlinski (approved):

PR #421, opened 2025-02-17
Return all CanDecide errors simultaneously

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


Milena-Czierlinski:
@Siolto This fixes that not all errors are shown directly, trying to accept a Request with invalid parameters, which we observed in the App.

PR #426, opened 2025-02-19
Update licenses

Milena-Czierlinski:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


jkoenig134:
this date defines when the license was acquired - so we don't have to update this every year.

Milena-Czierlinski:
Ah I see. Is it then wrong now in the connector-tui?

PR #432, opened 2025-02-24
Add UseCases to delete Files, RelationshipTemplates and Tokens

Magnus-Kuhn:
# Readiness checklist

- [X] I added/updated tests.
- [X] I labeled the PR.
- [x] I self-reviewed the PR.


packages/runtime/src/useCases/transport/files/DeleteFile.ts
```diff
@@ -0,0 +1,35 @@
...
+        super(validator);
+    }
+
+    protected async executeInternal(request: DeleteFileRequest): Promise<Result<void>> {
+        const file = await this.fileController.getFile(CoreId.from(request.fileId));
+        if (!file) {
+            return Result.fail(RuntimeErrors.general.recordNotFound(File));
+        }
+
+        await this.fileController.deleteFile(file);
```
Milena-Czierlinski, line 31: Do we need a Datawallet sync after this? 

Same for other use cases.

jkoenig134: we do.

jkoenig134 (changes requested):
Tests missing for checking that those items are actually deleted from the backbone when they are owned by the deleting party.

Also as far as I saw tests are missing for deleting peer items.

Magnus-Kuhn:
Those tests are in transport, do you want to duplicate them in the runtime?

jkoenig134:
> Those tests are in transport, do you want to duplicate them in the runtime?

let me think about it, but I think we should have them at both places. Other opinions @Milena-Czierlinski?

packages/transport/src/modules/tokens/TokenController.ts
```diff
@@ -296,7 +296,10 @@ export class TokenController extends TransportController {
     }
 
     public async delete(token: Token): Promise<void> {
-        if (token.isOwn) await this.client.deleteToken(token.id.toString());
+        if (token.isOwn) {
+            const response = await this.client.deleteToken(token.id.toString());
+            if (response.isError) throw response.error;
+        }
```
jkoenig134, line 302: make sure to catch these errors during decomposition, we shouldn't stop the decomposition b/c of an error occurs here

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
```diff
@@ -112,8 +112,10 @@ export class RelationshipTemplateController extends TransportController {
     }
 
     public async deleteRelationshipTemplate(template: RelationshipTemplate): Promise<void> {
-        const response = await this.client.deleteRelationshipTemplate(template.id.toString());
-        if (response.isError) throw response.error;
+        if (template.isOwn) {
+            const response = await this.client.deleteRelationshipTemplate(template.id.toString());
+            if (response.isError) throw response.error;
+        }
```
jkoenig134, line 118: make sure to catch these errors during decomposition, we shouldn't stop the decomposition b/c of an error occurs here

jkoenig134 (changes requested):

packages/transport/test/modules/files/FileController.test.ts
```diff
@@ -123,4 +123,35 @@ describe("FileController", function () {
...
+        test("should delete own file locally and from the backbone", async function () {
+            await sender.files.deleteFile(sentFile);
+            const fileOnBackbone = await recipient.files.fetchCaches([sentFile.id]);
+            expect(fileOnBackbone).toHaveLength(0);
+
+            const localFile = await sender.files.getFile(sentFile.id);
+            expect(localFile).toBeUndefined();
+        });
+
+        test("should delete a not owned file only locally", async function () {
```
jkoenig134, line 148: I don't like the wording `not owned`, please use `peer owned` everywhere.

Milena-Czierlinski:
> let me think about it, but I think we should have them at both places. Other opinions @Milena-Czierlinski?

In my opinion it is sufficient if there are tests in transport and that the runtime only tests the happy path(s) and validation errors from the use case.

packages/transport/src/modules/relationshipTemplates/RelationshipTemplateController.ts
jkoenig134: even better :)

packages/transport/test/modules/relationshipTemplates/RelationshipTemplateController.test.ts
```diff
@@ -248,4 +248,34 @@ describe("RelationshipTemplateController", function () {
...
+        test("should delete own relationshipTemplate locally and from the backbone", async function () {
+            await sender.relationshipTemplates.deleteRelationshipTemplate(ownTemplate);
+            const templateOnBackbone = await recipient.relationshipTemplates.fetchCaches([ownTemplate.id]);
+            expect(templateOnBackbone).toHaveLength(0);
+
+            const localTemplate = await sender.relationshipTemplates.getRelationshipTemplate(ownTemplate.id);
+            expect(localTemplate).toBeUndefined();
+        });
+
+        test("should delete a peer owned relationshipTemplate only locally", async function () {
```
jkoenig134, line 272: please write `RelationshipTemplate` or `relationship template` instead of `relationshipTemplate`

jkoenig134 (approved):
LGTM

PR #433, opened 2025-02-24
Delete incoming requests

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
+++ b/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
@@ -340,9 +340,9 @@ class Requests {
         return new CoreError("error.consumption.requests.cannotCreateRequestWithExpirationDateInPast", "You cannot create a Request with an expiration date that is in the past.");
     }
 
-    public cannotDeleteIncomingRequestThatIsNotExpired(id: string, status: string) {
+    public canOnlyDeleteIncomingRequestThatIsExpired(id: string, status: string) {
         return new CoreError(
-            "error.consumption.requests.cannotDeleteIncomingRequestThatIsNotExpired",
+            "error.consumption.requests.canOnlyDeleteIncomingRequestThatIsExpired",
             `The incoming Request '${id}' is in status '${status}'. At the moment, you can only delete incoming Requests that are expired.`
         );
     }
--- a/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
+++ b/packages/consumption/src/consumption/ConsumptionCoreErrors.ts
@@ -340,6 +340,13 @@ class Requests {
         return new CoreError("error.consumption.requests.cannotCreateRequestWithExpirationDateInPast", "You cannot create a Request with an expiration date that is in the past.");
     }
 
+    public cannotDeleteIncomingRequestThatIsNotExpired(id: string, status: string) {
+        return new CoreError(
+            "error.consumption.requests.cannotDeleteIncomingRequestThatIsNotExpired",
+            `The incoming Request '${id}' is in status '${status}'. At the moment, you can only delete incoming Requests that are expired.`
+        );
+    }
+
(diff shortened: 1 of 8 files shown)


packages/consumption/src/consumption/ConsumptionCoreErrors.ts
```diff
@@ -340,6 +340,13 @@ class Requests {
         return new CoreError("error.consumption.requests.cannotCreateRequestWithExpirationDateInPast", "You cannot create a Request with an expiration date that is in the past.");
     }
 
+    public cannotDeleteIncomingRequestThatIsNotExpired(id: string, status: string) {
```
Milena-Czierlinski, line 343:
```suggestion
    public canOnlyDeleteIncomingRequestThatIsExpired(id: string, status: string) {
```
Without the negations it's easier to read, but I understand why you used them. So, I'm fine with both.

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -451,6 +451,14 @@ export class IncomingRequestsController extends ConsumptionBaseController {
         await this.localRequests.update(requestDoc, request);
     }
 
+    public async delete(request: LocalRequest): Promise<void> {
+        if (request.status !== LocalRequestStatus.Expired) {
+            throw ConsumptionCoreErrors.requests.cannotDeleteIncomingRequestThatIsNotExpired(request.id.toString(), request.status);
+        }
```
Milena-Czierlinski, line 457: I would have expected this in the runtime use case.

jkoenig134: All request validations are done in consumption, just sticking with the default here.

Milena-Czierlinski: It's just that here we make the high-level choice to only delete expired Requests.

jkoenig134: I have a very differnt opinion on this, it's not high level - it can have interferences in every level so we should forbid it at lowest level possible.

packages/runtime/test/consumption/requests.test.ts
```diff
@@ -673,6 +673,30 @@ describe("Requests", () => {
...
+            const request = (await exchangeTemplateAndReceiverRequiresManualDecision(sRuntimeServices, rRuntimeServices, templateContent)).request;
+
+            await expect(rConsumptionServices.incomingRequests.delete({ requestId: request.id })).resolves.toBeAnError(
+                "is in status 'ManualDecisionRequired'. At the moment, you can only delete incoming Requests that are expired.",
+                "error.consumption.requests.canOnlyDeleteIncomingRequestThatIsExpired"
+            );
+        });
+
+        test("can delete a Request when it is expired", async () => {
+            const request = (await exchangeTemplateAndReceiverRequiresManualDecision(sRuntimeServices, rRuntimeServices, templateContent, DateTime.utc().plus({ seconds: 3 })))
```
Milena-Czierlinski, line 686: I think in other places were we test expired Requests we only add two seconds. Have you checked if that also works here and in the consumption test?

jkoenig134: it's copied from another test where 3 secs are used. I'd rather not trigger another build only for that.

Milena-Czierlinski: Ah okay, then we seem to not be consistent there. 😅

Milena-Czierlinski (approved):

packages/runtime/test/consumption/requests.test.ts
jkoenig134: yeah :/

PR #436, opened 2025-02-25
Add validation for `tags` of Attributes

sebbi08:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -355,6 +363,14 @@ export class AttributesController extends ConsumptionBaseController {
     }
 
     public async createSharedLocalAttribute(params: ICreateSharedLocalAttributeParams): Promise<LocalAttribute> {
+        const parsedParams = CreateSharedLocalAttributeParams.from(params);
+        if ("tags" in parsedParams.content && parsedParams.content.tags) {
```
Milena-Czierlinski, line 367: Why is that first part of the condition necessary? Similar in other places.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -355,6 +363,14 @@ export class AttributesController extends ConsumptionBaseController {
...
+            const tagValidationResult = await this.validateTags(parsedParams.content.tags, parsedParams.content.toJSON().value["@type"]);
+            if (tagValidationResult.isError()) {
+                throw tagValidationResult.error;
+            }
+        }
+
```
Milena-Czierlinski, line 373: I'm not sure if we want to have the validation here in general. This would be executed if a peer shared IdentityAttribute is created. Do we actually want to throw an error, if a peer shares an Attribute with an invalid tag or do we only want to validate creating an own RepositoryAttribute. @jkoenig134 Wdym? 

This might also have implications on the place of the succession validation. Furthermore, it must be handled analogously in the RequestItemProcessors.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1334,4 +1357,52 @@ export class AttributesController extends ConsumptionBaseController {
...
+            if (invalidTags.length === 1) {
+                return ValidationResult.error(ConsumptionCoreErrors.attributes.invalidTag(invalidTags[0]));
+            }
+            return ValidationResult.error(ConsumptionCoreErrors.attributes.invalidTags(invalidTags));
+        }
+
+        return ValidationResult.success();
+    }
+
+    public async validateTag(tag: string, attributeValueType: string): Promise<boolean> {
```
Milena-Czierlinski, line 1378: Personally, I find this function a bit hard to comprehend. I'll give it a thought if I can come up with an easier refactoring.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1334,4 +1357,52 @@ export class AttributesController extends ConsumptionBaseController {
...
+        for (const tag of tags) {
+            if (!(await this.validateTag(tag, attributeValueType))) {
+                invalidTags.push(tag);
+            }
+        }
+        if (invalidTags.length > 0) {
+            if (invalidTags.length === 1) {
+                return ValidationResult.error(ConsumptionCoreErrors.attributes.invalidTag(invalidTags[0]));
+            }
+            return ValidationResult.error(ConsumptionCoreErrors.attributes.invalidTags(invalidTags));
```
Milena-Czierlinski, line 1372: I don't think we need to differentiate these cases here and have two different errors. From my point of view it is okay to only have an error message that uses the plural, even if it references a list with only one element.

packages/consumption/src/consumption/ConsumptionCoreErrors.ts
```diff
@@ -266,6 +266,14 @@ class Attributes {
     public setDefaultRepositoryAttributesIsDisabled() {
         return new CoreError("error.consumption.attributes.setDefaultRepositoryAttributesIsDisabled", "Setting default RepositoryAttributes is disabled for this Account.");
     }
+
+    public invalidTag(tag: string): ApplicationError {
+        return new ApplicationError("error.consumption.attributes.invalidTag", `The tag '${tag}' is invalid.`);
+    }
+
+    public invalidTags(tags: string[]): ApplicationError {
+        return new ApplicationError("error.consumption.attributes.invalidTags", `The tags '${tags.join(", ")}' is invalid.`);
```
Milena-Czierlinski, line 275:
```suggestion
        return new ApplicationError("error.consumption.attributes.invalidTags", `The tags '${tags.join(", ")}' are invalid.`);
```
Also, I would state for which value type they are invalid.

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
```diff
@@ -97,6 +97,10 @@ export class CreateAttributeRequestItemProcessor extends GenericRequestItemProce
             }
         }
 
+        if (requestItem.attribute instanceof IdentityAttribute && requestItem.attribute.tags) {
+            return await this.consumptionController.attributes.validateTags(requestItem.attribute.tags, requestItem.attribute.toJSON().value["@type"]);
```
Milena-Czierlinski, line 101: Please return an `invalidAcceptParameters` error like e.g. in the case above, such that the person who reads it knows that they cannot accept the RequestItem since it would lead to an Attribute with invalid tags. 

Same in other places.

Milena-Czierlinski (commented):
I haven't checked the non-trivial tests, yet, since they might change.

packages/consumption/src/consumption/ConsumptionCoreErrors.ts
britsta, line 275: The current implementation would lead to the individual tags not each being enclosed with single ticks (`'tag1', 'tag2'`), but to all tags being enclosed together with one pair of single ticks (`'tag1, tag2'`), which is why I suggest the following change.

```suggestion
        return new ApplicationError("error.consumption.attributes.invalidTags", `The tags ${tags.map((tag) => `'${tag}'`).join(", ")} are invalid.`);
```

packages/consumption/src/modules/attributes/AttributesController.ts
britsta: I also think that one error code is sufficient. To keep it linguistically correct, we can then speak of "tag(s)" in the error message, as we do with the "recipient(s)" in the error message for error code `error.runtime.messages.peerIsInDeletion`, for example.

packages/consumption/src/consumption/ConsumptionCoreErrors.ts
Magnus-Kuhn: also `'${tags.join("', '")}'`

packages/consumption/src/modules/attributes/AttributesController.ts
Magnus-Kuhn: I'm also on it

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -3246,4 +3249,65 @@ describe("AttributesController", function () {
...
+            expect(await consumptionController.attributes.validateTag("emergency+%+first", "PhoneNumber")).toBe(true);
+            expect(await consumptionController.attributes.validateTag("emergency+%+second", "PhoneNumber")).toBe(true);
+            expect(await consumptionController.attributes.validateTag("x+%+my+%+custom+%+tag", "PhoneNumber")).toBe(true);
+            expect(await consumptionController.attributes.validateTag("X+%+my+%+custom+%+tag", "PhoneNumber")).toBe(true);
+        });
+
+        test("should validate invalid tags", async function () {
+            expect(await consumptionController.attributes.validateTag("privates", "PhoneNumber")).toBe(false);
+            expect(await consumptionController.attributes.validateTag("emergency", "PhoneNumber")).toBe(false);
+            expect(await consumptionController.attributes.validateTag("emergency+%+nonexisting", "PhoneNumber")).toBe(false);
```
britsta, line 3309: "Nonexisting" is not a standard English word, "nonexistent" could be used.

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -3246,4 +3249,65 @@ describe("AttributesController", function () {
...
+            expect(await consumptionController.attributes.validateTag("emergency+%+first+%+nonexisting", "PhoneNumber")).toBe(false);
```
britsta, line 3310: "Nonexisting" is not a standard English word, "nonexistent" could be used.

packages/consumption/test/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.test.ts
```diff
@@ -312,6 +312,20 @@ describe("CreateAttributeRequestItemProcessor", function () {
                     "This CreateAttributeRequestItem cannot be accepted as the provided RelationshipAttribute cannot be created because there is already a RelationshipAttribute in the context of this Relationship with the same key 'aKey', owner and value type."
             });
         });
+
+        test("cannot accept because the attribute has invalid tags", async function () {
```
britsta, line 316: Please also add a test for multiple invalid tags to check the correct formatting of the error message.

packages/consumption/test/modules/requests/itemProcessors/shareAttribute/ShareAttributeRequestItemProcessor.test.ts
```diff
@@ -1037,6 +1039,93 @@ describe("ShareAttributeRequestItemProcessor", function () {
         });
     });
 
+    describe("canAccept", function () {
+        test("returns an success when sharing a valid attribute", async function () {
```
britsta, line 1043:
```suggestion
        test("returns a success when sharing a valid attribute", async function () {
```

packages/consumption/src/modules/attributes/AttributesController.ts
Magnus-Kuhn:
```suggestion
        const tagCollection = await this.getAttributeTagCollection();
        let allowedTagsForValueType = tagCollection.tagsForAttributeValueTypes[attributeValueType];

        const tagParts = tag.split("+%+");
        try {
            const allowedTagsForLastPart = tagParts.slice(0, -1).reduce((allowedTags, tag) => allowedTags[tag].children!, allowedTagsForValueType);
            if (allowedTagsForLastPart[tagParts.at(-1)!].children) return false;
        } catch (e) {
            return false;
        }
```

packages/consumption/src/modules/attributes/AttributesController.ts
sebbi08: Yes it is because `.content' is `IdentityAttribute | RelationshipAttribute' and `RelationshipAttribute' does not have tags but `IdentityAttribute', I have changed that to an instance of to be clearer.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1334,4 +1357,52 @@ export class AttributesController extends ConsumptionBaseController {
...
+        }
+
+        return ValidationResult.success();
+    }
+
+    public async validateTag(tag: string, attributeValueType: string): Promise<boolean> {
+        if (tag.toLowerCase().startsWith("x+%+")) {
```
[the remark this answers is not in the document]
sebbi08: Because we defined that the custom tag start "x" should be case-insensitive. So "x" and "X" are valid starts.

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -3246,4 +3249,65 @@ describe("AttributesController", function () {
         expect(ownAttribute).toBeUndefined();
         expect(peerAttribute).toBeUndefined();
     });
+
+    describe("validate tags", function () {
+        let mockedAttributesController: AttributesController;
+        /* eslint-disable @typescript-eslint/naming-convention */
```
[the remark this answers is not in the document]
sebbi08: These comments are actually needed because some parts of this JSON do not conform to our naming convention, but are needed in this way. For example, `PhoneNumber` starts with an uppercase letter, which is not allowed, but because it is the attribute type, it is needed exactly as it is.

packages/consumption/test/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.test.ts
```diff
@@ -545,6 +547,70 @@ describe("ReadAttributeRequestItemProcessor", function () {
...
+            const canAcceptWithExistingAttributeResult = await processor.canAccept(
+                requestItem,
+                {
+                    accept: true,
+                    existingAttributeId: existingAttribute.id.toString()
+                },
+                request
+            );
+
+            expect(canAcceptWithExistingAttributeResult).errorValidationResult({
```
Magnus-Kuhn, line 592: more error message checking

packages/consumption/test/modules/requests/itemProcessors/shareAttribute/ShareAttributeRequestItemProcessor.test.ts
```diff
@@ -1037,6 +1039,95 @@ describe("ShareAttributeRequestItemProcessor", function () {
...
+
+            const canAcceptWithExistingAttributeResult = await processor.canAccept(
+                requestItem,
+                {
+                    accept: true
+                },
+                request
+            );
+
+            expect(canAcceptWithExistingAttributeResult).errorValidationResult({
```
Magnus-Kuhn, line 1125: and another error message check

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -3246,4 +3249,65 @@ describe("AttributesController", function () {
...
+        test("should validate valid tags", async function () {
+            expect(await consumptionController.attributes["validateTag"]("private", "PhoneNumber")).toBe(true);
+            expect(await consumptionController.attributes["validateTag"]("emergency+%+first", "PhoneNumber")).toBe(true);
+            expect(await consumptionController.attributes["validateTag"]("emergency+%+second", "PhoneNumber")).toBe(true);
+            expect(await consumptionController.attributes["validateTag"]("x+%+my+%+custom+%+tag", "PhoneNumber")).toBe(true);
+            expect(await consumptionController.attributes["validateTag"]("X+%+my+%+custom+%+tag", "PhoneNumber")).toBe(true);
+        });
+
+        test("should validate invalid tags", async function () {
+            expect(await consumptionController.attributes["validateTag"]("privates", "PhoneNumber")).toBe(false);
```
Magnus-Kuhn, line 3307: `nonexistent` instead of `privates`

packages/consumption/src/modules/attributes/AttributesController.ts
Magnus-Kuhn: what about it?

sebbi08: To be honest, I don't find this more readable because here all the "steps" are done in one go. @Milena-Czierlinski what do you think.

packages/consumption/test/modules/attributes/AttributesController.test.ts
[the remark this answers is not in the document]
britsta: How can I detect that these comments are needed? I have removed the comments locally and do not get an error for the uppercase `PhoneNumber` in this JSON.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1334,4 +1357,49 @@ export class AttributesController extends ConsumptionBaseController {
...
     }
+
+    public async validateTags(tags: string[], attributeValueType: string): Promise<ValidationResult> {
+        const invalidTags = [];
+        for (const tag of tags) {
+            if (!(await this.validateTag(tag, attributeValueType))) {
+                invalidTags.push(tag);
+            }
+        }
+        if (invalidTags.length > 0) {
```
britsta, line 1368: Please insert an empty line before this `if` statement.

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -3246,4 +3249,66 @@ describe("AttributesController", function () {
...
+        /* eslint-enable @typescript-eslint/naming-convention */
+
+        beforeAll(function () {
+            mockedAttributesController = spy(consumptionController.attributes);
+            when(mockedAttributesController.getAttributeTagCollection()).thenResolve(AttributeTagCollection.from(mockedTagCollection));
+        });
+        afterAll(function () {
+            reset(mockedAttributesController);
+        });
+        test("should validate valid tags", async function () {
```
britsta, line 3298: Please add an empty line before this test.

packages/consumption/test/modules/attributes/AttributesController.test.ts
```diff
@@ -3246,4 +3249,66 @@ describe("AttributesController", function () {
...
+                }
+            }
+        };
+        /* eslint-enable @typescript-eslint/naming-convention */
+
+        beforeAll(function () {
+            mockedAttributesController = spy(consumptionController.attributes);
+            when(mockedAttributesController.getAttributeTagCollection()).thenResolve(AttributeTagCollection.from(mockedTagCollection));
+        });
+        afterAll(function () {
```
britsta, line 3295: Please add an empty line before the `afterAll` block.

packages/consumption/test/modules/attributes/AttributesController.test.ts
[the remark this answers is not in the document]
britsta: Edit: As @sebbi08 told me, run `npm i` locally was necessary to detect that these comments are needed.

packages/consumption/src/modules/attributes/AttributesController.ts
Magnus-Kuhn: Looks good now, there is only one other comment left, otherwise I'd say LGTM

packages/consumption/src/modules/attributes/AttributesController.ts
jkoenig134: yep

jkoenig134 (approved):
LGTM, @sebbi08 please wait for @Milena-Czierlinski and @britsta approval :)

packages/consumption/test/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.test.ts
```diff
@@ -314,6 +314,34 @@ describe("CreateAttributeRequestItemProcessor", function () {
...
+            await Given.aRequestItemWithAnIdentityAttribute({
+                tags: ["tag1"],
+                attributeOwner: TestIdentity.CURRENT_IDENTITY
+            });
+
+            await When.iCallCanAccept();
+
+            await Then.theCanAcceptResultShouldBeAnErrorWith({
+                code: "error.consumption.requests.invalidRequestItem",
+                message: /The provided IdentityAttribute is invalid: The tag\(s\) 'tag1' is\/are invalid\./
```
Milena-Czierlinski, line 328: Usually, we write the error message as a string. Same in other places.

Milena-Czierlinski (commented):
Just one small comment, maybe not important.

Btw should this be marked as a breaking change?

packages/consumption/test/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.test.ts
sebbi08: Yea but here it is a core error because the other side somehow sneeked in an invalid attribute. And those core errors have time stamps therefore we need the regex

jkoenig134: I dont think so, the regex should be generated under the hood.

sebbi08: Ah yea there was a problem with the ( and ) in the string that's why I made it a regex

jkoenig134: IMO only having the error message for plural `The tags 'tag1' are invalid.` would be also ok and way easier to read.

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
```diff
@@ -97,6 +97,11 @@ export class CreateAttributeRequestItemProcessor extends GenericRequestItemProce
             }
         }
 
+        const tagValidationResult = await this.consumptionController.attributes.validateTags(requestItem.attribute);
+        if (tagValidationResult.isError()) {
+            return ValidationResult.error(ConsumptionCoreErrors.requests.invalidRequestItem(`${tagValidationResult.error.message}`));
```
jkoenig134, line 102:
```suggestion
            return ValidationResult.error(ConsumptionCoreErrors.requests.invalidRequestItem(tagValidationResult.error.message));
```

packages/consumption/src/modules/requests/itemProcessors/proposeAttribute/ProposeAttributeRequestItemProcessor.ts
```diff
@@ -197,6 +197,11 @@ export class ProposeAttributeRequestItemProcessor extends GenericRequestItemProc
             }
         }
 
+        const tagValidationResult = await this.consumptionController.attributes.validateTags(attribute);
+        if (tagValidationResult.isError()) {
+            return ValidationResult.error(ConsumptionCoreErrors.requests.invalidAcceptParameters(`${tagValidationResult.error.message}`));
```
jkoenig134, line 202: see above

packages/consumption/src/modules/requests/itemProcessors/shareAttribute/ShareAttributeRequestItemProcessor.ts
```diff
@@ -176,6 +176,15 @@ export class ShareAttributeRequestItemProcessor extends GenericRequestItemProces
         return ValidationResult.success();
     }
 
+    public override async canAccept(requestItem: ShareAttributeRequestItem, _params: AcceptRequestItemParametersJSON, _requestInfo: LocalRequestInfo): Promise<ValidationResult> {
+        const tagValidationResult = await this.consumptionController.attributes.validateTags(requestItem.attribute);
+        if (tagValidationResult.isError()) {
+            return ValidationResult.error(ConsumptionCoreErrors.requests.invalidRequestItem(`${tagValidationResult.error.message}`));
```
jkoenig134, line 182: see above

packages/consumption/src/modules/requests/itemProcessors/createAttribute/CreateAttributeRequestItemProcessor.ts
```diff
@@ -97,6 +102,11 @@ export class CreateAttributeRequestItemProcessor extends GenericRequestItemProce
             }
         }
 
+        const tagValidationResult = await this.consumptionController.attributes.validateTags(requestItem.attribute);
```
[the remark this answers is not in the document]
britsta: I'm actually not a fan of this. Early return of success makes sense in most cases, but is IMO error-prone in the `canCreateOutgoingRequestItem` methods of the RequestItemProcessors. Below the block for IdentityAttribute or RelationshipAttribute validation there can still be general Attribute validation and below that there should be the return of success. Otherwise, when adding general Attribute validation later it could be difficult to identify in which blocks there has already been an early return of success (I had problems in this PR because of such an early return).

Milena-Czierlinski (approved):

PR #439, opened 2025-02-27
Notifying about RepositoryAttribute succession isn't possible if additional own shared IdentityAttribute with deletionInfo exsists

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

Example: An Identity has a RepositoryAttribute and shared it with a peer, but the peer deleted the peer shared Attribute. As a result, the Identity has an own shared IdentityAttribute with `deletionInfo.deletionStatus = DeletedByPeer`. Then, the peer asks for that RepositoryAttribute again and the Identity shares it again. Now, the Identity has two own shared IdentityAttributes, one with deletionInfo and one without it. If the Identity then succeeds the RepositoryAttribute and wants to notify about the succession, it is necessary to chose the own shared IdentityAttribute without deletionInfo as predecessor.

### DIFF
--- a/packages/runtime/src/useCases/common/RuntimeErrors.ts
+++ b/packages/runtime/src/useCases/common/RuntimeErrors.ts
@@ -228,13 +228,6 @@ class Attributes {
         );
     }
 
-    public cannotSucceedAttributesWithDeletionInfo(ownSharedIdentityAttributeIds: CoreId[] | string[]): ApplicationError {
-        return new ApplicationError(
-            "error.runtime.attributes.cannotSucceedAttributesWithDeletionInfo",
-            `The own shared IdentityAttribute predecessor(s) ${ownSharedIdentityAttributeIds.map((ownSharedIdentityAttributeId) => `'${ownSharedIdentityAttributeId.toString()}'`).join(", ")} can't be succeeded, since they have a deletionInfo.`
-        );
-    }
-
     public isNotOwnSharedAttribute(attributeId: CoreId | string): ApplicationError {
         return new ApplicationError("error.runtime.attributes.isNotOwnSharedAttribute", `Attribute '${attributeId.toString()}' is not an own shared Attribute.`);
     }
@@ -254,6 +247,13 @@ class Attributes {
         return new ApplicationError(
             "error.runtime.attributes.hasSuccessor",
             `Attribute '${predecessor.id.toString()}' already has a successor ${predecessor.succeededBy?.toString()}.`
+        );
+    }
+
+    public cannotSucceedAttributesWithDeletionInfo(ownSharedIdentityAttributeIds: CoreId[] | string[]): ApplicationError {
+        return new ApplicationError(
+            "error.runtime.attributes.cannotSucceedAttributesWithDeletionInfo",
+            `The own shared IdentityAttribute predecessor(s) ${ownSharedIdentityAttributeIds.map((ownSharedIdentityAttributeId) => `'${ownSharedIdentityAttributeId.toString()}'`).join(", ")} can't be succeeded due to set deletionInfo.`
         );
     }
 
--- a/packages/runtime/src/useCases/common/RuntimeErrors.ts
+++ b/packages/runtime/src/useCases/common/RuntimeErrors.ts
@@ -228,6 +228,13 @@ class Attributes {
         );
     }
 
+    public cannotSucceedAttributesWithDeletionInfo(ownSharedIdentityAttributeIds: CoreId[] | string[]): ApplicationError {
+        return new ApplicationError(
+            "error.runtime.attributes.cannotSucceedAttributesWithDeletionInfo",
+            `The own shared IdentityAttribute predecessor(s) ${ownSharedIdentityAttributeIds.map((ownSharedIdentityAttributeId) => `'${ownSharedIdentityAttributeId.toString()}'`).join(", ")} can't be succeeded, since they have a deletionInfo.`
+        );
+    }
+
     public isNotOwnSharedAttribute(attributeId: CoreId | string): ApplicationError {
         return new ApplicationError("error.runtime.attributes.isNotOwnSharedAttribute", `Attribute '${attributeId.toString()}' is not an own shared Attribute.`);
     }
--- a/packages/runtime/src/useCases/consumption/attributes/NotifyPeerAboutRepositoryAttributeSuccession.ts
+++ b/packages/runtime/src/useCases/consumption/attributes/NotifyPeerAboutRepositoryAttributeSuccession.ts
@@ -54,18 +54,11 @@ export class NotifyPeerAboutRepositoryAttributeSuccessionUseCase extends UseCase
             return Result.fail(RuntimeErrors.attributes.noPreviousVersionOfRepositoryAttributeHasBeenSharedWithPeerBefore(repositoryAttributeSuccessorId, request.peer));
         }
 
-        let ownSharedIdentityAttributePredecessor: LocalAttribute | undefined;
-        for (const candidatePredecessor of candidatePredecessors) {
-            if (
-                !(
-                    candidatePredecessor.deletionInfo?.deletionStatus === LocalAttributeDeletionStatus.DeletedByPeer ||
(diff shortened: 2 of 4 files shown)


packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -228,6 +228,13 @@ class Attributes {
         );
     }
 
+    public cannotSucceedAttributesWithDeletionInfo(ownSharedIdentityAttributeIds: CoreId[] | string[]): ApplicationError {
+        return new ApplicationError(
+            "error.runtime.attributes.cannotSucceedAttributesWithDeletionInfo",
+            `The own shared IdentityAttribute predecessor(s) ${ownSharedIdentityAttributeIds.map((ownSharedIdentityAttributeId) => `'${ownSharedIdentityAttributeId.toString()}'`).join(", ")} can't be succeeded, since they have a deletionInfo.`
```
britsta, line 234: If it is only a single own shared IdentityAttribute predecessor, "they have" in the second part of the sentence is unfitting.

packages/runtime/src/useCases/common/RuntimeErrors.ts
```diff
@@ -228,6 +228,13 @@ class Attributes {
         );
     }
 
+    public cannotSucceedAttributesWithDeletionInfo(ownSharedIdentityAttributeIds: CoreId[] | string[]): ApplicationError {
```
britsta, line 231: Please place the error a little further down with the other errors for succession and deletion, for example after the `error.runtime.attributes.hasSuccessor` error code.

packages/runtime/src/useCases/consumption/attributes/NotifyPeerAboutRepositoryAttributeSuccession.ts
```diff
@@ -54,11 +54,29 @@ export class NotifyPeerAboutRepositoryAttributeSuccessionUseCase extends UseCase
...
 
-        if (candidatePredecessors[0].shareInfo?.sourceAttribute?.toString() === request.attributeId) {
-            return Result.fail(RuntimeErrors.attributes.repositoryAttributeHasAlreadyBeenSharedWithPeer(request.attributeId, request.peer, candidatePredecessors[0].id));
+        let ownSharedIdentityAttributePredecessor: LocalAttribute | undefined;
+        for (const candidatePredecessor of candidatePredecessors) {
+            if (
+                !(
+                    candidatePredecessor.deletionInfo?.deletionStatus === LocalAttributeDeletionStatus.DeletedByPeer ||
+                    candidatePredecessor.deletionInfo?.deletionStatus === LocalAttributeDeletionStatus.ToBeDeletedByPeer ||
+                    candidatePredecessor.deletionInfo?.deletionStatus === LocalAttributeDeletionStatus.DeletionRequestSent
```
britsta, line 63: That the `deletionStatus` must not be `"DeletionRequestSent"` is inconsistent with the handling of the AttributeSuccessionAcceptResponseItem in the ReadAttributeRequestItemProcessor. Is this intentional? I guess this is because the method `validateAttributeSuccessionCommon` in the AttributesController also uses the implementation chosen here, but I was still wondering why there is a difference to the handling of the AttributeSuccessionAcceptResponseItem.

packages/runtime/src/useCases/consumption/attributes/NotifyPeerAboutRepositoryAttributeSuccession.ts
```diff
@@ -54,11 +54,29 @@ export class NotifyPeerAboutRepositoryAttributeSuccessionUseCase extends UseCase
             return Result.fail(RuntimeErrors.attributes.noPreviousVersionOfRepositoryAttributeHasBeenSharedWithPeerBefore(repositoryAttributeSuccessorId, request.peer));
         }
 
-        if (candidatePredecessors[0].shareInfo?.sourceAttribute?.toString() === request.attributeId) {
-            return Result.fail(RuntimeErrors.attributes.repositoryAttributeHasAlreadyBeenSharedWithPeer(request.attributeId, request.peer, candidatePredecessors[0].id));
+        let ownSharedIdentityAttributePredecessor: LocalAttribute | undefined;
```
britsta, line 57: I would like the variable name to indicate that `"DeletedByPeer"`, `"ToBeDeletedByPeer"` or `"DeletionRequestSent"` as `deletionStatus` has been excluded, e.g. `candidatePredecessorWithoutDeletionInfo`. However, this name would not be entirely correct if `"DeletionRequestRejected"` is allowed as a `deletionStatus` (however, this also applies to the error code). Maybe you have an idea.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -1696,6 +1696,31 @@ describe(NotifyPeerAboutRepositoryAttributeSuccessionUseCase.name, () => {
         expect(ownSharedIdentityAttributeVersion2.succeeds).toStrictEqual(ownSharedIdentityAttributeVersion0.id);
     });
 
+    test("should allow to notify about successor if the predecessor was deleted by peer but additional predecessor exists", async () => {
```
britsta, line 1699: I'd expect this test after the `"should throw if the predecessor was deleted by peer"` test.

packages/runtime/src/useCases/common/RuntimeErrors.ts
Milena-Czierlinski: What do you suggest instead, without impeding the readability?

packages/runtime/src/useCases/consumption/attributes/NotifyPeerAboutRepositoryAttributeSuccession.ts
Milena-Czierlinski: I see the same problem like you regarding the allowed deletionStatus. The advantage of the current name, besides being (almost) not overly long is that we do not need another assignment after all checks have passed. However, if this topic is really important to you, please make another suggestion.

packages/runtime/src/useCases/consumption/attributes/NotifyPeerAboutRepositoryAttributeSuccession.ts
Milena-Czierlinski: Yes, I wanted to be consistent with validateAttributeSuccessionCommon. We can have another discussion about the handling in the ReadAttributeRequestItemProcessor, but let's don't do this via text. :D

packages/runtime/test/consumption/attributes.test.ts
Milena-Czierlinski: We have all happy tests in the beginning here and I would like to stay consistent with them.

packages/runtime/src/useCases/common/RuntimeErrors.ts
britsta:
```suggestion
            `The own shared IdentityAttribute predecessor(s) ${ownSharedIdentityAttributeIds.map((ownSharedIdentityAttributeId) => `'${ownSharedIdentityAttributeId.toString()}'`).join(", ")} can't be succeeded due to set deletionInfo.`
```

packages/runtime/src/useCases/consumption/attributes/NotifyPeerAboutRepositoryAttributeSuccession.ts
britsta: It's not so important if we don't have a better idea.

packages/runtime/src/useCases/consumption/attributes/NotifyPeerAboutRepositoryAttributeSuccession.ts
Milena-Czierlinski: We decided that it makes sense to allow to succeed Attributes with deletionStatus DeletionRequestSent. The consumption method will be adjusted accordingly.

Magnus-Kuhn (approved):

britsta (approved):

PR #442, opened 2025-03-03
Make `title` of AuthenticationRequestItem mandatory

britsta:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

The `title` of an AuthenticationRequestItem should now be mandatory so that the receiver always gets information about what is being authenticated for at least through the `title`.


Magnus-Kuhn (approved):

packages/runtime/test/modules/DeciderModule.test.ts
```diff
@@ -171,7 +171,7 @@ describe("DeciderModule", () => {
                 receivedRequest: {
                     "@type": "Request",
                     items: [
-                        { "@type": "AuthenticationRequestItem", mustBeAccepted: false },
+                        { "@type": "AuthenticationRequestItem", mustBeAccepted: false, title: "Title of RequestItem" },
```
jkoenig134, line 174:
```suggestion
                        { "@type": "AuthenticationRequestItem", mustBeAccepted: false, title: "aTitle" },
```
same everywhere else please.

britsta: I've just looked at the file again, I think @Milena-Czierlinski wrote "Title of RequestItem" on purpose, because "Title of Attribute" is also used in the file. In your opinion, should I still change everything to "aTitle" or should I leave it differentiable because, for example, an `attribute` within a RequestItem can have a `title`?

jkoenig134 (approved):

packages/runtime/test/modules/DeciderModule.test.ts
jkoenig134: it's ok. but I think in that case aRequestItemTitle would be closer to the other test code.

IF you want to refactor it, please do it on another PR (to main), if not, I approved.

jkoenig134 (approved):

PR #447, opened 2025-03-05
Pagination of Messages

Magnus-Kuhn:
# Readiness checklist

- [ ] I added/updated tests.
- [ ] I labeled the PR.
- [ ] I self-reviewed the PR.


jkoenig134:
@Magnus-Kuhn why is this a breaking change?

Magnus-Kuhn:
The total number of messages is returned as well in the use case as `{messages: MessageDTO[], messageCount: number}`.

jkoenig134:
> The total number of messages is returned as well in the use case as `{messages: MessageDTO[], messageCount: number}`.

IMO no real reason. We can keep the connector api the same and we can update the App accordingly.

PR #449, opened 2025-03-06
Allow to specify RepositoryAttribute duplicate in AcceptReadAttributeRequestItemParametersWithNewAttribute

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description
If a ReadAttributeRequestItem is answered with a new Attribute that is a duplicate of an existing RepositoryAttribute, no error should be thrown anymore. Instead, the newly created own shared IdentityAttribute should map to the already existing RepositoryAttribute and no new RepositoryAttribute is created in addition.


packages/consumption/src/modules/requests/itemProcessors/readAttribute/ReadAttributeRequestItemProcessor.ts
```diff
@@ -379,12 +366,76 @@ export class ReadAttributeRequestItemProcessor extends GenericRequestItemProcess
...
+            );
+            const succeededOwnSharedIdentityAttribute = ownSharedIdentityAttributesAfterSuccession.successor;
+
+            return AttributeSuccessionAcceptResponseItem.from({
                 result: ResponseItemResult.Accepted,
-                attributeId: sharedLocalAttribute.id,
-                attribute: sharedLocalAttribute.content
+                successorId: succeededOwnSharedIdentityAttribute.id,
+                successorContent: succeededOwnSharedIdentityAttribute.content,
+                predecessorId: latestSharedVersion.id
```
Milena-Czierlinski, line 438: This part appears also in the CreateAttributeRequestItemProcessor and soon also in the ProposeAttributeRequestItemProcessor. The only difference it that here a ReadAttributeAcceptResponseItem is returned in line 406, which is a Create- or ProposeAttributeAcceptResponseItem in the other processors, respectively. @jkoenig134 In your opinion should I make an effort to extract this logic to a separate file such that it can be reused by all processors? Unfortunately, this isn't as trivial as I initially thought and so I considered only extracting the lower part with the AttributeSuccession- and AttributeAlreadySharedAcceptResponseItem. But then I wasn't sure if this is actually better than having the complete logic in each processor.

jkoenig134 (approved):
great!

PR #456, opened 2025-03-12
Add TransferFileOwnershipRequestItem

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


packages/content/src/requests/items/transferFileOwnership/TransferFileOwnershipRequestItem.ts
```diff
@@ -0,0 +1,30 @@
+import { serialize, type, validate } from "@js-soft/ts-serval";
+import { FileReference, IFileReference } from "@nmshd/transport";
```
jkoenig134, line 2: content may **not** import from transport, you have to move reference and filereference to core-types. Please also move the according tests.

packages/runtime/src/dataViews/content/ResponseItemDVOs.ts
```diff
@@ -81,6 +82,13 @@ export interface RegisterAttributeListenerAcceptResponseItemDVO extends AcceptRe
     listener: LocalAttributeListenerDVO;
 }
 
+export interface TransferFileOwnershipAcceptResponseItemDVO extends AcceptResponseItemDVO {
+    type: "TransferFileOwnershipAcceptResponseItemDVO";
+    repositoryAttribute?: LocalAttributeDVO;
```
jkoenig134, line 87: why is this optional?

jkoenig134: nvm, I think I got it. If I am the recipient I don't have the repo attribute, right?

Milena-Czierlinski: Recipient of the ResponseItem, sender of the Request, yes.

packages/consumption/test/core/TestUtil.ts
```diff
@@ -462,15 +463,16 @@ export class TestUtil {
         });
     }
 
-    public static async uploadFile(from: AccountController, fileContent: CoreBuffer): Promise<File> {
+    public static async uploadFile(from: AccountController, parameters?: { fileContent?: CoreBuffer; expiredAt?: CoreDate; tags?: string[] }): Promise<File> {
         const params: ISendFileParameters = {
-            buffer: fileContent,
+            buffer: parameters?.fileContent ?? CoreBuffer.from(await fs.promises.readFile(`${__dirname}/../__assets__/test.txt`)),
```
jkoenig134, line 468:
```suggestion
            buffer: parameters?.fileContent ?? CoreBuffer.from("test"),
```
don't forget to remove that test.txt file.

packages/consumption/test/core/TestUtil.ts
```diff
@@ -462,15 +463,16 @@ export class TestUtil {
         });
     }
 
-    public static async uploadFile(from: AccountController, fileContent: CoreBuffer): Promise<File> {
+    public static async uploadFile(from: AccountController, parameters?: { fileContent?: CoreBuffer; expiredAt?: CoreDate; tags?: string[] }): Promise<File> {
```
jkoenig134, line 466:
```suggestion
    public static async uploadFile(from: AccountController, parameters?: { fileContent?: CoreBuffer; expiresAt?: CoreDate; tags?: string[] }): Promise<File> {
```

packages/consumption/test/core/TestUtil.ts
```diff
@@ -462,15 +463,16 @@ export class TestUtil {
...
         const params: ISendFileParameters = {
-            buffer: fileContent,
+            buffer: parameters?.fileContent ?? CoreBuffer.from(await fs.promises.readFile(`${__dirname}/../__assets__/test.txt`)),
             title: "aFileName",
-            description: "Dies ist eine Beschreibung",
-            filename: "Test.bin",
+            description: "This is a valid file description",
+            filename: "test.txt",
```
jkoenig134, line 471: if you refactor these properties please refactor them properly (`aDescription`, `aFilename`)

jkoenig134: I think this will bring up be merge conficts as @britsta has changed this on v7..

jkoenig134 (approved):

packages/runtime/src/modules/decide/RequestConfig.ts
```diff
@@ -126,6 +126,10 @@ export interface ShareAttributeRequestItemConfig extends RequestItemConfig {
     "content.item.attribute.value.description"?: string | string[];
 }
 
+export interface TransferFileOwnershipRequestItemConfig extends RequestItemConfig {
+    "content.item.@type": "TransferFileOwnershipRequestItem";
```
britsta, line 130: Is `fileReference` not listed here because it wouldn't make much sense or are there other reasons? Maybe someone wants to accept all ownership transfers of files whose `fileReference` they already know because they have already been shared with them without ownership transfer before.

packages/runtime/src/modules/decide/RequestConfig.ts
```diff
@@ -126,6 +126,10 @@ export interface ShareAttributeRequestItemConfig extends RequestItemConfig {
     "content.item.attribute.value.description"?: string | string[];
 }
 
+export interface TransferFileOwnershipRequestItemConfig extends RequestItemConfig {
```
britsta, line 129: Please add a test for the `TransferFileOwnershipRequestItemConfig` to the `DeciderModule.test.ts`.

packages/runtime/src/modules/decide/RequestConfig.ts
Milena-Czierlinski: We currently don't provide the option to specify anything with a fix id, it is the same for Attributes, as it doesn't make a lot of sense. For the App the default configuration will probably be that any TransferFileOwnershipRequestItem will be accepted automatically.

jkoenig134: Yeah, specifying anything here would mean that only one file at all can be accepted automatically.

For the future it could make sense to have a list of trusted peers to auto accept, but I think that would count for every request item.

packages/runtime/src/modules/decide/RequestConfig.ts
jkoenig134: I'd be totally ok with bringing this test in another PR, so that we can iterate with that feature in App / Connector.

packages/runtime/src/modules/decide/RequestConfig.ts
britsta: It is okay for me if we don't add the property here. I just want to make sure it was intentionally. 💪

> For the future it could make sense to have a list of trusted peers to auto accept, but I think that would count for every request item.

I think this can already be partially achieved with the `peer` property of the GeneralRequestConfig.

PR #463, opened 2025-03-19
Add caching of tags

sebbi08:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [ ] I self-reviewed the PR.


Milena-Czierlinski (dismissed):
@sebbi08 Walked me through his changes and I am fine with them. @jkoenig134 Do you want to add your review, too?

jkoenig134:
I do, and I have some comments already.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -47,6 +47,8 @@ import { IdentityAttributeQueryTranslator, RelationshipAttributeQueryTranslator,
 export class AttributesController extends ConsumptionBaseController {
     private attributes: SynchronizedCollection;
     private attributeTagClient: TagClient;
+    private cachedAttributeTagCollection?: AttributeTagCollection;
+    private cachedAttributeTagCollectionTimestamp?: CoreDate;
```
jkoenig134, line 51: that collection is lost when restarting the app, this is not good, b/c it looses the offline functionality.

Also the cache should be not lazily loaded, instead prepared for usage.

packages/transport/src/modules/tags/backbone/TagClient.ts
```diff
@@ -1,9 +1,26 @@
...
+        }
+        const result = await this.get<BackboneGetTagCollection>("/api/v1/Tags", undefined, {
+            headers,
+            validateStatus: (status) => status === 200 || status === 304
+        });
+        if (result.responseStatus === 304) {
+            return undefined;
+        }
+        this.tagCollectionETag = result.etag;
+        return result;
```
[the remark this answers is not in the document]
jkoenig134: github marked this as outdated, but it isn't.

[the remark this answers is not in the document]
sebbi08: Yea, but than we would need to return the etag with the `BackboneGetTagCollection` or it needs to e part of that type. Also, I think we should leaf logic around the HTTP request in the clients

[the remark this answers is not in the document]
jkoenig134: Yeah, but that's application logic and not client logic. A client must be stateless.

packages/consumption/src/consumption/ConsumptionController.ts
```diff
@@ -64,6 +65,11 @@ export class ConsumptionController {
         return this._drafts;
     }
 
+    private _tags: TagsController;
+    public get tags(): TagsController {
+        return this._tags;
+    }
```
jkoenig134, line 71: we had the decision to NOT have tags in an own controller b/c it would have to be named AttributeTags which is ugly. Please move that logic back to AttributesController.

sebbi08: Yea but with the decision that files can have tags we will also need those methods for files also the AttributeControler is already very bloated.

jkoenig134: > files can have tags

as soon as this is not tackled by the backbone, I don't think this has to be considered when implementing this feature.

> the AttributeControler is already very bloated

moving stuff to a completely new module is not the only solution, so no argument

packages/consumption/test/modules/attributes/AttributeTagCollection.test.ts
```diff
@@ -69,7 +69,7 @@ describe("AttributeTagCollection", function () {
     });
 
     test("should receive the legal tags from the Backbone", async function () {
-        when(mockedClient.getTagCollection()).thenResolve(ClientResult.ok(mockTags));
+        when(mockedClient.getTagCollection(anything())).thenResolve(ClientResult.ok(mockTags));
```
Milena-Czierlinski, line 72: What is the reason for this change? Same for the other test.

jkoenig134: the getTagCollection method was changed

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1376,7 +1394,73 @@ export class AttributesController extends ConsumptionBaseController {
...
+            const attributeTagCollection = AttributeTagCollection.from(backboneTagCollection.value);
+            await this.setTagCollection(attributeTagCollection);
+
+            resolve(attributeTagCollection);
+        });
+
+        return await this.readTagCollectionPromise.then((value) => {
+            this.readTagCollectionPromise = undefined;
+            return value;
+        });
```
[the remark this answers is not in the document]
sebbi08: Do you mean that I should create a separate function instead of the new Promise part ?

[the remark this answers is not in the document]
jkoenig134: also, you are missing try..catch..finally to clean up the promise properly.

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1375,8 +1393,75 @@ export class AttributesController extends ConsumptionBaseController {
         });
     }
 
+    public async _getAttributeTagCollection(): Promise<AttributeTagCollection> {
```
jkoenig134, line 1396: private

jkoenig134, line 1396: ... and below the caller

jkoenig134 (approved):

PR #466, opened 2025-03-21
Synchronizable jsons are missing ids

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [ ] I labeled the PR.
- [ ] I self-reviewed the PR.

# Description

the property id is defined in CoreSynchronizable and ICoreSynchronizable, but were missed in the jsons.


Milena-Czierlinski (approved):
Have you checked that these are the only places where id is missing or do we need to investigate other jsons, too?

jkoenig134:
> Have you checked that these are the only places where id is missing or do we need to investigate other jsons, too?

I did.

PR #469, opened 2025-03-26
Allow a display name to be specified optionally, but not required

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

As the module names are just nice to have and written in the same way (almost always) I thought why not make it easier for integrators to register new modules, by having the possibility to NOT have to name them, instead the module decides for its own name.

### DIFF
--- a/packages/app-runtime/src/AppRuntime.ts
+++ b/packages/app-runtime/src/AppRuntime.ts
@@ -256,7 +256,7 @@ export class AppRuntime extends Runtime<AppConfig> {
         const moduleConstructor = AppRuntime.moduleRegistry[moduleConfiguration.location];
         if (!moduleConstructor) {
             const error = new Error(
-                `Module '${moduleConfiguration.location}' could not be loaded, because it was not registered. Please register all modules before running init. Available modules: ${Object.keys(
+                `Module at location '${moduleConfiguration.location}' could not be loaded, because it was not registered. Please register all modules before running init. Available modules: ${Object.keys(
                     AppRuntime.moduleRegistry
                 ).join(", ")}`
             );


packages/app-runtime/src/AppRuntime.ts
```diff
@@ -256,7 +256,9 @@ export class AppRuntime extends Runtime<AppConfig> {
         const moduleConstructor = AppRuntime.moduleRegistry[moduleConfiguration.location];
         if (!moduleConstructor) {
             const error = new Error(
-                `Module '${this.getModuleName(moduleConfiguration)}' could not be loaded, because it was not registered. Please register all modules before running init.`
+                `Module '${moduleConfiguration.location}' could not be loaded, because it was not registered. Please register all modules before running init. Available modules: ${Object.keys(
```
tnotheis, line 259: Should probably use the (display) name.

packages/runtime/src/Runtime.ts
```diff
@@ -325,12 +325,12 @@ export abstract class Runtime<TConfig extends RuntimeConfig = RuntimeConfig> {
             const moduleConfiguration = this.runtimeConfig.modules[moduleName];
 
             if (!moduleConfiguration.enabled) {
-                this.logger.debug(`Skip loading module '${this.getModuleName(moduleConfiguration)}' because it is not enabled.`);
+                this.logger.debug(`Skip loading module at location '${moduleConfiguration.location}' because it is not enabled.`);
```
tnotheis, line 328: Should probably use the (display) name.

tnotheis (changes requested):

packages/app-runtime/src/AppRuntime.ts
jkoenig134: nope, we don't have the display name at this point. This is the only ugly thing with my approach ;)

tnotheis (approved):

PR #475, opened 2025-04-04
Add FormFieldRequestItem

britsta:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description
It should be possible to create forms for requesting unstructured data that is not to be saved as Attributes. The elements of the form should be part of a Request. There are multiple kinds of form elements, which can all be represented by one dedicated RequestItem: The so-called FormFieldRequestItem can be used to request data of a specified value type (String, Integer, Double, Boolean and Date) or to prompt the recipient to select options, which are displayed using a different UI (radio button groups, dropdown menus and checklists) depending on the number of options and the specified selection type (`allowMultipleSelection` flag). Also, it is possible to request ratings that can be displayed as stars in the App's UI by using the FormFieldRequestItem.


packages/consumption/src/modules/requests/itemProcessors/freeValueForm/FreeValueFormRequestItemProcessor.ts
```diff
@@ -0,0 +1,44 @@
...
+                )
+            );
+        }
+
+        return ValidationResult.success();
+    }
+
+    private static canBeConvertedToValidNumber(value: string): boolean {
+        const valueIsEmptyString = value.trim() === "";
+        return !valueIsEmptyString && !isNaN(Number(value));
```
Magnus-Kuhn, line 29: `isFinite()` would be technically better since it also bans infinity

packages/consumption/src/modules/requests/itemProcessors/freeValueForm/FreeValueFormRequestItemProcessor.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    private static canBeConvertedToValidDate(value: string): boolean {
+        const date = new Date(value);
+        return !isNaN(date.getTime());
```
Magnus-Kuhn, line 34: the 31st of February passes this validation

packages/consumption/src/modules/requests/itemProcessors/selectionForm/SelectionFormRequestItemProcessor.ts
```diff
@@ -0,0 +1,60 @@
...
+                )
+            );
+        }
+
+        const uniqueOptions = new Set(parsedParams.options);
+        if (uniqueOptions.size !== parsedParams.options.length) {
+            return ValidationResult.error(ConsumptionCoreErrors.requests.invalidAcceptParameters("The options specified for accepting a SelectionFormRequestItem must be unique."));
+        }
+
+        for (const option of parsedParams.options) {
```
Magnus-Kuhn, line 42: I would consider returning all wrong options in one go

packages/consumption/src/modules/requests/itemProcessors/freeValueForm/FreeValueFormRequestItemProcessor.ts
Magnus-Kuhn: I guess we have to specify accepted formats simply in order to avoid ambiguity between e. g. DD/MM and MM/DD

jkoenig134: ISO should be the only accepted format. I already discussed this with @britsta, we will use CoreDate to validate this.

packages/consumption/src/modules/requests/itemProcessors/formField/AcceptFormFieldRequestItemParameters.ts
```diff
@@ -0,0 +1,50 @@
+import { Serializable, serialize, type, validate, ValidationError } from "@js-soft/ts-serval";
+import { nameof } from "ts-simple-nameof";
+import { ConsumptionError } from "../../../../consumption/ConsumptionError";
+import { AcceptRequestItemParametersJSON } from "../../incoming/decide/AcceptRequestItemParameters";
+
+export interface AcceptFormFieldRequestItemParametersJSON extends AcceptRequestItemParametersJSON {
+    freeValue?: string;
+    options?: string[];
```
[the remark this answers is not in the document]
britsta: Yeah, definitely worth considering, although I might adapt the property name.

If I'll extend the `freeValue` type from `string` to `string | number | boolean`, you'd also be in favor of combining the properties to `response: string | number | boolean | string[];`, right?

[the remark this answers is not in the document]
jkoenig134: yep

packages/consumption/src/modules/requests/itemProcessors/formField/FormFieldRequestItemProcessor.ts
```diff
@@ -0,0 +1,146 @@
...
+            if (uniqueOptions.size !== parsedParams.response.length) {
+                return ValidationResult.error(ConsumptionCoreErrors.requests.invalidAcceptParameters("The options specified for accepting a selection form field must be unique."));
+            }
+
+            const unknownOptions: string[] = [];
+            for (const option of parsedParams.response) {
+                if (!requestItem.settings.options.includes(option)) {
+                    unknownOptions.push(option);
+                }
+            }
```
Magnus-Kuhn, line 110:
```suggestion
            const unknownOptions = _.difference(parsedParams.response, requestItem.settings.options);
```

packages/content/test/requests/items/FormFieldAcceptResponseItem.test.ts
```diff
@@ -0,0 +1,41 @@
+import { FormFieldAcceptResponseItem, ResponseItemResult } from "../../../src";
+
+describe("creation of FormFieldAcceptResponseItem", () => {
```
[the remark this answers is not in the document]
britsta: We had a tricky problem during implementation. These were the tests @jkoenig134 and I added to get to the bottom of them. :D @Magnus-Kuhn, would you like to have more tests in this file or do you think that this test file with the minimal examples is no longer necessary?

packages/runtime/test/dataViews/requestItems/FormFieldRequestItemDVO.test.ts
```diff
@@ -0,0 +1,410 @@
...
+let responseItems: DecideRequestItemParametersJSON[];
+
+afterAll(() => serviceProvider.stop());
+
+beforeEach(function () {
+    eventBus1.reset();
+    eventBus2.reset();
+});
+
+describe("FormFieldRequestItemDVO with StringFormFieldSettings", () => {
```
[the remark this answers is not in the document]
britsta: The selection form field is very different from the other form fields. For the other form fields, I have chosen the string form field as a representative for testing.

packages/consumption/src/modules/requests/itemProcessors/formField/FormFieldRequestItemProcessor.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    public override canAccept(requestItem: FormFieldRequestItem, params: AcceptFormFieldRequestItemParametersJSON): ValidationResult {
+        const parsedParams = AcceptFormFieldRequestItemParameters.from(params);
+
+        if (!(requestItem.settings instanceof SelectionFormFieldSettings)) {
+            if (Array.isArray(parsedParams.response)) {
+                return ValidationResult.error(ConsumptionCoreErrors.requests.invalidAcceptParameters("Only a selection form field can be accepted with an array."));
+            }
+
+            if (
```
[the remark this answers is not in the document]
britsta: Technically speaking, you're right. However, it may be more efficient to only perform these checks if the form field is not a selection form field. In addition, the current structure emphasizes the different nature of the selection form field from the other form fields.

packages/content/test/requests/items/FormFieldAcceptResponseItem.test.ts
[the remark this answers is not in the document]
britsta: Yeah, it would also be fine for me to remove this test file if @jkoenig134 doesn't mind.

packages/consumption/src/modules/requests/itemProcessors/formField/FormFieldRequestItemProcessor.ts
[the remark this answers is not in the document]
Magnus-Kuhn: I don't expect this to be a significant factor for performance.
And my scenario is that I'm in the code, wonder why there is this amount of indentation and then see that this is for a condition that can't be fulfilled anyway :D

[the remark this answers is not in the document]
britsta: All right, I'll refactor it, then it's less nested :D

packages/consumption/test/modules/requests/itemProcessors/formField/FormFieldRequestItemProcessor.test.ts
```diff
@@ -0,0 +1,1024 @@
...
+    const aMin = 1;
+    const aMax = 10;
+    const aUnit = "aUnit";
+
+    const aString = "aString";
+    const anInteger = 123456789;
+    const aDouble = 123456789.123456789;
+    const aBoolean = true;
+    const aDate = "2000-01-01T00:00:00.000Z";
+    const aRating = 5;
```
jkoenig134, line 47: I'm not a big fan of all these const values. Why do you use them? For me it requires a massive amount of jumping back and forth to understand what the test does.

packages/consumption/test/modules/requests/itemProcessors/formField/FormFieldRequestItemProcessor.test.ts
```diff
@@ -0,0 +1,1024 @@
...
+                    accept: true,
+                    response: anInteger
+                };
+
+                const result = processor.canAccept(requestItem, acceptParams);
+
+                expect(result).successfulValidationResult();
+            });
+
+            test("returns an error when it is tried to accept an integer form field with no free integer", function () {
```
jkoenig134, line 409: > with no free integer

please just write "with a double"

In general I'm not a fan of the wording `free x` can we get rid of that?

packages/content/src/attributes/ThirdPartyRelationshipAttributeQuery.ts
```diff
@@ -4,7 +4,7 @@ import { AbstractAttributeQuery, AbstractAttributeQueryJSON, IAbstractAttributeQ
 export interface ThirdPartyRelationshipAttributeQueryJSON extends AbstractAttributeQueryJSON {
     "@type": "ThirdPartyRelationshipAttributeQuery";
     key: string;
-    owner: ThirdPartyRelationshipAttributeQueryOwner | `${ThirdPartyRelationshipAttributeQueryOwner}`;
+    owner: `${ThirdPartyRelationshipAttributeQueryOwner}`;
```
jkoenig134, line 7: this should not be part of this PR.

jkoenig134: but as it is here .. leave it.

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
```diff
@@ -0,0 +1,29 @@
+import { serialize, type, validate } from "@js-soft/ts-serval";
+import { FormFieldSettings, FormFieldSettingsJSON, IFormFieldSettings } from "./FormFieldSettings";
+
+export const minRating = 1;
```
jkoenig134, line 4: This is a very bad naming for a field that is exported from the whole library.

Imagine you install this lib from npmjs and find `minRating` exported without any further explanation...

Please make this a static field (or maybe better would be a getter) of RatingFormFieldSettings OR (and I would be a fan of this approach) just remove it. It's well known that 0 is not a valid rating and we can just assume the scale.

packages/content/src/requests/items/formField/settings/StringFormFieldSettings.ts
```diff
@@ -0,0 +1,38 @@
+import { serialize, type, validate } from "@js-soft/ts-serval";
+import { FormFieldSettings, FormFieldSettingsJSON, IFormFieldSettings } from "./FormFieldSettings";
+
+export interface StringFormFieldSettingsJSON extends FormFieldSettingsJSON {
+    "@type": "StringFormFieldSettings";
+    allowNewLines?: true;
```
jkoenig134, line 6: actually newline is one word. So this should be `allowNewlines`. Same everywhere else.

packages/content/test/requests/items/FormFieldAcceptResponseItem.test.ts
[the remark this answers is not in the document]
jkoenig134: IMO it's not bad to at least check the success cases, so I see no reason deleting it.

packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -900,6 +927,16 @@ export class DataViewExpander {
                         name: name
                     } as FreeTextAcceptResponseItemDVO;
 
+                case "FormFieldAcceptResponseItem":
+                    const freeValueFormResponseItem = responseItem as FormFieldAcceptResponseItemJSON;
```
jkoenig134, line 931: freeValue?

packages/runtime/test/dataViews/requestItems/FormFieldRequestItemDVO.test.ts
```diff
@@ -0,0 +1,410 @@
...
+    const aMin = 1;
+    const aMax = 100;
```
jkoenig134, line 51: same as in the other test file. If you want to use these constants please move them to every test to avoid jumping back and forth.

jkoenig134 (changes requested):

packages/consumption/test/modules/requests/itemProcessors/formField/FormFieldRequestItemProcessor.test.ts
britsta: I use variables to have descriptive names for values. I only declare them once at the beginning to avoid redundancy. However, I have no problem refactoring it so that the variable declarations are in the tests themselves.

jkoenig134: I don't think `aMin` and `aMax` are very descriptive. Also having those variables will lead to those variables being used in the expect which is an antipattern.

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
```diff
@@ -0,0 +1,33 @@
...
+@type("RatingFormFieldSettings")
+export class RatingFormFieldSettings extends FormFieldSettings implements IRatingFormFieldSettings {
+    @serialize()
+    @validate({ min: 5, max: 10 })
+    public maxRating: MaxRating;
+
+    private readonly minRating = 1;
+    public getMinRating(): number {
+        return this.minRating;
+    }
```
jkoenig134, line 24: As in my comment this has to be static and also a getter, not a function named get... :)

packages/content/src/requests/items/formField/settings/SelectionFormFieldSettings.ts
```diff
@@ -0,0 +1,32 @@
+import { serialize, type, validate } from "@js-soft/ts-serval";
+import { FormFieldSettings, FormFieldSettingsJSON, IFormFieldSettings } from "./FormFieldSettings";
+
+export interface SelectionFormFieldSettingsJSON extends FormFieldSettingsJSON {
+    "@type": "SelectionFormFieldSettings";
+    options: string[];
+    allowMultipleSelection?: true;
```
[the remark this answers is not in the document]
britsta: I am referring to the term "Mehrfachauswahl" / "multiple selection".

packages/consumption/src/modules/requests/itemProcessors/formField/FormFieldRequestItemProcessor.ts
```diff
@@ -0,0 +1,138 @@
...
+        if (
+            (requestItem.settings instanceof StringFormFieldSettings && typeof parsedParams.response !== "string") ||
+            (requestItem.settings instanceof IntegerFormFieldSettings && !Number.isInteger(parsedParams.response)) ||
+            (requestItem.settings instanceof DoubleFormFieldSettings && typeof parsedParams.response !== "number") ||
+            (requestItem.settings instanceof BooleanFormFieldSettings && typeof parsedParams.response !== "boolean") ||
+            (requestItem.settings instanceof DateFormFieldSettings && !FormFieldRequestItemProcessor.isValidDate(parsedParams.response)) ||
+            (requestItem.settings instanceof RatingFormFieldSettings &&
+                !FormFieldRequestItemProcessor.isValidRating(parsedParams.response, RatingFormFieldSettings.minRating, requestItem.settings.maxRating))
+        ) {
+            return ValidationResult.error(ConsumptionCoreErrors.requests.invalidAcceptParameters(`The response provided cannot be used to accept the form field.`));
```
Milena-Czierlinski, line 66:
```suggestion
            return ValidationResult.error(ConsumptionCoreErrors.requests.invalidAcceptParameters("The response provided cannot be used to accept the form field."));
```

packages/consumption/src/modules/requests/itemProcessors/formField/FormFieldRequestItemProcessor.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    public override canAccept(requestItem: FormFieldRequestItem, params: AcceptFormFieldRequestItemParametersJSON): ValidationResult {
+        const parsedParams = AcceptFormFieldRequestItemParameters.from(params);
+
+        if (!(requestItem.settings instanceof SelectionFormFieldSettings)) {
+            if (Array.isArray(parsedParams.response)) {
+                return ValidationResult.error(ConsumptionCoreErrors.requests.invalidAcceptParameters("Only a selection form field can be accepted with an array."));
+            }
+        }
```
Milena-Czierlinski, line 55:
```suggestion
       if (Array.isArray(parsedParams.response) && !(requestItem.settings instanceof SelectionFormFieldSettings)) {
            return ValidationResult.error(ConsumptionCoreErrors.requests.invalidAcceptParameters("Only a selection form field can be accepted with an array."));
        }
```

packages/consumption/src/modules/requests/itemProcessors/formField/FormFieldRequestItemProcessor.ts
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            if (parsedParams.response.length === 0) {
+                return ValidationResult.error(ConsumptionCoreErrors.requests.invalidAcceptParameters("At least one option must be specified to accept a selection form field."));
+            }
+
+            const uniqueOptions = new Set(parsedParams.response);
+            if (uniqueOptions.size !== parsedParams.response.length) {
+                return ValidationResult.error(ConsumptionCoreErrors.requests.invalidAcceptParameters("The options specified for accepting a selection form field must be unique."));
+            }
```
Milena-Czierlinski, line 101: @britsta Can you please explain to me what this does?

Milena-Czierlinski: > Selection form fields should usually provide distinct options for selection

I am struggling with imaging how the response for a selection form field looks like and what it means for it to be unique or not.

britsta: I have just edited my previous comment, does that answer your question?

Otherwise, perhaps an example: If I have created a selection form field that provides the options "A" and "B" and an integrator responds to it, then they should only be able to respond with ["A"], ["B"] or ["A", "B"], but not with ["A", "A"] or ["B", "B", "B", "B", "B", "B"], for example. App users would not be able to reply with duplicates because of the UI.

Milena-Czierlinski: Yes this helps, thank you! Another question for my understanding: Does the `options` array contain the strings that are going to be displayed by the UI? I.e. might the strings be complete sentences?

jkoenig134: But do we actually need that?

Milena-Czierlinski: > It's a very simple (and imo readable) uniqueness check :)

That part I had no problem understanding. ;)

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
```diff
@@ -0,0 +1,33 @@
...
+@type("RatingFormFieldSettings")
+export class RatingFormFieldSettings extends FormFieldSettings implements IRatingFormFieldSettings {
+    @serialize()
+    @validate({ min: 5, max: 10 })
+    public maxRating: MaxRating;
+
+    private static readonly MIN_RATING = 1;
+    public static get minRating(): number {
+        return this.MIN_RATING;
+    }
```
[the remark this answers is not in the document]
britsta: The value of the minimum rating is required in the validation in the FormFieldRequestItemProcessor for valid rating. Since magic numbers are not a good coding practice, I would like to declare it in some way.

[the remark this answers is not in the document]
jkoenig134: What's a magic number?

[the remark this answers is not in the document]
britsta: As I don't want to parrot Wiki: [Magic Numbers](https://en.wikipedia.org/wiki/Magic_number_(programming))

packages/consumption/src/modules/requests/itemProcessors/formField/FormFieldRequestItemProcessor.ts
britsta: @Milena-Czierlinski In your "option" hehe :P

I also think we don't need this.

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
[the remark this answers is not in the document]
jkoenig134: > As I don't want to parrot Wiki: [Magic Numbers](https://en.wikipedia.org/wiki/Magic_number_(programming))

In that case - why not just define it where you use it?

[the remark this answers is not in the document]
jkoenig134: > I can well imagine that one might think that 0 would also be a valid minimum ranking. Personally, I wouldn't have been sure whether it should have been 0 or 1. By naming the constant, it is clear to everyone at all times where this number comes from.

0 ist not possible to press in a UI.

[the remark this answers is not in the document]
britsta: Because the RatingFormFieldSettings are the place where it is determined how we (the Runtime) think ratings. It is quite possible that we will have to use the minimum rating again elsewhere in the Runtime. It is then easier to use the number from the RatingFormFieldSettings instead of declaring it anew each time (this is also more error-prone, because otherwise someone could suddenly come up with a new minimum rating).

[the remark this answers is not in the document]
jkoenig134: Ok, then please do the changes I suggested then I'm fine. I still don't like this, but it's a detail not worth to spend more time on.

packages/consumption/src/modules/requests/itemProcessors/formField/FormFieldRequestItemProcessor.ts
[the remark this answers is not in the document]
britsta: Good idea, I'll try this approach.

packages/content/src/requests/items/formField/settings/BooleanFormFieldSettings.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public canAccept(response: string | number | boolean | string[]): string | undefined {
+        if (Array.isArray(response)) {
+            return "Only a selection form field can be accepted with an array.";
+        }
+
+        if (typeof response !== "boolean") {
+            return "The response provided cannot be used to accept the form field.";
+        }
```
jkoenig134, line 22: now as we have brought the validations inside the Settings we can create way better error messages:

```suggestion
        if (typeof response !== "boolean") {
            return "Only a boolean can be used to respond to this item.";
        }
```

same for all other places, the array thing is only needed b/c it all in place before.

packages/content/src/requests/items/formField/settings/DateFormFieldSettings.ts
```diff
@@ -0,0 +1,40 @@
...
+        if (!DateFormFieldSettings.isValidDate(response)) {
+            return "The response provided cannot be used to accept the form field.";
+        }
+
+        return;
+    }
+
+    private static isValidDate(value: any): boolean {
+        return typeof value === "string" && CoreDate.from(value).dateTime.isValid;
+    }
```
jkoenig134, line 31: not static please. And IMO this can also be inline now.

packages/content/src/requests/items/formField/settings/DateFormFieldSettings.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    public canAccept(response: string | number | boolean | string[]): string | undefined {
+        if (Array.isArray(response)) {
+            return "Only a selection form field can be accepted with an array.";
+        }
+
+        if (!DateFormFieldSettings.isValidDate(response)) {
+            return "The response provided cannot be used to accept the form field.";
+        }
```
jkoenig134, line 24:
```suggestion
        const isValidDate = typeof value === "string" && CoreDate.from(value).dateTime.isValid
        if (!isValidDate) {
            return "The response provided cannot be used to accept the form field.";
        }
```

packages/content/src/requests/items/formField/settings/DateFormFieldSettings.ts
jkoenig134: see above

packages/content/src/requests/items/formField/settings/DateFormFieldSettings.ts
jkoenig134: and maybe we need a better error message explaining what WOULD be valid here instead of just saying it is not.

packages/content/src/requests/items/formField/settings/DoubleFormFieldSettings.ts
```diff
@@ -0,0 +1,66 @@
...
+        if (Array.isArray(response)) {
+            return "Only a selection form field can be accepted with an array.";
+        }
+
+        if (typeof response !== "number") {
+            return "The response provided cannot be used to accept the form field.";
+        }
+
+        if (this.max && response > this.max) {
+            return "The response cannot be greater than the max.";
```
jkoenig134, line 49: please show the max value in message

packages/content/src/requests/items/formField/settings/DoubleFormFieldSettings.ts
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        if (this.min && response < this.min) {
+            return "The response cannot be smaller than the min.";
```
jkoenig134, line 53: please show the min value in the message

packages/content/src/requests/items/formField/settings/IntegerFormFieldSettings.ts
```diff
@@ -0,0 +1,66 @@
...
+            return "The response provided cannot be used to accept the form field.";
+        }
+
+        if (this.max && response > this.max) {
+            return "The response cannot be greater than the max.";
+        }
+
+        if (this.min && response < this.min) {
+            return "The response cannot be smaller than the min.";
+        }
```
jkoenig134, line 54: see double

packages/content/src/requests/items/formField/settings/IntegerFormFieldSettings.ts
```diff
@@ -0,0 +1,66 @@
...
+    @validate({ nullable: true })
+    public unit?: string;
+
+    @serialize()
+    @validate({ nullable: true })
+    public min?: number;
+
+    @serialize()
+    @validate({ nullable: true })
+    public max?: number;
```
jkoenig134, line 29: we need a validation that min and max are integers and not doubles.

jkoenig134: You can use serval postFrom

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
```diff
@@ -0,0 +1,52 @@
...
+
+@type("RatingFormFieldSettings")
+export class RatingFormFieldSettings extends FormFieldSettings implements IRatingFormFieldSettings {
+    @serialize()
+    @validate({ min: 5, max: 10 })
+    public maxRating: MaxRating;
+
+    public static get minRating(): number {
+        return 1;
+    }
```
jkoenig134, line 23: static on top please, and this can be private as long a noone else uses this.

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export interface IRatingFormFieldSettings extends IFormFieldSettings {
+    maxRating: MaxRating;
+}
+
+@type("RatingFormFieldSettings")
+export class RatingFormFieldSettings extends FormFieldSettings implements IRatingFormFieldSettings {
+    @serialize()
+    @validate({ min: 5, max: 10 })
+    public maxRating: MaxRating;
```
jkoenig134, line 19:
```suggestion
    public maxRating: number;
```

I think everything else will break serval

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
```diff
@@ -0,0 +1,52 @@
...
+        if (!RatingFormFieldSettings.isValidRating(response, RatingFormFieldSettings.minRating, this.maxRating)) {
+            return "The response provided cannot be used to accept the form field.";
+        }
+
+        return;
+    }
+
+    private static isValidRating(value: any, minRating: number, maxRating: number): boolean {
+        return Number.isInteger(value) && value >= minRating && value <= maxRating;
+    }
```
jkoenig134, line 43: inline please (with a good named local variable)

packages/content/src/requests/items/formField/settings/SelectionFormFieldSettings.ts
```diff
@@ -0,0 +1,72 @@
...
+        if (uniqueOptions.size !== this.options.length) {
+            return "A selection form field must provide unique options.";
+        }
+
+        return;
+    }
+
+    public canAccept(response: string | number | boolean | string[]): string | undefined {
+        if (!Array.isArray(response)) {
+            return "A selection form field must be accepted with an array.";
```
jkoenig134, line 41: nope? If allowMultipleSelection is false it should be a string.

jkoenig134: should be checked first.

packages/content/src/requests/items/formField/settings/StringFormFieldSettings.ts
```diff
@@ -0,0 +1,66 @@
...
+            return "The response provided cannot be used to accept the form field.";
+        }
+
+        if (this.max && response.length > this.max) {
+            return "The length of the response cannot be greater than the max.";
+        }
+
+        if (this.min && response.length < this.min) {
+            return "The length of the response cannot be smaller than the min.";
+        }
```
jkoenig134, line 54: include min max in error pls

jkoenig134 (changes requested):

packages/content/src/requests/items/formField/settings/SelectionFormFieldSettings.ts
britsta: Even if only one option is selected, we expect an array, as also becomes clear in the tests: An array that contains exactly one string, e.g. `["optionA"]`.

jkoenig134: I know that we currently accept an array with one element when allowMultipleSelection is false, but we shouldn't.

IMO its way cleaner to get one string instead of an array with one string.

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
britsta: If I do this inline, I need an additional check whether `typeof response === “number”` which is why I find it easier to use the auxiliary method.

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
britsta: Since I have improved the error messages, I now use `minRating` in tests. I also think it is valid public info.

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
jkoenig134: Ehm, I don't think so? value is any in this method, so it should be fine?!

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
jkoenig134: Ok, then leave it public, I don't care. But put it on top.

packages/content/src/requests/items/formField/settings/DateFormFieldSettings.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface IDateFormFieldSettings extends IFormFieldSettings {}
+
+@type("DateFormFieldSettings")
+export class DateFormFieldSettings extends FormFieldSettings implements IDateFormFieldSettings {
+    public canCreate(): string | undefined {
+        return;
+    }
+
+    public canAccept(response: string | number | boolean | string[]): string | undefined {
+        if (!(typeof response === "string" && CoreDate.from(response).dateTime.isValid)) {
```
jkoenig134, line 18: This would be way easier to read.
```suggestion
        if (typeof response !== "string" || !CoreDate.from(response).dateTime.isValid) {
```

packages/content/src/requests/items/formField/settings/RatingFormFieldSettings.ts
```diff
@@ -0,0 +1,48 @@
...
+
+@type("RatingFormFieldSettings")
+export class RatingFormFieldSettings extends FormFieldSettings implements IRatingFormFieldSettings {
+    public static get minRating(): number {
+        return 1;
+    }
+
+    @serialize()
+    @validate({ min: 5, max: 10 })
+    public maxRating: MaxRating;
```
jkoenig134, line 23:
```suggestion
    public maxRating: number;
```

jkoenig134: and maybe also validate that it's an integer

packages/content/src/requests/items/formField/settings/StringFormFieldSettings.ts
```diff
@@ -0,0 +1,62 @@
...
+    public min?: number;
+
+    @serialize()
+    @validate({ nullable: true })
+    public max?: number;
+
+    public canCreate(): string | undefined {
+        if (this.max && this.min && this.max < this.min) {
+            return "The max cannot be smaller than the min.";
+        }
```
jkoenig134, line 34: Integer check?

packages/content/src/requests/items/formField/settings/BooleanFormFieldSettings.ts
```diff
@@ -0,0 +1,26 @@
+import { ISerializable, Serializable, type } from "@js-soft/ts-serval";
+
+export interface BooleanFormFieldSettingsJSON {
+    "@type": "BooleanFormFieldSettings";
+}
+
+export interface IBooleanFormFieldSettings extends ISerializable {}
```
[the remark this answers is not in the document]
britsta: Good catch!

packages/content/src/requests/items/formField/FormFieldAcceptResponseItem.ts
```diff
@@ -0,0 +1,48 @@
...
+            throw new Error("this should never happen");
+        }
+
+        if (Array.isArray(value.response) && !value.response.every((option) => typeof option === "string")) {
+            throw new ValidationError(
+                FormFieldAcceptResponseItem.name,
+                nameof<FormFieldAcceptResponseItem>((x) => x.response),
+                `If the ${nameof<FormFieldAcceptResponseItem>((x) => x.response)} is an array, it must be a string array.`
+            );
+        }
```
jkoenig134, line 40: actually this should be in the validation function above ..

jkoenig134 (approved):
LGTM, but please wait for @Milena-Czierlinski's review :)

PR #480, opened 2025-04-10
Allow semi-individualized passwords

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [ ] I self-reviewed the PR.


packages/core-types/src/SharedPasswordProtection.ts
```diff
@@ -2,9 +2,30 @@ import { ISerializable, Serializable, serialize, validate } from "@js-soft/ts-se
...
+export enum PasswordLocationIndicatorMedium {
+    RecoveryKit,
+    Self,
+    Letter,
+    RegistrationLetter,
+    Mail,
+    Sms,
+    App,
+    Website
+}
```
Milena-Czierlinski, line 17: I stumbled across numeric enums and thought this would be nice for this use case. However, I'm not so sure anymore, since the runtime schema generator doesn't allow me to use the imported PasswordLocationIndicator. Now, having only numbers available as PasswordLocationIndicator when creating something with a password protection seems to fail the whole attempt of this enum. I guess we'll need a string enum here after all. @jkoenig134 What do you think?

jkoenig134: > I guess we'll need a string enum here after all.

yep

packages/core-types/src/SharedPasswordProtection.ts
```diff
@@ -2,11 +2,40 @@ import { ISerializable, Serializable, serialize, validate } from "@js-soft/ts-se
...
+    return `must be a number between 0 and 99 or one of the following strings: ${Object.values(PasswordLocationIndicatorMedium).join(", ")}`;
+}
+
 export interface ISharedPasswordProtection extends ISerializable {
     passwordType: "pw" | `pin${number}`;
     salt: ICoreBuffer;
+    passwordLocationIndicator?: PasswordLocationIndicator;
 }
 
+// TODO: do we need @type here?
```
Milena-Czierlinski, line 38: I stumbled upon this. @jkoenig134 What is the rule whether we need this or not?

jkoenig134: We need `@type` when we want that ts-serval can deserialize without context.

With this class we **always** run `SharedPasswordProtection.from(...)` so serval knows what to deserialize.

packages/runtime/src/useCases/transport/devices/CreateDeviceOnboardingToken.ts
```diff
@@ -33,14 +37,22 @@ export class CreateDeviceOnboardingTokenUseCase extends UseCase<CreateDeviceOnbo
...
+                  passwordLocationIndicator: request.passwordProtection.passwordLocationIndicator as PasswordLocationIndicator
+              })
+            : undefined;
 
-        const tokenContent = TokenContentDeviceSharedSecret.from({ sharedSecret });
         const token = await this.tokenController.sendToken({
             content: tokenContent,
             expiresAt: expiresAt,
             ephemeral: true,
-            passwordProtection: PasswordProtectionCreationParameters.create(request.passwordProtection)
```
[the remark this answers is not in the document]
jkoenig134: please inline that again, same with the other occurences.

packages/runtime/src/useCases/transport/relationshipTemplates/CreateTokenForOwnRelationshipTemplate.ts
```diff
@@ -58,17 +59,25 @@ export class CreateTokenForOwnTemplateUseCase extends UseCase<CreateTokenForOwnT
...
+                  passwordLocationIndicator: request.passwordProtection.passwordLocationIndicator as PasswordLocationIndicator
+              })
+            : undefined;
+
+        if (template.passwordProtection && !template.passwordProtection.matchesCreationParameters(tokenPasswordProtection)) {
             return Result.fail(RuntimeErrors.relationshipTemplates.passwordProtectionMustBeInherited());
         }
 
-        const passwordProtection = template.passwordProtection ? SharedPasswordProtection.from(template.passwordProtection) : undefined;
+        const templatePasswordProtection = template.passwordProtection ? SharedPasswordProtection.from(template.passwordProtection) : undefined;
```
jkoenig134, line 74: this is now a bit of a mess. this has to be inlined.

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
```diff
@@ -72,10 +74,21 @@ export class GetRelationshipTemplatesUseCase extends UseCase<GetRelationshipTemp
                 if (input === "!") {
                     query[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.passwordType)}`] = "pw";
                 }
+            },
+            [`${nameof<RelationshipTemplateDTO>((r) => r.passwordProtection)}.passwordLocationIndicator`]: (query: any, input: string) => {
+                if (this.stringContainsNumber(input)) {
+                    query[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.passwordLocationIndicator)}`] = Number(input);
```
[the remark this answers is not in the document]
jkoenig134: same everywhere else.

jkoenig134 (changes requested):

packages/runtime/src/useCases/transport/devices/CreateDeviceOnboardingToken.ts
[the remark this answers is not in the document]
jkoenig134: I think a typecast should solve that.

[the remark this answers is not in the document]
Milena-Czierlinski: Isn't that basically what I've done? :) What exactly do you suggest?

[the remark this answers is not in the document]
jkoenig134: putting it back as a one liner and not introduce so much code for a small feature.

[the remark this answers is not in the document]
Milena-Czierlinski: I would love to do that, but I don't see how to make it happen. If you have a specific solution in mind, that solves the compiler errors I described, please share it. :D

packages/runtime/src/useCases/common/Schemas.ts
```diff
@@ -19981,12 +19981,14 @@ export const CreateDeviceOnboardingTokenRequest: any = {
...
-                            "type": "string"
+                            "type": "string",
+                            "minLength": 1
                         },
                         "passwordIsPin": {
                             "type": "boolean",
                             "const": true
-                        }
+                        },
+                        "passwordLocationIndicator": {}
```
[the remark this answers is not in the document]
Milena-Czierlinski: Hmm, on the other hand what else do we expect for typing it unknown?

packages/app-runtime/src/extensibility/ui/IUIBridge.ts
```diff
@@ -11,5 +12,5 @@ export interface IUIBridge {
     showRequest(account: LocalAccountDTO, request: LocalRequestDVO): Promise<Result<void>>;
     showError(error: UserfriendlyApplicationError, account?: LocalAccountDTO): Promise<Result<void>>;
     requestAccountSelection(possibleAccounts: LocalAccountDTO[], title?: string, description?: string): Promise<Result<LocalAccountDTO | undefined>>;
-    enterPassword(passwordType: "pw" | "pin", pinLength?: number, attempt?: number): Promise<Result<string>>;
+    enterPassword(passwordType: "pw" | "pin", pinLength?: number, attempt?: number, passwordLocationIndicator?: PasswordLocationIndicator): Promise<Result<string>>;
```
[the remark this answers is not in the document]
Magnus-Kuhn: Why are we offering both options anyway? Who wants to use numbers over strings?
It would also be possible to perform the string-number-mapping in the `PasswordProtectionCreationParameters`, that way most things would become numbers - if only numbers is preferred.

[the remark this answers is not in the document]
Milena-Czierlinski: @Magnus-Kuhn We offer the strings for the integrator for convenience and the numbers for flexibility. In the code I think it is nice to keep both types for us developers. However, in the truncated references we save data by mapping the strings to numbers.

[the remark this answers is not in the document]
jkoenig134: But in the UI Code it's just harder to also process strings, I'd be happy if we remove it from here.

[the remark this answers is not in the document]
Magnus-Kuhn: I think it's harder for us to keep in mind that there are both a number and a string meaning the same thing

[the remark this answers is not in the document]
jkoenig134: I don't care how hard it is for us, it has to be easy for the integrator.

@Magnus-Kuhn if you have a strong opinion on this topic let's discuss this elsewhere, I don't think this leads anywhere.

packages/core-types/src/SharedPasswordProtection.ts
```diff
@@ -24,20 +56,45 @@ export class SharedPasswordProtection extends Serializable implements ISharedPas
...
+    }
+
+    private static mapNumberToPasswordLocationIndicatorMedium(value: number): PasswordLocationIndicator {
+        const passwordLocationIndicatorMediumValues = Object.values(PasswordLocationIndicatorMedium);
+
+        if (value >= 0 && value < passwordLocationIndicatorMediumValues.length) {
+            return passwordLocationIndicatorMediumValues[value];
+        }
+
+        return value as PasswordLocationIndicator;
```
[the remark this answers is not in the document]
Milena-Czierlinski: @jkoenig134 and I had this discussion regarding numbers lower than 50 that don't have a representation in the PasswordLocationIndicatorMedium and decided, that we don't want to catch this, since the App will just show the default. We could handle it the same way here. 🤔

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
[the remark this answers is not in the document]
Milena-Czierlinski: I refactored it, but if I don't use `Number(input)` I get an error. Should we investigate this further?

packages/core-types/src/SharedPasswordProtection.ts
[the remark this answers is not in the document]
jkoenig134: I think we will run into backwards compatibility issues when we validate it too strict. Let's keep it as-is.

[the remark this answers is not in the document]
jkoenig134: Ehm, I don't get you.. this is not the salt, this is the PasswordLocationIndicator and there should be no influence.

I'll discuss details with this file with @Milena-Czierlinski in a call, if you want you can join.

packages/runtime/src/useCases/transport/relationshipTemplates/CreateTokenQRCodeForOwnRelationshipTemplate.ts
```diff
@@ -23,9 +23,13 @@ export interface CreateTokenQRCodeForOwnTemplateRequest {
...
         password: string;
         passwordIsPin?: true;
+        passwordLocationIndicator?: unknown;
     };
 }
 
+export type CreateTokenQRCodeForOwnTemplateRequest = SchemaValidatableCreateTokenQRCodeForOwnTemplateRequest & {
+    passwordProtection?: { passwordLocationIndicator?: PasswordLocationIndicator };
+};
 class Validator extends TokenAndTemplateCreationValidator<CreateTokenQRCodeForOwnTemplateRequest> {
```
jkoenig134, line 33: missing newline

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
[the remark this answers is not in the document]
jkoenig134: Also you can use isNan to determine if the string is numeric or not:

```txt
> parseSafe = (x) => isNaN(x) ? x : parseInt(x)
[Function: parseSafe]
> parseSafe("1.23543")
1
> parseSafe("a")
'a'
```

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN for usage.

packages/runtime/src/useCases/common/validation/TokenAndTemplateCreationValidator.ts
```diff
@@ -41,6 +42,26 @@ export class TokenAndTemplateCreationValidator<
...
             }
+
+            if (passwordProtection.passwordLocationIndicator) {
+                if (
+                    (typeof passwordProtection.passwordLocationIndicator === "string" &&
+                        (!Object.values(PasswordLocationIndicatorMedium).includes(passwordProtection.passwordLocationIndicator as PasswordLocationIndicatorMedium) ||
+                            passwordProtection.passwordLocationIndicator === PasswordLocationIndicatorMedium.RecoveryKit)) ||
+                    (typeof passwordProtection.passwordLocationIndicator === "number" &&
+                        passwordProtection.passwordLocationIndicator < 50 &&
+                        passwordProtection.passwordLocationIndicator > 99)
```
jkoenig134, line 53: please extract some of these checks into a boolean variable. This will be way easier to read.

packages/runtime/src/useCases/common/validation/TokenAndTemplateCreationValidator.ts
```diff
@@ -41,6 +42,26 @@ export class TokenAndTemplateCreationValidator<
                     );
                 }
             }
+
+            if (passwordProtection.passwordLocationIndicator) {
+                if (
+                    (typeof passwordProtection.passwordLocationIndicator === "string" &&
+                        (!Object.values(PasswordLocationIndicatorMedium).includes(passwordProtection.passwordLocationIndicator as PasswordLocationIndicatorMedium) ||
```
jkoenig134, line 49: I think if you use Object.keys instead of Object.values you don't need the typecast.

packages/runtime/src/useCases/common/validation/TokenAndTemplateCreationValidator.ts
Milena-Czierlinski: Fair. I had such a version, but then used this since it was shorter. But I'll refactor it back. :D

packages/core-types/src/SharedPasswordProtection.ts
```diff
@@ -2,9 +2,37 @@ import { ISerializable, Serializable, serialize, validate } from "@js-soft/ts-se
...
+type Enumerate<N extends number, Acc extends number[] = []> = Acc["length"] extends N ? Acc[number] : Enumerate<N, [...Acc, Acc["length"]]>;
+type IntRange<From extends number, To extends number> = Exclude<Enumerate<To>, Enumerate<From>>;
+
+export enum PasswordLocationIndicatorMedium {
+    RecoveryKit = "RecoveryKit",
+    Self = "Self",
+    Letter = "Letter",
+    RegistrationLetter = "RegistrationLetter",
+    Mail = "Mail",
+    Sms = "Sms",
```
britsta, line 14: The same applies to each occurrence.

```suggestion
    SMS = "SMS",
```

Milena-Czierlinski: Mhh subjectively I prefer `Sms` over `SMS`. It matches our variable casing, although SMS is correct in natural language. Can we say which casing is better regarding some objective measure?

jkoenig134: SMS is correct, but you can also use sms

Milena-Czierlinski: Okay, I'll change the enum value to SMS. What about the string? Should we allow any casing? Same for the other strings, i.e. should an integrator be allowed to write "letter" and "Letter"?

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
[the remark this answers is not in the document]
Milena-Czierlinski: isNan expects a number as input in TypeScript.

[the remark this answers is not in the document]
jkoenig134: ehm ... it doesn't!

As I just saw, please also use Number.isNaN instead of global isNaN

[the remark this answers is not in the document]
jkoenig134: That should do that trick.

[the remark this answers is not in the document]
jkoenig134: Argh, no it doesn't: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN#difference_between_number.isnan_and_global_isnan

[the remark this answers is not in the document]
Milena-Czierlinski: So should I just leave it as it is? :)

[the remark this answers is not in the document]
jkoenig134: I think there is a utility method from lodash we could use.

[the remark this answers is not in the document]
jkoenig134: ... leave it as is.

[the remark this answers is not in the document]
Milena-Czierlinski: I just found lodash's `isNumeric()` 😅.

packages/runtime/src/useCases/index.ts
```diff
@@ -1,7 +1,9 @@
 export * from "./anonymous";
+// TODO: why don't we just export common?
 export * from "./common/AbstractCorrelator";
 export * from "./common/Base64ForIdPrefix";
 export * from "./common/OwnerRestriction";
+export * from "./common/PasswordLocationIndicator";
 export * from "./common/RuntimeErrors";
 export * from "./consumption";
```
Milena-Czierlinski, line 8: There is an `index` file in common. So can we just export common here? @jkoenig134

jkoenig134: circular dependencies.

packages/core-types/test/SharedPasswordProtection.test.ts
```diff
@@ -0,0 +1,114 @@
...
+                salt: CoreBuffer.random(16)
+            });
+
+            const truncatedSharedPasswordProtection = sharedPasswordProtection.truncate();
+            const splittedPasswordParts = truncatedSharedPasswordProtection.split("&");
+            expect(splittedPasswordParts).toHaveLength(2);
+        });
+    });
+
+    describe("deserialization", () => {
```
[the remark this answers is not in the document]
Milena-Czierlinski: Is this still necessary, now that we have moved the conversion?

packages/app-runtime/test/runtime/AppStringProcessor.test.ts
```diff
@@ -192,6 +193,36 @@ describe("AppStringProcessor", function () {
...
+            expiresAt: CoreDate.utc().add({ days: 1 }).toISOString(),
+            passwordProtection: { password: "password", passwordLocationIndicator: "SMS" }
+        });
+
+        mockUiBridge.setPasswordToReturnForAttempt(1, "password");
+        mockUiBridge.accountIdToReturn = runtime2SessionA.account.id;
+
+        await runtime2.stringProcessor.processTruncatedReference(templateResult.value.truncatedReference);
+
+        expect(mockUiBridge).enterPasswordCalled("pw", undefined, undefined, PasswordLocationIndicatorOptions.SMS);
```
jkoenig134, line 208: shouldn't this be the index?

Milena-Czierlinski: It is 🤔

jkoenig134: Didn't saw the other change :)

packages/runtime/src/useCases/common/validation/TokenAndTemplateCreationValidator.ts
```diff
@@ -41,8 +43,37 @@ export class TokenAndTemplateCreationValidator<
                     );
                 }
             }
+
+            if (passwordProtection.passwordLocationIndicator && !TokenAndTemplateCreationValidator.isValidPasswordLocationIndicator(passwordProtection.passwordLocationIndicator)) {
+                validationResult.addFailure(
+                    new ValidationFailure(
+                        RuntimeErrors.general.invalidPropertyValue(
+                            `must be a number from 50 to 99 or one of the following strings: ${Object.values(PasswordLocationIndicatorOptions).slice(1).join(", ")}`
```
jkoenig134, line 51: `.slice(1)` will break as someone moves the items. Please filter out `RecoveryKit` instead.

packages/runtime/src/useCases/transport/devices/CreateDeviceOnboardingToken.ts
```diff
@@ -33,9 +41,10 @@ export class CreateDeviceOnboardingTokenUseCase extends UseCase<CreateDeviceOnbo
 
     protected async executeInternal(request: CreateDeviceOnboardingTokenRequest): Promise<Result<TokenDTO>> {
         const sharedSecret = await this.devicesController.getSharedSecret(CoreId.from(request.id), request.profileName);
+        const tokenContent = TokenContentDeviceSharedSecret.from({ sharedSecret });
```
jkoenig134, line 44: why does tokenContent doesn't get its own line like it had before, but expiresAt does?! All these refactorings .........

packages/runtime/src/useCases/transport/identityRecoveryKits/CreateIdentityRecoveryKit.ts
```diff
@@ -43,11 +51,14 @@ export class CreateIdentityRecoveryKitUseCase extends UseCase<CreateIdentityReco
 
         const newBackupDevice = await this.devicesController.sendDevice({ isAdmin: true, isBackupDevice: true, name: "Backup Device" });
         const sharedSecret = await this.devicesController.getSharedSecret(newBackupDevice.id, request.profileName);
+
+        const passwordProtection = { ...request.passwordProtection, passwordLocationIndicator: "RecoveryKit" as PasswordLocationIndicator };
```
jkoenig134, line 55: Why do you have to cast?

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
```diff
@@ -72,6 +74,11 @@ export class GetRelationshipTemplatesUseCase extends UseCase<GetRelationshipTemp
                 if (input === "!") {
                     query[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.passwordType)}`] = "pw";
                 }
+            },
+            [`${nameof<RelationshipTemplateDTO>((r) => r.passwordProtection)}.passwordLocationIndicator`]: (query: any, input: string) => {
+                const stringIsNumeric = /^\d+$/.test(input);
+                const queryInput = stringIsNumeric ? parseInt(input) : PasswordLocationIndicatorOptions[input as keyof typeof PasswordLocationIndicatorOptions];
```
jkoenig134: I don't get the `PasswordLocationIndicatorOptions[input as keyof typeof PasswordLocationIndicatorOptions]` part.. What happens if nothing is found b/c it's a random string?

packages/runtime/src/types/transport/PasswordProtectionDTO.ts
```diff
@@ -1,4 +1,7 @@
+import { PasswordLocationIndicator } from "@nmshd/transport";
```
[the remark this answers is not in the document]
jkoenig134: yep, please move all that to core-types, then it can be code-shared by the connector.

packages/runtime/src/useCases/transport/tokens/GetTokens.ts
```diff
@@ -62,6 +64,11 @@ export class GetTokensUseCase extends UseCase<GetTokensRequest, TokenDTO[]> {
                 if (input === "!") {
                     query[`${nameof<Token>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.passwordType)}`] = "pw";
                 }
+            },
+            [`${nameof<TokenDTO>((r) => r.passwordProtection)}.passwordLocationIndicator`]: (query: any, input: string) => {
+                const stringIsNumeric = /^\d+$/.test(input);
+                const queryInput = stringIsNumeric ? parseInt(input) : PasswordLocationIndicatorOptions[input as keyof typeof PasswordLocationIndicatorOptions];
```
jkoenig134: same issues as the other query method.

packages/runtime/test/transport/passwordProtection/passwordLocationIndicator.test.ts
jkoenig134, line 1: that's a strange filename. You are unit testing the TokenAndTemplateCreationValidator here, so please move that to the proper file.

packages/transport/src/core/types/PasswordProtectionCreationParameters.ts
```diff
@@ -1,8 +1,25 @@
...
+export enum PasswordLocationIndicatorOptions {
+    RecoveryKit = 0,
+    Self = 1,
+    Letter = 2,
+    RegistrationLetter = 3,
+    Email = 4,
+    SMS = 5,
+    App = 6,
+    Website = 7
+}
```
jkoenig134, line 15: this has to be moved away from here as it has nothing to do with the transport lib anymore.

packages/transport/src/core/types/PasswordProtectionCreationParameters.ts
```diff
@@ -1,8 +1,25 @@
...
+
+export type PasswordLocationIndicator = keyof typeof PasswordLocationIndicatorOptions | IntRange<50, 100>;
```
jkoenig134, line 17: this too.

packages/transport/src/core/types/PasswordProtection.ts
```diff
@@ -22,25 +23,36 @@ export class PasswordProtection extends Serializable implements IPasswordProtect
...
             passwordType: this.passwordType,
-            salt: this.salt
+            salt: this.salt,
+            passwordLocationIndicator: this.passwordLocationIndicator
         });
     }
 
-    public matchesInputForNewPasswordProtection(newPasswordProtection: { password: string; passwordIsPin?: true } | undefined): boolean {
+    public matchesInputForNewPasswordProtection(
+        newPasswordProtection: { password: string; passwordIsPin?: true; passwordLocationIndicator?: PasswordLocationIndicator } | undefined
```
jkoenig134, line 43: I think this should accept a number instead of a PasswordLocationIndicator

jkoenig134 (changes requested):

packages/runtime/src/useCases/transport/identityRecoveryKits/CreateIdentityRecoveryKit.ts
Milena-Czierlinski: Because below PasswordProtectionCreationParameters.create expects the type PasswordLocationIndicator and not string.

Milena-Czierlinski: Well, apparently the compiler needs it more explicitly anyhow.

jkoenig134: Hm, why don't you use 
```suggestion
        const passwordProtection = { ...request.passwordProtection, passwordLocationIndicator: PasswordLocationIndicator.RecoveryKit };
```
?

Milena-Czierlinski: Because PasswordLocationIndicator refers to a type. Maybe you mean PasswordLocationIndicatorOptions.RecoveryKit. However, that is equal to 0 and thus not included in the type PasswordLocationIndicator, where the numbers only start at 50.

jkoenig134: eeeeeeeh

packages/transport/src/core/types/PasswordProtection.ts
Milena-Czierlinski: Why? I don't see any advantage and it would make the code more complex.

packages/transport/src/core/types/PasswordProtectionCreationParameters.ts
```diff
@@ -14,16 +31,33 @@ export class PasswordProtectionCreationParameters extends Serializable implement
...
+            passwordType: params.passwordIsPin ? `pin${params.password.length}` : "pw",
+            passwordLocationIndicator: passwordLocationIndicator
         });
     }
+
+    private static mapPasswordLocationIndicatorOptionToNumber(value: PasswordLocationIndicator): number {
+        if (typeof value === "number") return value;
+
+        const numericValue = PasswordLocationIndicatorOptions[value];
+        return numericValue;
```
jkoenig134, line 46: this should throw when not found.

packages/transport/src/core/types/PasswordProtection.ts
Milena-Czierlinski: Nah, I changed it again, to reduce the amount of changes. The conversion happens in PasswordProtectionCreationParameters.create. And this is called by matchesInputForNewPasswordProtection.

packages/runtime/src/useCases/transport/devices/CreateDeviceOnboardingToken.ts
Milena-Czierlinski: Because sharedSecret is needed for tokenContent. 

> All these refactorings .........

😒

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
Milena-Czierlinski: For LokiJS it worked as expected, for MongoDB it didn't. However, eslint seems to have a problem with it, that doesn't seem to be completely valid for me. Even though I don't want to claim that I solved the problem in the prettiest way possible.

packages/runtime/src/useCases/common/PasswordProtectionMapper.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    private static mapNumberToPasswordLocationIndicatorOption(value: number): PasswordLocationIndicator {
+        const passwordLocationIndicatorOptions = Object.values(PasswordLocationIndicatorOptions);
+
+        if (value >= 0 && value < passwordLocationIndicatorOptions.length) {
+            return PasswordLocationIndicatorOptions[value] as PasswordLocationIndicator;
+        }
+
+        return value as PasswordLocationIndicator;
```
jkoenig134, line 26:
```suggestion
        return PasswordLocationIndicatorOptions[number] ?? number;
```

in this case you can just inline it.

packages/runtime/src/useCases/transport/relationshipTemplates/GetRelationshipTemplates.ts
```diff
@@ -72,6 +75,15 @@ export class GetRelationshipTemplatesUseCase extends UseCase<GetRelationshipTemp
...
                 }
+            },
+            [`${nameof<RelationshipTemplateDTO>((r) => r.passwordProtection)}.passwordLocationIndicator`]: (query: any, input: string) => {
+                let queryInput = -1;
+
+                const stringIsNumeric = /^\d+$/.test(input);
+                if (stringIsNumeric) queryInput = parseInt(input);
+                else if (input in PasswordLocationIndicatorOptions) queryInput = PasswordLocationIndicatorOptions[input as keyof typeof PasswordLocationIndicatorOptions];
+
+                query[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.passwordLocationIndicator)}`] = queryInput;
```
jkoenig134, line 86: I'd prefer this:

```suggestion
                const stringIsNumeric = /^\d+$/.test(input);
                const queryValue = stringIsNumeric ? parseInt(input) : ((PasswordLocationIndicatorOptions[input as any] as any) ?? -1);

                query[`${nameof<RelationshipTemplate>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.passwordLocationIndicator)}`] = queryValue;
```

packages/runtime/src/useCases/transport/tokens/GetTokens.ts
```diff
@@ -62,6 +65,15 @@ export class GetTokensUseCase extends UseCase<GetTokensRequest, TokenDTO[]> {
...
                 }
+            },
+            [`${nameof<TokenDTO>((r) => r.passwordProtection)}.passwordLocationIndicator`]: (query: any, input: string) => {
+                let queryInput = -1;
+
+                const stringIsNumeric = /^\d+$/.test(input);
+                if (stringIsNumeric) queryInput = parseInt(input);
+                else if (input in PasswordLocationIndicatorOptions) queryInput = PasswordLocationIndicatorOptions[input as keyof typeof PasswordLocationIndicatorOptions];
+
+                query[`${nameof<Token>((t) => t.passwordProtection)}.${nameof<PasswordProtection>((t) => t.passwordLocationIndicator)}`] = queryInput;
```
jkoenig134, line 76: see other

packages/transport/src/core/TransportCoreErrors.ts
```diff
@@ -227,6 +227,13 @@ class General {
     public noPasswordProvided() {
         return new CoreError("error.transport.noPasswordProvided", "You need to provide a password to perform this operation.");
     }
+
+    public invalidPasswordLocationIndicator(value: unknown) {
+        return new CoreError(
+            "error.transport.invalidPasswordLocationIndicator",
+            `${value} is not a valid PasswordLocationIndicator. Instead, it must be a number from 50 to 99 or one of the following strings: ${Object.values(PasswordLocationIndicatorOptions).join(", ")}`
+        );
+    }
```
jkoenig134, line 236: also this is no transport logic.

packages/transport/src/core/types/PasswordProtection.ts
jkoenig134: I'd still prefer if transport doesn't care about the passwordLocationIndicator, but we can refactor this later, it will only be under the hood.

jkoenig134 (changes requested):

jkoenig134 (approved):

PR #485, opened 2025-04-22
Add sendMailDisabled to RelationshipDVO

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/runtime/src/dataViews/DataViewExpander.ts
+++ b/packages/runtime/src/dataViews/DataViewExpander.ts
@@ -1795,7 +1795,7 @@ export class DataViewExpander {
 
         const sendMailDisabledResult = await this.consumption.attributes.getPeerSharedAttributes({
             peer: relationship.peer,
-            query: { "content.value.@type": "Consent", "content.isTechnical": "true", "content.key": "__App_Contact_sendMailDisabled" }
+            query: { "content.value.@type": "Consent", "content.key": "__App_Contact_sendMailDisabled" }
         });
         const sendMailDisabled = sendMailDisabledResult.value.length > 0;
 
--- a/packages/runtime/src/dataViews/DataViewExpander.ts
+++ b/packages/runtime/src/dataViews/DataViewExpander.ts
@@ -1793,6 +1793,12 @@ export class DataViewExpander {
             }
         }
 
+        const sendMailDisabledResult = await this.consumption.attributes.getPeerSharedAttributes({
+            peer: relationship.peer,
+            query: { "content.value.@type": "Consent", "content.isTechnical": "true", "content.key": "__App_Contact_sendMailDisabled" }
+        });
+        const sendMailDisabled = sendMailDisabledResult.value.length > 0;
+
         let direction = RelationshipDirection.Incoming;
         if (!relationship.template.isOwn) {
             direction = RelationshipDirection.Outgoing;
@@ -1857,7 +1863,8 @@ export class DataViewExpander {
             nameMap: stringByType,
             templateId: relationship.template.id,
             auditLog: relationship.auditLog,
-            creationContent: relationship.creationContent
+            creationContent: relationship.creationContent,
+            sendMailDisabled
         };
     }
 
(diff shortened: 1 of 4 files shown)


packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -1793,6 +1793,12 @@ export class DataViewExpander {
             }
         }
 
+        const sendMailDisabledResult = await this.consumption.attributes.getPeerSharedAttributes({
+            peer: relationship.peer,
+            query: { "content.value.@type": "Consent", "content.isTechnical": "true", "content.key": "__App_Contact_sendMailDisabled" }
```
jkoenig134, line 1798:
```suggestion
            query: { "content.value.@type": "Consent", "content.key": "__App_Contact_sendMailDisabled" }
```

jkoenig134: The consent is not technical. If it were technical the user would not see it after it was accepted.

packages/runtime/test/lib/testUtils.ts
```diff
@@ -398,42 +397,16 @@ export async function establishRelationship(transportServices1: TransportService
...
 
-    const createRelationshipResponse = await runtimeServices2.transport.relationships.createRelationship({
-        templateId: template.id,
-        creationContent: creationContent ?? emptyRelationshipCreationContent
-    });
-    expect(createRelationshipResponse).toBeSuccessful();
+    const requestId = (await runtimeServices2.consumption.incomingRequests.getRequests({ query: { "source.reference": template.id } })).value[0].id;
+
+    const acceptRequestResult = await runtimeServices2.consumption.incomingRequests.accept({ requestId, items: acceptRequestItemsParameters });
+    expect(acceptRequestResult).toBeSuccessful();
```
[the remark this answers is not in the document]
Milena-Czierlinski: The RelationshipDVO test is the only place where this testUtil is used. Its main difference is that a content for the Template can be provided.

packages/runtime/test/dataViews/RelationshipDVO.test.ts
```diff
@@ -1,15 +1,15 @@
...
+import { establishRelationshipWithContents, executeFullCreateAndShareRelationshipAttributeFlow, RuntimeServiceProvider, TestRuntimeServices } from "../lib";
 
 const serviceProvider = new RuntimeServiceProvider();
 
 let runtimeServices1: TestRuntimeServices;
 let runtimeServices2: TestRuntimeServices;
 
 beforeAll(async () => {
-    const runtimeServices = await serviceProvider.launch(2);
+    const runtimeServices = await serviceProvider.launch(2, { enableRequestModule: true, enableDeciderModule: true });
```
jkoenig134, line 12: please disable the modules and undo that changes in testUtil

packages/runtime/test/lib/testUtils.ts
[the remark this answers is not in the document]
jkoenig134: Yeah, but the code was perfectly fine. We don't want to have the modules enabled in that test. Also - this is a refactoring that shouldn't be in this PR.

packages/runtime/test/dataViews/RelationshipDVO.test.ts
Milena-Czierlinski: Why? There are many tests that use the modules.

jkoenig134: >There are many tests that use the modules.

which doesn't mean that it's correct.

Milena-Czierlinski: Can you please elaborate what is wrong about using these modules?

jkoenig134 (approved):

packages/runtime/src/dataViews/DataViewExpander.ts
Milena-Czierlinski: Okay, no problemo. Please check if the consent it not technical in the example you showed me earlier. I kinda thought it was. :)

packages/runtime/test/dataViews/RelationshipDVO.test.ts
jkoenig134: We want to test stuff in isolation and not with modules involved that aren't tested rn.

jkoenig134: But I don't have time for that discussion rn and the feature is too important to wait.

packages/runtime/src/dataViews/DataViewExpander.ts
jkoenig134: It was, but I removed that.

PR #500, opened 2025-04-30
It is possible to set `0` as PasswordLocationIndicator

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/runtime/src/useCases/common/validation/TokenAndTemplateCreationValidator.ts
+++ b/packages/runtime/src/useCases/common/validation/TokenAndTemplateCreationValidator.ts
@@ -43,7 +43,10 @@ export class TokenAndTemplateCreationValidator<
                 }
             }
 
-            if (passwordProtection.passwordLocationIndicator && !TokenAndTemplateCreationValidator.isValidPasswordLocationIndicator(passwordProtection.passwordLocationIndicator)) {
+            if (
+                passwordProtection.passwordLocationIndicator !== undefined &&
+                !TokenAndTemplateCreationValidator.isValidPasswordLocationIndicator(passwordProtection.passwordLocationIndicator)
+            ) {
                 validationResult.addFailure(
                     new ValidationFailure(
                         RuntimeErrors.general.invalidPropertyValue(
--- a/packages/runtime/test/common/TokenAndTemplateCreationValidator.test.ts
+++ b/packages/runtime/test/common/TokenAndTemplateCreationValidator.test.ts
@@ -42,6 +42,11 @@ describe("TokenAndTemplateCreationValidator", () => {
             expect(result).toBe(false);
         });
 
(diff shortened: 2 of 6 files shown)


packages/runtime/src/useCases/common/validation/TokenAndTemplateCreationValidator.ts
```diff
@@ -43,7 +43,10 @@ export class TokenAndTemplateCreationValidator<
                 }
             }
 
-            if (passwordProtection.passwordLocationIndicator && !TokenAndTemplateCreationValidator.isValidPasswordLocationIndicator(passwordProtection.passwordLocationIndicator)) {
+            if (
+                passwordProtection.passwordLocationIndicator !== undefined &&
```
jkoenig134, line 47: please use typeof for this check

Milena-Czierlinski: No problem. :) What is the advantage of that?

jkoenig134: I heard that it's safer than comparing with undefined. And we use this .. everywhere and discussed that many times ;)

jkoenig134: If we decide to do this different we should do that everywhere.

jkoenig134: Maybe we should change that everywhere, as it's easier to read and should be safe for all our usecases: https://stackoverflow.com/a/4725697/10029378

Milena-Czierlinski: > And we use this .. everywhere and discussed that many times ;)

We currently use both versions about equally often. We used to use `if (typeof x !== "undefined")` everywhere until you said you changed your mind and it is easier to read if we use `if (!x)`. ;)

jkoenig134: `if (!x)` and `if (typeof x !== "undefined")` are NOT equivalent as you see here.

jkoenig134 (approved):

packages/runtime/src/useCases/common/validation/TokenAndTemplateCreationValidator.ts
Milena-Czierlinski: Yes, I know. I just wanted to explain why the answer to my question above wasn't as apparent as you made it seem in your initial response.

jkoenig134: With https://github.com/nmshd/runtime/pull/502 this discussion shouldn't occur anymore :)

PR #504, opened 2025-05-05
Validate tags in CanCreateRepositoryAttribute

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/consumption/src/modules/attributes/AttributesController.ts
+++ b/packages/consumption/src/modules/attributes/AttributesController.ts
@@ -225,6 +225,9 @@ export class AttributesController extends ConsumptionBaseController {
 
         const parsedParams = CreateRepositoryAttributeParams.from(params);
 
+        const tagValidationResult = await this.validateTags(parsedParams.content);
+        if (tagValidationResult.isError()) throw tagValidationResult.error;
+
         const trimmedAttribute = {
             ...parsedParams.content.toJSON(),
             value: this.trimAttributeValue(parsedParams.content.value.toJSON() as AttributeValues.Identity.Json)
--- a/packages/consumption/src/modules/attributes/AttributesController.ts
+++ b/packages/consumption/src/modules/attributes/AttributesController.ts
@@ -225,9 +225,6 @@ export class AttributesController extends ConsumptionBaseController {
 
         const parsedParams = CreateRepositoryAttributeParams.from(params);
 
-        const tagValidationResult = await this.validateTags(parsedParams.content);
-        if (tagValidationResult.isError()) throw tagValidationResult.error;
-
         const trimmedAttribute = {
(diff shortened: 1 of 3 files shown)


packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -225,9 +225,6 @@ export class AttributesController extends ConsumptionBaseController {
 
         const parsedParams = CreateRepositoryAttributeParams.from(params);
 
-        const tagValidationResult = await this.validateTags(parsedParams.content);
-        if (tagValidationResult.isError()) throw tagValidationResult.error;
-
```
Milena-Czierlinski, line 230: I'm not sure whether we should keep or remove this here. Principally, the validation is called everywhere, before calling this function, so we don't need this here. However, keeping it would add another safety layer. But then again, we don't have a re-check in this function for e.g. the prevent of duplicates. @jkoenig134 What do you mean?

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -912,6 +912,26 @@ describe(CanCreateRepositoryAttributeUseCase.name, () => {
         const result = await services1.consumption.attributes.canCreateRepositoryAttribute(canCreateAttributeWithoutOptionalPropertyRequest);
         expect(result.value.isSuccess).toBe(true);
     });
+
+    test("should not allow to create a RepositoryAttribute with invalid tags", async () => {
```
jkoenig134, line 916: the test name indicates that you run createRepositoryAttribute here, but you call canCreateRepositoryAttribute. Please make sure that this gets clear from the test name.

Milena-Czierlinski: The test name uses the same pattern than all other tests in the CanCreateRepositoryAttribute describe block. If all of them should be changed, what would you suggest?

jkoenig134: I do not talk about all the other tests, because that's not the scope of this PR.

Milena-Czierlinski: I wouldn't change the names of the other tests within this PR. However, having a naming that matches the environment IS part of this PR. For this reason, I think it is valid to talk about it here.

jkoenig134 (approved):

packages/runtime/test/consumption/attributes.test.ts
jkoenig134: > Please don't resolve conversations that aren't resolved.

For me this is using brain capacity and time I do not have. So it is resolved. I explained why it's bad naming, if you do not agree just merge as is.

PR #507, opened 2025-05-05
Validate tags of Files

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

The validation of tags in the TransferFileOwnershipRequestItemProcessor should never return an error, since the tags are already validated upon the upload of the file. However, we have this redundant validation also for the tags of attributes in the respective processors, so I added them here, as well. But, I can't write tests for them, since uploading a file with invalid tags is prevented.


Milena-Czierlinski:
The refactoring is handled in a separate PR: #506 .

packages/runtime/src/useCases/transport/files/UploadOwnFile.ts
```diff
@@ -62,13 +63,19 @@ export class UploadOwnFileUseCase extends UseCase<UploadOwnFileRequest, FileDTO>
...
         @Inject validator: Validator
     ) {
         super(validator);
         validator.maxFileSize = fileController.config.platformMaxUnencryptedFileSize;
     }
 
     protected async executeInternal(request: UploadOwnFileRequest): Promise<Result<FileDTO>> {
+        if (request.tags && request.tags.length > 0) {
+            const tagValidationResult = await this.attributesController.validateTagsForType(request.tags, "IdentityFileReference");
+            if (tagValidationResult.isError()) return Result.fail(tagValidationResult.error);
```
jkoenig134, line 76: sus. Why are files validated with attribute logic?

Milena-Czierlinski: Because the tags logic is in the AttributesController and not in a separate controller, which @sebbi08 suggested. And in the end, we only want to allow tags that configured for the Attribute value type IdentityFileReference.

jkoenig134: Hmm, I think we should only validate this for the TransferFileOwnershipRequestItem then...

Milena-Czierlinski: I see that differently. On the one hand, did we only introduce tags on File such that they can be converted to tags of IdentityFilfeReference Attributes. So, the validation should be performed on every uploaded file. On the other hand, would it be really annoying if an integrator uploaded a number of Files and then wants to transfer their ownership and only gets this error in the later step, which would lead to more work on their end.

jkoenig134: As we have to re-think the whole tag format, go.

jkoenig134: > So, the validation should be performed on every uploaded file.

I do not second that. If you send tags to a file that are not compatible with our request format then you're the one that's fd up, not us.

PR #524, opened 2025-05-07
Add the backboneBaseUrl to all truncated references

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/runtime/src/ConfigHolder.ts
+++ b/packages/runtime/src/ConfigHolder.ts
@@ -1,0 +1,13 @@
+import { RuntimeConfig } from "./RuntimeConfig";
+
+export class ConfigHolder {
+    readonly #config: RuntimeConfig;
+
+    public getConfig(): RuntimeConfig {
+        return this.#config;
+    }
+
+    public constructor(config: RuntimeConfig) {
+        this.#config = config;
+    }
+}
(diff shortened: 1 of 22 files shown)


tnotheis (approved):

packages/runtime/src/ConfigHolder.ts
```diff
@@ -0,0 +1,13 @@
+import { RuntimeConfig } from "./RuntimeConfig";
+
+export class ConfigHolder {
+    readonly #config: RuntimeConfig;
```
Milena-Czierlinski, line 4: @jkoenig134 Out of interest, why is there a `#` in the variable name?

jkoenig134: This variable is private - js private not ts private.

So it cannot be accessed outside of the class.

The private keyword from ts is unsafe as it just marks variables as private, but then they are accessible with [] or as any, or in js

jkoenig134: image

Milena-Czierlinski: Great, thank you! :D

PR #525, opened 2025-05-07
Mark Attributes as deleted if Relationship is decomposed

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


packages/runtime/src/modules/RequestModule.ts
```diff
@@ -332,6 +332,46 @@ export class RequestModule extends RuntimeModule {
...
+                return;
+            }
+
+            for (const peerSharedAttribute of peerSharedAttributesResult.value) {
+                await services.consumptionServices.attributes.setDeletionInfoOfAttribute({
+                    attributeId: peerSharedAttribute.id,
+                    deletionInfo: peerSharedAttributeDeletionInfo
+                });
+            }
+        }
```
jkoenig134, line 373: this is attribute logic, not request and therefore in the wrong module.

packages/runtime/src/useCases/consumption/attributes/SetDeletionInfoOfAttribute.ts
```diff
@@ -0,0 +1,44 @@
...
+        @Inject private readonly attributeController: AttributesController,
+        @Inject private readonly accountController: AccountController,
+        @Inject validator: Validator
+    ) {
+        super(validator);
+    }
+
+    protected async executeInternal(request: SetDeletionInfoOfAttributeRequest): Promise<Result<LocalAttributeDTO>> {
+        const attribute = await this.attributeController.getLocalAttribute(CoreId.from(request.attributeId));
+        if (!attribute) return Result.fail(RuntimeErrors.general.recordNotFound(LocalAttribute.name));
```
jkoenig134, line 35: there should be a check that this is only allowed if the other side is in deleted

packages/runtime/src/useCases/consumption/attributes/SetDeletionInfoOfAttribute.ts
jkoenig134, line 1: please add tests for this usecase, especially for the check that I supposed.

jkoenig134 (changes requested):

packages/runtime/src/modules/RequestModule.ts
Milena-Czierlinski: I had the same thought, but didn't know what would be a better place. Where would you suggest to but it?

packages/runtime/src/useCases/consumption/attributes/SetDeletionInfoOfAttribute.ts
britsta: I also think that the use case should be restricted so that the LocalAttributeDeletionInfo can only be set if the Relationship has `"DeletionProposed"` as `status`.

packages/runtime/src/modules/RequestModule.ts
jkoenig134: I think the only solution is a new module..

jkoenig134: or some kind of cleanup solution that is done without eventing directly in consumtion.

packages/runtime/src/modules/RequestModule.ts
```diff
@@ -332,6 +332,46 @@ export class RequestModule extends RuntimeModule {
             return;
         }
 
+        if (relationship.status === RelationshipStatus.DeletionProposed) {
+            const deletionDate = CoreDate.utc().toString();
```
britsta, line 336: The `deletionDate` should rather be the date of the decomposition of the Relationship, right?

packages/runtime/src/modules/RequestModule.ts
jkoenig134: hm, I see we abuse this module for all kinds of things, just leave it where it is for now.

packages/runtime/src/modules/RequestModule.ts
jkoenig134: do we have that date?

Milena-Czierlinski: If we set the deletionInfo by receiving a Notification, we also set the time of receiving the Notification.

jkoenig134: > If we set the deletionInfo by receiving a Notification, we also set the time of receiving the Notification.

If we have the timestamp the other side deleted the attribute ... why not use this timestamp?

jkoenig134: > If we set the deletionInfo by receiving a Notification, we also set the time of receiving the Notification.

Please don't always use other code as an excuse.

Milena-Czierlinski: I suppose we could get it from the RelationshipAuditLogEntry.

packages/runtime/src/useCases/consumption/attributes/SetDeletionInfoOfAttribute.ts
```diff
@@ -0,0 +1,44 @@
...
+    };
+}
+
+class Validator extends SchemaValidator<SetDeletionInfoOfAttributeRequest> {
+    public constructor(@Inject schemaRepository: SchemaRepository) {
+        super(schemaRepository.getSchema("SetDeletionInfoOfAttributeRequest"));
+    }
+}
+
+export class SetDeletionInfoOfAttributeUseCase extends UseCase<SetDeletionInfoOfAttributeRequest, LocalAttributeDTO> {
```
britsta, line 24: We should also validate that the use case can only be called with reasonable values, i.e. `"DeletedByOwner"` as `deletionStatus` only if the integrator themselves is not the `owner` and `"DeletedByPeer"` as `deletionStatus` only if they are the `owner`, or that no future `deletionDate` may be set, as this use case could otherwise be used for improper purposes. 

Instead of adding these validations, it might be easier if we implement the use case more specifically: `setDeletionInfoOfAttributesOfRelationshipDecomposedByPeer` (of course, a shorter name would be better and should be chosen if possible), for example. The use case would then no longer need a `deletionInfo` input parameter, and we could insert the entire logic of the RequestModule here. I would consider this the safer solution if we only plan to utilize the use case for this purpose anyway.

Milena-Czierlinski: `LocalAttribute.setDeletionInfo` actually already contains validation.

Milena-Czierlinski: Nevertheless, we could also have a more dedicated use case if we wish to have that. :)

britsta: > `LocalAttribute.setDeletionInfo` actually already contains validation.

Ahh, that's nice!

> Nevertheless, we could also have a more dedicated use case if we wish to have that. :)

Yeah, I would still bring up the second part of my comment.

packages/consumption/src/consumption/ConsumptionCoreErrors.ts
```diff
@@ -229,6 +229,13 @@ class Attributes {
         );
     }
 
+    public wrongRelationshipStatusToSetDeletionInfo() {
+        return new CoreError(
+            "error.consumption.attributes.wrongRelationshipStatusToSetDeletionInfo",
+            "In order to manually set the deletionInfo of an Attribute, the corresponding Relationship must be in status DeletionProposed."
```
britsta, line 235: For consistency with other error messages.

```suggestion
            "In order to manually set the deletionInfo of an Attribute, the corresponding Relationship must be in status 'DeletionProposed'."
```

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1464,4 +1464,47 @@ export class AttributesController extends ConsumptionBaseController {
...
+        await this.setDeletionInfoOfPeerSharedAttributes(relationship.peer.address.toString(), deletionDate);
+    }
+
+    private async setDeletionInfoOfOwnSharedAttributes(peer: String, deletionDate: CoreDate): Promise<void> {
+        const ownSharedAttributeDeletionInfo = LocalAttributeDeletionInfo.from({
+            deletionStatus: LocalAttributeDeletionStatus.DeletedByPeer,
+            deletionDate
+        });
+
+        const ownSharedAttributes = await this.getLocalAttributes({ "shareInfo.peer": peer, "content.owner": this.identity.address.toString() });
```
britsta, line 1488: Only those own shared Attributes that have not yet `"DeletedByPeer"` as `status` should be queried here. Otherwise, the correct `deletionDate` would be incorrectly overwritten with the date of the decomposition, right?

packages/consumption/src/modules/attributes/AttributesController.ts
```diff
@@ -1464,4 +1464,47 @@ export class AttributesController extends ConsumptionBaseController {
...
+        await this.setDeletionInfoOfAttributes(ownSharedAttributes, ownSharedAttributeDeletionInfo);
+    }
+
+    private async setDeletionInfoOfPeerSharedAttributes(peer: String, deletionDate: CoreDate): Promise<void> {
+        const peerSharedAttributeDeletionInfo = LocalAttributeDeletionInfo.from({
+            deletionStatus: LocalAttributeDeletionStatus.DeletedByOwner,
+            deletionDate
+        });
+
+        const peerSharedAttributes = await this.getLocalAttributes({ "shareInfo.peer": peer, "content.owner": peer });
```
britsta, line 1499: Only those peer shared Attributes that have not yet `"DeletedByOwner"` as `status` should be queried here. Otherwise, the correct `deletionDate` would be incorrectly overwritten with the date of the decomposition.

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -3362,3 +3368,85 @@ describe("ThirdPartyRelationshipAttributes", () => {
...
+
+    test("should return an error if there is no matching Relationship", async () => {
+        await services1.transport.relationships.terminateRelationship({ relationshipId });
+        await services1.transport.relationships.decomposeRelationship({ relationshipId });
+
+        const result = await services1.consumption.attributes.setAttributeDeletionInfoOfDeletionProposedRelationship({ relationshipId });
+        expect(result).toBeAnError("'Relationship' not found.", "error.transport.recordNotFound");
+    });
+
+    test("should return an error if the Relationship doesn't have status DeletionProposed", async () => {
```
britsta, line 3445:
```suggestion
    test("should return an error if the Relationship doesn't have status 'DeletionProposed'", async () => {
```

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -3362,3 +3368,85 @@ describe("ThirdPartyRelationshipAttributes", () => {
...
+    let services2: TestRuntimeServices;
+    let relationshipId: string;
+
+    beforeEach(async () => {
+        [services1, services2] = await runtimeServiceProvider.launch(2, { enableRequestModule: true, enableDeciderModule: true });
+        const relationship = await ensureActiveRelationship(services1.transport, services2.transport);
+        relationshipId = relationship.id;
+    }, 30000);
+
+    test("peer shared attributes should be marked as deleted for peer", async () => {
```
britsta, line 3383:
```suggestion
    test("peer shared Attributes should be marked as deleted for peer", async () => {
```

packages/runtime/test/consumption/attributes.test.ts
```diff
@@ -3362,3 +3368,85 @@ describe("ThirdPartyRelationshipAttributes", () => {
...
+        const peerSharedAttributes = (await services2.consumption.attributes.getPeerSharedAttributes({ peer: services1.address })).value;
+        expect(peerSharedAttributes).toHaveLength(1);
+        expect(peerSharedAttributes[0].deletionInfo!.deletionStatus).toBe(LocalAttributeDeletionStatus.DeletedByOwner);
+
+        const relationship = (await services2.transport.relationships.getRelationship({ id: relationshipId })).value;
+        const deletionDate = relationship.auditLog[relationship.auditLog.length - 1].createdAt;
+        expect(peerSharedAttributes[0].deletionInfo!.deletionDate).toStrictEqual(deletionDate);
+    });
+
+    test("own shared attributes should be marked as deleted for peer", async () => {
```
britsta, line 3410:
```suggestion
    test("own shared Attributes should be marked as deleted for peer", async () => {
```

jkoenig134 (dismissed):
awesome!

britsta (approved):
LGTM 🐸

PR #531, opened 2025-05-08
Update the FileDVO to contain `reference.truncated` and `reference.url`

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

`truncatedReference` is removed b/c the app doesn't need that value and it seems to be broken anyways.

### DIFF
--- a/packages/runtime/src/dataViews/DataViewExpander.ts
+++ b/packages/runtime/src/dataViews/DataViewExpander.ts
@@ -1811,7 +1811,7 @@ export class DataViewExpander {
             filename: file.filename,
             filesize: file.filesize,
             createdBy: await this.expandAddress(file.createdBy),
-            truncatedReference: file.truncatedReference
+            reference: file.reference
         };
     }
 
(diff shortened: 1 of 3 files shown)


britsta (approved):

packages/runtime/src/dataViews/DataViewExpander.ts
```diff
@@ -1811,7 +1811,7 @@ export class DataViewExpander {
             filename: file.filename,
             filesize: file.filesize,
             createdBy: await this.expandAddress(file.createdBy),
-            truncatedReference: file.truncatedReference
+            reference: file.reference
```
Milena-Czierlinski, line 1814: For the DTOs we have removed truncatedReference only with release/v7. Why can we do it on the DVO already here?

jkoenig134: see PR description.

Milena-Czierlinski: Wow, I totally overlooked that. 😅

Milena-Czierlinski (approved):

PR #532, opened 2025-05-08
Misleading error code used

britsta:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description
The `invalidAcceptParameters` error code should not be used if the accept parameters when trying to accept a RequestItem are valid but the RequestItem cannot be accepted for other reasons. Otherwise, the error code gives the impression that the recipient can change the fact that the RequestItem is not acceptable by using other accept parameters, which is not the case.

The `invalidRequestItem` error code may be appropriate instead if the RequestItem contains errors that should have been prevented on the sender side anyway. In other cases, more specific error messages could be used.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
                 "@js-soft/eslint-config-ts": "^1.6.14",
                 "@js-soft/license-check": "^1.0.9",
                 "@types/jest": "^29.5.14",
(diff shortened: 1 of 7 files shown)


jkoenig134 (approved):
great! :)

PR #552, opened 2025-05-15
Mail recipients are not validated

Magnus-Kuhn:
# Readiness checklist

- [X] I added/updated tests.
- [X] I labeled the PR.
- [X] I self-reviewed the PR.

# Description

It's currently possible to put wrong values into to/cc of a Mail, causing the app's inbox to not work.

### DIFF
--- a/packages/runtime/src/useCases/transport/messages/SendMessage.ts
+++ b/packages/runtime/src/useCases/transport/messages/SendMessage.ts
@@ -77,6 +77,9 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO> 
         const recipientsValidationError = await this.validateMessageRecipients(transformedContent, recipients);
         if (recipientsValidationError) return recipientsValidationError;
 
+        const mailRecipientsValidationError = this.validateMailRecipients(transformedContent, recipients);
+        if (mailRecipientsValidationError) return mailRecipientsValidationError;
+
         if (transformedContent instanceof Request) {
             const requestValidationError = await this.validateRequestAsMessageContent(transformedContent, CoreAddress.from(recipients[0]));
             if (requestValidationError) return requestValidationError;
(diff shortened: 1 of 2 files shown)


jkoenig134 (approved):

packages/runtime/src/useCases/transport/messages/SendMessage.ts
```diff
@@ -133,6 +136,27 @@ export class SendMessageUseCase extends UseCase<SendMessageRequest, MessageDTO>
         return;
     }
 
+    private validateMailRecipients(content: Serializable, recipients: string[]): ApplicationError | undefined {
```
britsta, line 139: I would rename the method as it gives the impression that the Mail recipients are passed as input parameters when in fact the Message recipients are passed.

jkoenig134:
```suggestion
    private validateMailRecipients(mail: Serializable, messageRecipients: string[]): ApplicationError | undefined {
```
@britsta I think the method name is fine, the argument wording is not.

britsta: Also fine for me @jkoenig134

britsta (approved):

jkoenig134 (dismissed):

britsta (commented):
Do we also want to validate that it is not possible to specify the same address more than once in `to` (or in `cc`) of the Mail?

jkoenig134:
> > Do we also want to validate that it is not possible to specify the same address more than once in `to` (or in `cc`) of the Mail?
> 
> We could add this kind of validation, we've neglected this uniqueness a lot though - the message recipients, attribute tags,... are not validated in this regard

Is this the only argument on why not to validate how @britsta suggested? In that case I'd say validate it :)

britsta:
> We could add this kind of validation, we've neglected this uniqueness a lot though - the message recipients, attribute tags,... are not validated in this regard.

You're right, there are more places in the code base where we don't validate this kind of uniqueness than we validate it. However, I think we can add it if we are on the validation of the mail recipients anyway.

jkoenig134 (approved):

britsta (approved):

PR #580, opened 2025-05-28
Support `ownershipToken` of Files

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


packages/transport/src/modules/sync/ChangedItems.ts
Milena-Czierlinski, line 1: The Backbone sends us the id of the File that is locked. If we only use this in the event in transport and runtime, we wouldn't need to adjust ChangedItems. However, whenever possible we use the objects over the ids in the events. I'm not sure which is better here. @jkoenig134 Do you have an opinion on this?

jkoenig134: I'm not sure if I understand what your question is. Maybe we do a quick call?

packages/runtime/test/transport/files.test.ts
```diff
@@ -338,23 +338,23 @@ describe("Load peer file with token reference", () => {
         const response = await transportServices2.files.getOrLoadFile({ reference: token.truncatedReference });
 
         expect(response).toBeSuccessful();
-        expect(response.value).toMatchObject({ ...file, isOwn: false });
+        expect(response.value).toMatchObject({ ...file, isOwn: false, ownershipToken: undefined });
```
jkoenig134, line 341: Hm, we should do a quick call about all these added lines in the existing tests.

jkoenig134 (dismissed):
some little details

packages/transport/test/modules/files/FileController.test.ts
```diff
@@ -154,4 +154,115 @@ describe("FileController", function () {
...
+            expect(updatedFile.cache!.owner).toStrictEqual(sender.identity.address);
+            expect(updatedFile.isOwn).toBe(true);
+
+            expect(updatedFile.ownershipToken).toBeDefined();
+            expect(updatedFile.ownershipToken).not.toBe(ownershipToken);
+        });
+
+        test("should not claim the ownership of a File with an invalid ownershipToken", async function () {
+            await TestUtil.expectThrowsRequestErrorAsync(recipient.files.claimFileOwnership(senderFile.id, "invalid-token"), "error.platform.unexpected", 403);
+        });
```
britsta, line 222: Writing `"anInvalidOwnershipToken"` (or `"anInvalidToken"`) should match our test value naming convention better than using `"invalid-token"`. Similar for other tests.

```suggestion
        test("should not claim the ownership of a File with an invalid ownershipToken", async function () {
            await TestUtil.expectThrowsRequestErrorAsync(recipient.files.claimFileOwnership(senderFile.id, "anInvalidOwnershipToken"), "error.platform.unexpected", 403);
        });
```

britsta (approved):

PR #589, opened 2025-06-03
Transfer File ownership using `ownershipToken`

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/consumption/src/modules/requests/itemProcessors/transferFileOwnership/TransferFileOwnershipRequestItemProcessor.ts
+++ b/packages/consumption/src/modules/requests/itemProcessors/transferFileOwnership/TransferFileOwnershipRequestItemProcessor.ts
@@ -117,21 +117,9 @@ export class TransferFileOwnershipRequestItemProcessor extends GenericRequestIte
     ): Promise<TransferFileOwnershipAcceptResponseItem> {
         const peerFile = await this.accountController.files.getOrLoadFileByReference(requestItem.fileReference);
 
-        let ownFile: File;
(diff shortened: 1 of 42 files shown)


packages/consumption/src/modules/requests/itemProcessors/transferFileOwnership/TransferFileOwnershipRequestItemProcessor.ts
```diff
@@ -94,25 +116,30 @@ export class TransferFileOwnershipRequestItemProcessor extends GenericRequestIte
...
+            ownFile = await this.accountController.files.sendFile({
+                buffer: fileContent,
+                title: peerFile.cache!.title,
+                description: peerFile.cache!.description,
+                filename: peerFile.cache!.filename,
+                mimetype: peerFile.cache!.mimetype,
+                expiresAt: CoreDate.from("9999-12-31T00:00:00.000Z"),
+                tags: peerFile.cache!.tags
+            });
+        }
```
jkoenig134, line 134: we should stop creating this kind of code. there are many effective ways to avoid those floating non-const variables. In that case I'd heavily vote for extracting everything below to a private function and then you can also get rid of the else :)

jkoenig134: the more I think about it - you should extract this if else to a private function

jkoenig134 (approved):

PR #590, opened 2025-06-03
Check `ownershipIsLocked` after failed validation

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


jkoenig134 (dismissed):
hm, maybe we should reword this that it doesn't sound like we test backbone features

jkoenig134:
> > hm, maybe we should reword this that it doesn't sound like we test backbone features
> 
> Better now?

I meant the test names, but I approved, so if you don't agree please merge.

jkoenig134 (approved):

PR #602, opened 2025-06-10
The ownership of the original File might not be transferred using a TransferFileOwnershipRequestItem

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/consumption/test/modules/requests/itemProcessors/transferFileOwnership/TransferFileOwnershipRequestItemProcessor.test.ts
+++ b/packages/consumption/test/modules/requests/itemProcessors/transferFileOwnership/TransferFileOwnershipRequestItemProcessor.test.ts
@@ -51,7 +51,8 @@ describe("TransferFileOwnershipRequestItemProcessor", function () {
 
             const requestItem = TransferFileOwnershipRequestItem.from({
                 mustBeAccepted: false,
-                fileReference: senderFile.toFileReference(senderAccountController.config.baseUrl).truncate()
+                fileReference: senderFile.toFileReference(senderAccountController.config.baseUrl).truncate(),
+                ownershipToken: senderFile.ownershipToken!
             });
             const request = Request.from({ items: [requestItem] });
 
@@ -64,7 +65,8 @@ describe("TransferFileOwnershipRequestItemProcessor", function () {
 
             const requestItem = TransferFileOwnershipRequestItem.from({
                 mustBeAccepted: false,
-                fileReference: senderFile.toFileReference(senderAccountController.config.baseUrl)
+                fileReference: senderFile.toFileReference(senderAccountController.config.baseUrl),
+                ownershipToken: senderFile.ownershipToken!
             });
             const request = Request.from({ items: [requestItem] });
 
@@ -78,7 +80,7 @@ describe("TransferFileOwnershipRequestItemProcessor", function () {
             const requestItem = TransferFileOwnershipRequestItem.from({
                 mustBeAccepted: false,
                 fileReference: senderFile.toFileReference(senderAccountController.config.baseUrl),
-                ownershipToken: senderFile.ownershipToken
+                ownershipToken: senderFile.ownershipToken!
             });
             const request = Request.from({ items: [requestItem] });
 
@@ -91,7 +93,8 @@ describe("TransferFileOwnershipRequestItemProcessor", function () {
 
             const requestItem = TransferFileOwnershipRequestItem.from({
                 mustBeAccepted: false,
-                fileReference: thirdPartyFile.toFileReference(thirdPartyAccountController.config.baseUrl)
+                fileReference: thirdPartyFile.toFileReference(thirdPartyAccountController.config.baseUrl),
+                ownershipToken: thirdPartyFile.ownershipToken!
             });
             const request = Request.from({ items: [requestItem] });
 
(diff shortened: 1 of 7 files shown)


packages/consumption/test/modules/requests/itemProcessors/transferFileOwnership/TransferFileOwnershipRequestItemProcessor.test.ts
```diff
@@ -78,7 +80,7 @@ describe("TransferFileOwnershipRequestItemProcessor", function () {
             const requestItem = TransferFileOwnershipRequestItem.from({
                 mustBeAccepted: false,
                 fileReference: senderFile.toFileReference(senderAccountController.config.baseUrl),
-                ownershipToken: senderFile.ownershipToken
+                ownershipToken: senderFile.ownershipToken!
             });
```
britsta, line 84: The corresponding test `"returns success if the ownership of an own File should be transferred using an ownershipToken"` isn't needed anymore as the previous test `"returns success if the ownership of an own File should be transferred entering a FileReference"` now also must use an ownershipToken. 

Furthermore, the previous tests could be renamed.

packages/consumption/test/modules/requests/itemProcessors/transferFileOwnership/TransferFileOwnershipRequestItemProcessor.test.ts
```diff
@@ -184,7 +190,7 @@ describe("TransferFileOwnershipRequestItemProcessor", function () {
...
```
britsta, line 194: The corresponding test `"returns success when checking if the transfer of ownership of a File can be accepted with an ownershipToken"` isn't needed anymore as the previous test `"returns success when checking if the transfer of ownership of a File can be accepted that is owned by the sender"` now also must use an ownershipToken.

packages/consumption/test/modules/requests/itemProcessors/transferFileOwnership/TransferFileOwnershipRequestItemProcessor.test.ts
Milena-Czierlinski: You're absolutely correct!  I'll remove it and some more tests.

What do you mean by renaming the previous test? The name ist chosen to underline, that either a FileReference or a truncated reference (as in the first test) can be entered to create the RequestItem.

britsta: Ahh I see, then renaming is not necessary. 👍

packages/consumption/test/modules/requests/itemProcessors/transferFileOwnership/TransferFileOwnershipRequestItemProcessor.test.ts
```diff
@@ -314,46 +277,7 @@ describe("TransferFileOwnershipRequestItemProcessor", function () {
...
-
-            const repositoryAttribute = await recipientConsumptionController.attributes.getLocalAttribute(ownSharedIdentityAttribute!.shareInfo!.sourceAttribute!);
-            expect(repositoryAttribute!.shareInfo).toBeUndefined();
-            expect(repositoryAttribute!.content.value).toBeInstanceOf(IdentityFileReference);
-            expect((repositoryAttribute!.content as IdentityAttribute).tags).toStrictEqual(["x:tag"]);
-
-            const fileReference = FileReference.from((repositoryAttribute!.content.value as IdentityFileReference).value);
-            const file = await recipientAccountController.files.getFile(fileReference.id);
-            expect(file!.isOwn).toBe(true);
-            expect(file!.cache!.tags).toStrictEqual(["x:tag"]);
```
britsta, line 343: This check could be added for the `claimedFile` of the test below.

packages/consumption/test/modules/requests/itemProcessors/transferFileOwnership/TransferFileOwnershipRequestItemProcessor.test.ts
```diff
@@ -397,47 +321,7 @@ describe("TransferFileOwnershipRequestItemProcessor", function () {
...
-
-            const peerSharedIdentityAttribute = await senderConsumptionController.attributes.getLocalAttribute(responseItem.attributeId);
-            expect(peerSharedIdentityAttribute!.shareInfo!.peer).toStrictEqual(recipient);
-            expect(peerSharedIdentityAttribute!.shareInfo!.requestReference).toStrictEqual(requestInfo.id);
-            expect(peerSharedIdentityAttribute!.shareInfo!.sourceAttribute).toBeUndefined();
-
-            const truncatedFileReference = (peerSharedIdentityAttribute!.content.value as IdentityFileReference).value;
-            const file = await senderAccountController.files.getOrLoadFileByReference(FileReference.from(truncatedFileReference));
-            expect(file.isOwn).toBe(false);
-            expect(file.cache!.tags).toStrictEqual(["x:tag"]);
```
britsta, line 426: This check could be added for the `claimedFile` of the test below.

Milena-Czierlinski: I don't think this is necessary and in fact moves the focus from what is important in this test. The only thing that changes about the file is its owner. Therefore, I wouldn't add checks for properties that aren't touched by the processor. The tag is only relevant for the created Attribute. Hence, I'll add a check for that. In the test you commented on above, the checks of the tag of the Attributes are already there.

britsta (approved):

PR #625, opened 2025-06-25
Update communication language

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/app-runtime/src/modules/AppLanguageModule.ts
+++ b/packages/app-runtime/src/modules/AppLanguageModule.ts
@@ -1,0 +1,54 @@
+import { LanguageISO639 } from "@nmshd/core-types";
+import { AccountSelectedEvent, AppLanguageChangedEvent } from "../events";
+import { AppRuntimeModule, AppRuntimeModuleConfiguration } from "./AppRuntimeModule";
+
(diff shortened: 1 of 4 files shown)


packages/app-runtime/src/modules/AppLanguageModule.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    private async handleAccountSelected(event: AccountSelectedEvent) {
+        this.logger.trace("AppLanguageModule.handleAccountSelected", event);
+        const languageResult = await this.runtime.languageProvider.getAppLanguage();
+        if (languageResult.isError) {
+            this.logger.error(languageResult.error);
+            return;
+        }
+
+        await this.updateLanguageForLocalAccount(event.data.address, languageResult.value);
```
Milena-Czierlinski, line 37: Do we store the communication language somewhere such that we can only update it if it actually changes?

jkoenig134: nah, also this is the same as we do with the push token

jkoenig134: It is! But then this has to be synced et al, I don't know how long this can take. But yeah, we can do that ASAP :)

PR #633, opened 2025-07-01
Endless loop when updating request expiry

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

Before #612 there could be multiple requests with the same ID. This can cause an endless loop when those requests are queried. With this fix we stop this endless loop.

### DIFF
--- a/packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
+++ b/packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
@@ -433,7 +433,7 @@ export class IncomingRequestsController extends ConsumptionBaseController {
             isOwn: false
         });
 
-        const requestPromises = requestDocs.map((r) => this.updateRequestExpiry(LocalRequest.from(r)));
+        const requestPromises = requestDocs.map((r) => this.updateRequestExpiry(r, LocalRequest.from(r)));
         return await Promise.all(requestPromises);
     }
 
@@ -443,7 +443,7 @@ export class IncomingRequestsController extends ConsumptionBaseController {
 
         const localRequest = LocalRequest.from(requestDoc);
 
-        return await this.updateRequestExpiry(localRequest);
+        return await this.updateRequestExpiry(requestDoc, localRequest);
     }
 
     private async getOrThrow(id: CoreId | string) {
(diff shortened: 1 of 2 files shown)


packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
```diff
@@ -483,12 +483,12 @@ export class IncomingRequestsController extends ConsumptionBaseController {
...
     }
 
-    private async updateRequestExpiry(request: LocalRequest) {
+    private async updateRequestExpiry(requestDoc: any, request: LocalRequest) {
         const oldStatus = request.status;
 
         const statusUpdated = request.updateStatusBasedOnExpiration();
         if (statusUpdated) {
-            await this.update(request);
+            await this.localRequests.update(requestDoc, request);
```
Milena-Czierlinski, line 491: There are other places in this file, where `this.update` is called which internally uses `findOne`. I know that trying to update the Request expiry was the main problem, but did you take the other occurrences into consideration, as well?

jkoenig134: I did, but that's not what we want to fix here.

Milena-Czierlinski: And what was your outcome? 😂 Should adjust something there too, even if it's not part of this PR?

Milena-Czierlinski (approved):

packages/consumption/src/modules/requests/incoming/IncomingRequestsController.ts
jkoenig134: Hm, the outcome is that there will be no more duplicate entries in the future so we don't want to waste time for things that are irrelevant

Milena-Czierlinski: I understand that there won't be new duplicate entries in the future. The question I am wondering about is if the currently present duplicates might cause problems in other places than the Request expiry, e.g. if another status is not set. We could argue that hopefully all these Requests have an expiration date set, that is already in the past and therefore we don't have to invest time into investigating such cases and if they might cause problems we haven't discovered yet. However, I wouldn't necessarily view this as a waste of time. But I am okay with prioritizing other topics higher.

jkoenig134: if no expiration was set we have no endless loop

PR #645, opened 2025-07-03
An external event that was created during a sync can be missed until the next sync

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/.dev/compose.backbone.yml
+++ b/.dev/compose.backbone.yml
@@ -1,6 +1,6 @@
 services:
   consumer-api:
-    image: ghcr.io/nmshd/backbone-consumer-api:6.52.1
+    image: ghcr.io/nmshd/backbone-consumer-api:6.52.2
     container_name: consumer-api
     hostname: consumer-api
     environment:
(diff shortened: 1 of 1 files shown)


packages/transport/src/modules/sync/SyncController.ts
```diff
@@ -122,16 +122,21 @@ export class SyncController extends TransportController {
     }
 
     private async syncExternalEvents(changedItems: ChangedItems): Promise<void> {
-        const syncRunWasStarted = await this.startExternalEventsSyncRun();
+        let newUnsyncedExternalEventsExist = false;
```
tnotheis, line 125: I don't know if a comment might be necessary that explains how this case could happen and why we should avoid it.

jkoenig134: Yeah, maybe we should.

jkoenig134: I'll discuss this change with @britsta and @Milena-Czierlinski and we will see how easy this is to understand.

jkoenig134: We all had the same opinion that the comment is not necessary here :)

britsta: As just discussed with @jkoenig134, I don't think a comment is necessary because the variable names are well chosen. However, I have no strong opinion against adding a comment.

jkoenig134:
#647 updates the backbone for this feature.

tnotheis (approved):

PR #671, opened 2025-07-18
Remove token caches

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [ ] I self-reviewed the PR.


packages/transport/test/modules/tokens/TokenController.test.ts
```diff
@@ -369,68 +365,31 @@ describe("TokenController", function () {
...
-            ephemeral: false,
-            passwordProtection: { password: "1234", passwordType: "pin4" }
-        });
-        const reference2 = sentToken2.toTokenReference(sender.config.baseUrl);
-
-        const receivedToken1 = await recipient.tokens.loadPeerTokenByReference(reference1, false, "password");
-        const receivedToken2 = await recipient.tokens.loadPeerTokenByReference(reference2, false, "1234");
-        const fetchCachesResult = await recipient.tokens.fetchCaches([receivedToken1.id, receivedToken2.id]);
-        expect(fetchCachesResult).toHaveLength(2);
-    });
```
britsta, line 396: Why was this test removed?

packages/transport/test/modules/tokens/TokenController.test.ts
```diff
@@ -369,68 +365,31 @@ describe("TokenController", function () {
...
 
-        test("should delete a peer owned token only locally", async function () {
-            await recipient.tokens.delete(receivedToken);
-            const tokenOnBackbone = await sender.tokens.fetchCaches([sentToken.id]);
-            expect(tokenOnBackbone).toHaveLength(1);
-
-            const localToken = await recipient.tokens.getToken(sentToken.id);
-            expect(localToken).toBeUndefined();
+            await expect(recipient.tokens.loadPeerTokenByReference(sentToken.toTokenReference(sender.config.baseUrl), false)).rejects.toThrow("error.platform.recordNotFound");
         });
```
britsta, line 382: Where is it checked within the `test("should delete own token locally and from the Backbone")` that the deletion of an own Token leads to the deletion of it from the Backbone?

Furthermore, the name of the test needs to be adapted if the `test("should delete a peer owned token only locally")` should be included within this test now.

britsta (changes requested):

packages/transport/test/modules/tokens/TokenController.test.ts
jkoenig134: ehm, no? the sender deletes the token, and we want to make sure, that it's also deleted from the backbone. Nothing peer owned here.

jkoenig134: I added a comment + changed the name of the other test to make that clear.

britsta (approved):

PR #678, opened 2025-07-22
Remove certificates transport module

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

The code is not used and consists mostly of comments.


jkoenig134:
I think the certificate in Device was another certificate, but b/c it was always an empty string I removed it.

PR #713, opened 2025-08-12
Backbone eventual consistency can cause item to not be synced until the next push notification

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/app-runtime/src/modules/PushNotificationModule.ts
+++ b/packages/app-runtime/src/modules/PushNotificationModule.ts
@@ -51,9 +51,7 @@ export class PushNotificationModule extends AppRuntimeModule<PushNotificationMod
                 case BackboneEventName.ExternalEventCreated:
                     const result = await this.syncEverythingUntilHasChanges(services);
 
-                    if (result) {
-                        this.runtime.eventBus.publish(new ExternalEventReceivedEvent(accRef, result));
-                    }
+                    if (result) this.runtime.eventBus.publish(new ExternalEventReceivedEvent(accRef, result));
 
                     break;
                 default:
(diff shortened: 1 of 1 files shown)


packages/app-runtime/src/modules/PushNotificationModule.ts
```diff
@@ -63,6 +64,41 @@ export class PushNotificationModule extends AppRuntimeModule<PushNotificationMod
...
+    private async syncEverythingUntilHasChanges(services: RuntimeServices): Promise<SyncEverythingResponse | undefined> {
+        const hasChanges = (value: SyncEverythingResponse): boolean =>
+            Object.values(value)
+                .filter((v) => Array.isArray(v))
+                .some((array) => array.length > 0);
+
+        const syncResult = await services.transportServices.account.syncEverything();
+
+        if (syncResult.isError) {
+            this.logger.error(syncResult.error);
```
britsta, line 76: Why didn't we have an early return if an error has occurred once? Is it likely that retrying will help in that case?

packages/app-runtime/src/modules/PushNotificationModule.ts
```diff
@@ -63,6 +64,41 @@ export class PushNotificationModule extends AppRuntimeModule<PushNotificationMod
...
+        } else if (hasChanges(syncResult.value)) {
+            return syncResult.value;
+        }
+
+        let retries = 0;
+        const maxRetries = 5;
```
britsta, line 82: The first attempt already happened at this point.

```suggestion

        this.logger.info(`No changes found in sync attempt 1. Retrying...`);

        let retries = 1;
        const maxRetries = 5;
```

britsta: Assuming there is no error, the first logging info will be `No changes found in sync attempt 1. Retrying...` even though there were already 2 sync attempts.

jkoenig134: updated - but in my way

PR #727, opened 2025-09-03
Allow to show the device onboarding QRCode on the onboarded device instead of the onboarding

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

### DIFF
--- a/packages/transport/src/modules/tokens/local/EmptyToken.ts
+++ b/packages/transport/src/modules/tokens/local/EmptyToken.ts
@@ -25,10 +25,6 @@ export class EmptyToken extends Serializable implements IEmptyToken {
     @serialize()
     public expiresAt: CoreDate;
 
-    @validate({ nullable: true })
-    @serialize()
-    public password: string;
-
     @validate()
     @serialize()
     public passwordProtection: PasswordProtection;


packages/app-runtime/test/runtime/Onboarding.test.ts
```diff
@@ -82,4 +84,24 @@ describe("Onboarding", function () {
         });
         expect(recoveryKitResponse2).toBeSuccessful();
     });
+
+    test("should onboard with an anonomously created empty token", async () => {
```
tnotheis, line 88:
```suggestion
    test("should onboard with an anonymously created empty token", async () => {
```

packages/app-runtime/test/runtime/Onboarding.test.ts
```diff
@@ -82,4 +84,24 @@ describe("Onboarding", function () {
...
+        const emptyToken = await onboardingRuntime.anonymousServices.tokens.createEmptyToken();
+
+        const device = await services.transportServices.devices.createDevice({});
+        const onboardingInfo = await services.transportServices.devices.getDeviceOnboardingInfo({ id: device.value.id });
+
+        const tokenContent = TokenContentDeviceSharedSecret.from({ sharedSecret: DeviceMapper.toDeviceSharedSecret(onboardingInfo.value) });
+        const result = await services.transportServices.tokens.updateTokenContent({ reference: emptyToken.value.reference.truncated, content: tokenContent.toJSON() });
+        expect(result).toBeSuccessful();
```
tnotheis, line 96: I would probably prefer a single method doing all of this (=> fetching the token, creating the content, updating the token), so that the user of the app runtime doesn't know how to do it.

packages/app-runtime/test/runtime/Onboarding.test.ts
```diff
@@ -82,4 +84,24 @@ describe("Onboarding", function () {
...
+
+        const token = await onboardingRuntime.anonymousServices.tokens.loadPeerToken({ reference: emptyToken.value.reference.truncated });
+        expect(token.value.content["@type"]).toBe("TokenContentDeviceSharedSecret");
+
+        const content = token.value.content as ITokenContentDeviceSharedSecret;
+        const deviceOnboardingDTO = DeviceMapper.toDeviceOnboardingInfoDTO(TokenContentDeviceSharedSecret.from(content).sharedSecret);
+
+        const onboardResult = await onboardingRuntime.accountServices.onboardAccount(deviceOnboardingDTO);
```
tnotheis, line 104: Same here: a single method that completes the onboarding process by loading the updated token and creating the device would be preferable.

packages/core-types/src/SharedPasswordProtection.ts
```diff
@@ -29,23 +34,30 @@ export class SharedPasswordProtection extends Serializable implements ISharedPas
         if (value === undefined || value === "") return undefined;
```
tnotheis, line 34: There are a lot of changes in this method, but no tests were updated/added.

packages/runtime/src/useCases/common/PasswordProtectionMapper.ts
```diff
@@ -3,6 +3,8 @@ import { PasswordProtectionDTO } from "@nmshd/runtime-types";
 import { PasswordProtection, PasswordProtectionCreationParameters } from "@nmshd/transport";
 
 export class PasswordProtectionMapper {
+    public static toPasswordProtectionDTO(passwordProtection: PasswordProtection): PasswordProtectionDTO;
+    public static toPasswordProtectionDTO(passwordProtection?: PasswordProtection): PasswordProtectionDTO | undefined;
```
tnotheis, line 7: This definition is exactly the same as the one below. Can't you remove it? (I'm not very familiar with the typescript way of overloading)

packages/runtime/src/useCases/transport/tokens/TokenMapper.ts
```diff
@@ -32,6 +32,22 @@ export class TokenMapper {
...
+            id: token.id.toString(),
+            expiresAt: token.expiresAt.toString(),
+            passwordProtection: PasswordProtectionMapper.toPasswordProtectionDTO(token.passwordProtection),
+            reference: {
+                truncated: reference.truncate(),
+                url: reference.toUrl()
+            },
+            isEphemeral: ephemeral
+        };
+    }
```
tnotheis, line 49: Shouldn't there be an EmptyTokenMapper for this method?

packages/runtime/src/useCases/transport/tokens/UpdateTokenContent.ts
```diff
@@ -0,0 +1,52 @@
...
+            tokenContent = Serializable.fromUnknown(request.content);
+        } catch {
+            throw RuntimeErrors.general.invalidTokenContent();
+        }
+
+        const reference = TokenReference.fromTruncated(request.reference);
+
+        const passwordProtection = reference.passwordProtection;
+        if (passwordProtection && !passwordProtection.password) {
+            throw new Error("RuntimeErrors.general.noPasswordProvided()");
```
tnotheis, line 40: This error code (if it should be one) looks kind of weird:
- Did you add these parentheses on purpose?
- The "RuntimeErrors" prefix is something we don't usually do.

Did you mean to do the following?


```suggestion
            throw new Error(RuntimeErrors.general.noPasswordProvided());
```

tnotheis (changes requested):

packages/runtime/src/useCases/transport/tokens/TokenMapper.ts
jkoenig134: Isn't an empty token also a token?

packages/app-runtime/test/runtime/Onboarding.test.ts
jkoenig134: As discussed, we will do this in the app

tnotheis (dismissed):

packages/core-types/src/SharedPasswordProtection.ts
```diff
@@ -29,23 +34,30 @@ export class SharedPasswordProtection extends Serializable implements ISharedPas
...
 
         const splittedPasswordParts = value.split("&");
-        if (splittedPasswordParts.length !== 2 && splittedPasswordParts.length !== 3) {
+        if (![2, 3, 4].includes(splittedPasswordParts.length)) {
             throw new CoreError("error.core-types.invalidTruncatedReference", "The password part of a TruncatedReference must consist of exactly 2 or 3 components.");
```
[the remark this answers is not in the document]
jkoenig134: As you see I added one other segment (the cleartext password), which is the 4. I updated the error message

packages/transport/src/modules/tokens/local/EmptyToken.ts
```diff
@@ -0,0 +1,48 @@
...
+    @serialize()
+    public secretKey: CryptoSecretKey;
+
+    @validate()
+    @serialize()
+    public expiresAt: CoreDate;
+
+    @validate({ nullable: true })
+    @serialize()
+    public password: string;
```
britsta, line 30: Why does the `password` not occur in the interface `IEmptyToken`?

packages/runtime/src/useCases/transport/devices/FillDeviceOnboardingTokenWithNewDevice.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    protected async executeInternal(request: FillDeviceOnboardingTokenWithNewDeviceRequest): Promise<Result<TokenDTO>> {
+        const reference = TokenReference.fromTruncated(request.reference);
+
+        const passwordProtection = reference.passwordProtection;
+        if (!passwordProtection?.password) throw RuntimeErrors.devices.referenceNotPointingToAnEmptyToken();
+
+        const isEmptyToken = await this.tokenController.isEmptyToken(reference);
+        if (!isEmptyToken) throw RuntimeErrors.devices.referenceNotPointingToAnEmptyToken();
```
britsta, line 37: This means non-empty Tokens can have a clear-text password as well?

PR #1028, opened 2026-04-15
Add IdentityAttribute value type `SchematizedJSON`

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description
This can be used like the ProprietaryJSON as an IdentityAttribute.


packages/content/src/attributes/types/identity/SchematizedJSON.ts
```diff
@@ -0,0 +1,56 @@
+import { serialize, type, validate } from "@js-soft/ts-serval";
+import { AbstractAttributeValue, AbstractAttributeValueJSON, IAbstractAttributeValue } from "../../AbstractAttributeValue";
+
+export interface SchematizedJSONJSON extends AbstractAttributeValueJSON {
```
britsta, line 6: > SchematizedJSONJSON

This is funny but just like the ProprietaryJSONJSON 😄

packages/content/src/attributes/types/identity/SchematizedJSON.ts
```diff
@@ -0,0 +1,56 @@
...
+    } catch (e) {
+        if (e instanceof SyntaxError) {
+            return "must be a valid JSON object";
+        }
+
+        return "could not validate value";
+    }
+
+    return undefined;
+}
```
britsta: This is the same validation function as for the ProprietaryJSON `value`. Maybe we can refactor this such that we only need to define this function once.

packages/content/src/attributes/types/identity/SchematizedJSON.ts
```diff
@@ -0,0 +1,56 @@
...
+    schemaURL?: string;
+}
+
+export interface ISchematizedJSON extends IAbstractAttributeValue {
+    value: unknown;
+    schemaURL?: string;
+}
+
+@type("SchematizedJSON")
+export class SchematizedJSON extends AbstractAttributeValue implements ISchematizedJSON {
```
britsta, line 18: Why did you decide for omitting the `valueHints` and `renderHints` methods the ProprietaryJSON class has?

packages/content/test/attributeValues/JSON.test.ts
britsta, line 1: You've combined the tests such as for the ProprietaryXML and SchematizedXML, right? I'd prefer splitting up the file or find a more appropriate file name.

packages/content/src/attributes/types/identity/SchematizedJSON.ts
Milena-Czierlinski: I forgot it. 😬

packages/content/src/attributes/types/utils/validateJSON.ts
```diff
@@ -0,0 +1,16 @@
...
+        if (string.length > 4096) {
+            return "stringified value must not be longer than 4096 characters";
+        }
+    } catch (e) {
+        if (e instanceof SyntaxError) {
+            return "must be a valid JSON object";
+        }
+
+        return "could not validate value";
+    }
```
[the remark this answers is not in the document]
Milena-Czierlinski: This will be handled by the catch block.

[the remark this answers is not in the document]
tnotheis: Is it? I think Copilot is right. When the `string` variable is `undefined`, the check for its length will throw a `TypeError`, which is not an instance of `SyntaxError` and thus will lead to the error message "could not validate value". 

What Copilot is suggesting is that you give a clearer error message in this case.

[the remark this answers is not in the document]
Milena-Czierlinski: Yes, the error message will be `"could not validate value"`. For me that would be sufficient, but I have no problem adding a clearer message for that case if you prefer it.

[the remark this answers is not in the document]
tnotheis: I don't have any hard opinion here. Just wanted to make clear what copilot meant. :)

tnotheis (approved):

PR #1094, opened 2026-07-15
Cache VC `displayInformation` images

Milena-Czierlinski:
# Readiness checklist

- [x] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


packages/transport/src/modules/files/FileController.ts
```diff
@@ -277,4 +278,53 @@ export class FileController extends TransportController {
...
+        }
+
+        const contentTypeHeader = response.headers["content-type"];
+        const mimetype = typeof contentTypeHeader === "string" && contentTypeHeader.length > 0 ? contentTypeHeader.split(";")[0] : "application/octet-stream";
+
+        const file = await this.sendFile({
+            buffer: CoreBuffer.from(response.data),
+            mimetype,
+            filename: imageUrl,
+            expiresAt: CoreDate.from("9999-12-31T00:00:00.000Z")
```
britsta, line 325: Does this make sense? Can the VC have an earlier expiration date?

Milena-Czierlinski: The VC will surely have an earlier expiration date. However, the VC is not automatically removed from the wallet once its expiration date is reached. Instead, it needs to be deleted manually. For it to be still displayed correctly once it is expired (also in offline mode or if the issuing server is unavailable, which is the motivation of caching the images), we need the cached image to be still valid even if the VC is already expired.

But, I think this PR misses the automatic deletion of the cached image files once the VerifiableCredential Attribute is deleted.

britsta: Thank you for the explanation. The deletion of the cached image files once the VC Attribute is deleted makes sense.

Milena-Czierlinski: I implemented the deletion now. Also, I adjusted the type of the cachedFiles from FileReference to FileReference.truncate() (i.e. string).

britsta (approved):

